PR #476, opened 2014-08-20
json_spirit package

highperformancecoder:
json_spirit builds out of the box as a cmake project. The only gotcha is that the original source site protects the download URL with a login form, so I have had mirror the original package on my website.

Also not sure what the update target is supposed to do...


highperformancecoder:
Thanks  @xantares and @TimothyGu. I might wait until I'm in a particularly masochistic mood before trying then. Until then, users of my Windows build will need to do without python, or use builds made by someone else who has succeeded with standard MinGW.

I now have two packages (SciDaVis and Minsky) that use MXE for the Windows build. MXE is an awesome project.

PR #579, opened 2014-12-08
luabind depends either on lua or on luajit

starius:
Add new configuration option LUA, which
can be "lua" or "luajit" (by default, "lua").

This option is used currently only by package luabind.
Previously package luabind depended on lua.
Luabind built with Lua (5.2 for MXE) is
not binary compatible with LuaJIT (which is
binary compatible to Lua 5.1, not Lua 5.2).
That is why an application using luabind is to use
lua instead of luabind.

Currently the problem is solved.
Compile luabind against luajit:

```
$ make luabind LUA=luajit
```

and link your application against luajit andluabind.

To compile luabind against Lua (previous and currently
default behaviour):

```
$ make luabind LUA=lua
```

or just

```
$ make luabind
```

http://permalink.gmane.org/gmane.comp.gnu.mingw.cross-env/3691

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,10 @@ WGET       := wget --no-check-certificate \
                    --user-agent=$(shell wget --version | \
                    $(SED) -n 's,GNU \(Wget\) \([0-9.]*\).*,\1/\2,p')
 
+# Lua implementation used as dependency of other packages
+# e.g. luabind (possible values: lua or luajit)
+LUA        := lua
+
 REQUIREMENTS := autoconf automake autopoint bash bison bzip2 cmake flex \
                 gcc g++ gperf intltoolize $(LIBTOOL) $(LIBTOOLIZE) \
                 $(MAKE) openssl $(PATCH) $(PERL) python ruby scons \
--- a/src/luabind-1-cmakelists.patch
+++ b/src/luabind-1-cmakelists.patch
@@ -3,16 +3,16 @@ See index.html for further information.
 
 commit 9b4639e25442a3b6d0337d8e602a8332ec0e26e0
 Author: Boris Nagaev <bnagaev@gmail.com>
-Date:   Fri Aug 22 15:17:21 2014 +0400
+Date:   Mon Dec  8 00:28:20 2014 +0300
 
     CMakeLists.txt
 
 diff --git a/CMakeLists.txt b/CMakeLists.txt
 new file mode 100644
-index 0000000..acc47ae
+index 0000000..c0d2627
 --- /dev/null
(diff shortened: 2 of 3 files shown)


TimothyGu:
I don't really like how this is handled. For example, if a person first build with `LUA=lua` and then switched to `LUA=luajit` it wouldn't rebuild.

starius:
How can this be fixed?
Possible solution:
 * save current value of LUA after successful installation
 * compare current value of LUA with saved value (if it exists) and rebuild all packages depending on LUA if these value do not match

Please help me with implementing this.

TimothyGu:
@starius There are no clean ways of implementing it right now. Let me think about it for a day and get back to you.

tonytheodore:
A general solution to this would also be useful for jpeg/libjpeg-turbo.

tonytheodore:
I think I'd rather see something like `lua` being a virtual package that builds either (or both?) packages and installs or "activates" the user-selected one. That way, no changes need to be made to downstream packages.

starius:
@tonytheodore the problem with lua/luajit is that they are not fully compatible. Current Lua version is 5.3, but LuaJIT is compatible with lua 5.1, not lua >= 5.2. If package luabind links with lua and package X links with luabind, then package X has to use (and link) lua, not luajit. And vice versa. Your idea about virtual packages would work for Lua if LuaJIT was 100% compatible with Lua 5.3.

There are macros LUA_COMPAT_* which tell lua to be compatible with previous versions. Maybe they help. I am not sure about binary compatibility.

Maybe we should split package lua to lua5.1, lua5.2 and lua5.3. Lua5.1 has two implementations: lua-puc-5.1 and luajit. People who are satisfied with lua 5.1 or luajit and do not need lua 5.2 and lua 5.3 features, can use lua5.1. Package luabind linked against lua5.1 is binary compatible with both PUC lua 5.1 and luajit. There is still a problem: how to use luabind with lua 5.2 or lua 5.3? Maybe we should provide multiple luabind versions as well: luabind-lua5.1 luabind-lua5.2 luabind-lua5.3. And use this approach for any package depending on lua.

tonytheodore:
Are these two packages always static or have we just not gotten around to enabling shared builds? I ask as I'm trying to understand what should be a target (with built in directory separation and generic common toolchain dependencies) and what should be a user-variant.

To trigger rebuilds with changed options, we can simply take a hash of all relevant flags and append it to the install markers.

starius:
> Are these two packages always static or have we just not gotten around to enabling shared builds?

I think, the latter. On Debian, luabind and luajit are available as normal libraries: .a + .so.

PR #584, opened 2014-12-10
qtbase: Append Qt5 cmake dir to CMAKE_PREFIX_PATH.

uwehermann:
This allows cmake-using programs to find Qt5 components.

### DIFF
--- a/src/qtbase.mk
+++ b/src/qtbase.mk
@@ -76,6 +76,10 @@ define $(PKG)_BUILD
         -I'$(1)/test-$(PKG)-pkgconfig' \
         `'$(TARGET)-pkg-config' Qt5Widgets --cflags --libs`
 
+    # setup cmake toolchain
+    $(SED) -i '/CMAKE_PREFIX_PATH/d' '$(CMAKE_TOOLCHAIN_FILE)'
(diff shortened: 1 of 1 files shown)


TimothyGu:
@uwehermann Are you interested in fixing the `CMAKE_SYSTEM_PREFIX_PATH` thing?

uwehermann:
I'm interested in having a fix, but I probably won't have time to work on anything in the next few days. Go ahead if you intend to fix this!

DrAlexx:
Hi guys I  cought the same problem with qt5.
But I have another solution:
We really don't need to change CMAKE_PREFIX_PATH.
There is a Qt5_DIR variable. It should contains a path that looks like that:
${HOME}/MXE/usr/i686-w64-mingw32.shared/qt5/lib/cmake/Qt5
I put Qt5_DIR  with such path in the toolchain-file and all my builds work fine.

And yet another more. There are a lot of manuals how to use Qt5 via cmake. 
Some examples from these are obsolete and don't work with  Qt5_DIR.
A new and correct way for using Qt5 via cmake is:

find_package(Qt5 REQUIRED Core Concurrent Gui ..... )
target_link_libraries(my_app Qt5::Core Qt5::Concurrent Qt5::Gui ... )

PR #594, opened 2014-12-27
Fix boost

pavelvat:
https://github.com/boostorg/interprocess/pull/15


TimothyGu:
Can you please explain the issue this fixes?

pavelvat:
See https://github.com/boostorg/interprocess/pull/15

BOOST_USE_WINDOWS_H is defined in boost/detail/interlocked.hpp (after apply the patch https://github.com/mxe/mxe/blob/master/src/boost-1-fixes.patch), so #ifdef BOOST_USE_WINDOWS_H always false https://github.com/boostorg/interprocess/blob/master/include/boost/interprocess/detail/win32_api.hpp#L31

TimothyGu:
@pavelvat OK, merged.

PR #597, opened 2015-01-07
mingw-w64: Use --enable-idl so widl can be used.

sergio91pt:
Any attempt to generate headers for an idl that imports a "standard" idl
will fail if the respective idl files are not in the include path.

MingW-W64 copies the idl .h files but not the idl files unless --enable-idl
is used.

### DIFF
--- a/src/mingw-w64.mk
+++ b/src/mingw-w64.mk
@@ -22,7 +22,8 @@ define $(PKG)_BUILD_mingw-w64
     cd '$(1).headers-build' && '$(1)/mingw-w64-headers/configure' \
         --host='$(TARGET)' \
         --prefix='$(PREFIX)/$(TARGET)' \
-        --enable-sdk=all
+        --enable-sdk=all \
+        --enable-idl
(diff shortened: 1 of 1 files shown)


TimothyGu:
Thanks. It would be better if you add `widl` as a dependency of mingw-w64 package.

sergio91pt:
Shouldn't it be `mingw-w64` as a dependency in [widl.mk](https://github.com/mxe/mxe/blob/master/src/widl.mk) instead?

TimothyGu:
@sergio91pt No, because widl does not depend on mingw-w64. It runs on the build machine like GCC.

sergio91pt:
`widl` is only a depency to build mingw-w64-headers if `--enable-widl <PATH>` is set, which rebuilds the header files of all idl files. Those generated .h files are already included in the src tarball and always end up in the include folder, what `--enable-idl` does is simply copy the respective *.idl files to the include folder.

TimothyGu:
Ah, okay.

PR #598, opened 2015-01-07
OpenCV: Fix paths and add shared support

sergio91pt:
Tested with x86_64-w64-mingw32.static and x86_64-w64-mingw32.shared.

Cherrypicks an upstream commit that cleanups the pkg-config file.
Also removes the version postfix on Windows dll's (matching other platforms).

These changes allows us to easilly support shared builds.
Removing the postfix avoids breaking programs that only link
to a subset of opencv modules, everytime a new version is released.

It should be noted that OpenCV forces static linking of libopencv_ts.
There are also 3 files that are left at usr/<triplet>/ that this commit does not
attempt to fix (LICENSE, OpenCVConfig.cmake and OpenCVConfig-version.cmake).

They cannot be moved to lib/ because a different OpenCVConfig.cmake is placed
there. On Ubuntu, for example, those cmake files are in /usr/share/OpenCV/.

### DIFF
--- a/src/opencv-2-fixes-cmake.patch
+++ b/src/opencv-2-fixes-cmake.patch
@@ -1,0 +1,70 @@
+This file is part of MXE.
+See index.html for further information.
+
+diff -ru a/cmake/OpenCVGenConfig.cmake b/cmake/OpenCVGenConfig.cmake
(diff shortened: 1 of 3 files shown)


TimothyGu:
Merged, thanks!

PR #600, opened 2015-01-15
Add libepoxy

TobiX:
This adds libepoxy (and the required xorg-macros), a modern alternative to GLEW.

PS: I noticed that there are many packages with quite similar $(PKG)_UPDATE parts, for example many github projects. Would it be a good idea to create some common templates for these cases?

### DIFF
--- a/index.html
+++ b/index.html
@@ -2522,6 +2522,10 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
         <td class="website"><a href="http://sourceforge.net/projects/xmlwrapp/">xmlwrapp</a></td>
     </tr>
     <tr>
+        <td class="package">xorg-macros</td>
+        <td class="website"><a href="http://cgit.freedesktop.org/xorg/util/macros/">X.org utility macros</a></td>
+    </tr>
+    <tr>
         <td class="package">xvidcore</td>
         <td class="website"><a href="http://www.xvid.org/">xvidcore</a></td>
     </tr>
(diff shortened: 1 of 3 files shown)


TimothyGu:
@TobiX said:

> PS: I noticed that there are many packages with quite similar $(PKG)_UPDATE parts, for example many github projects. Would it be a good idea to create some common templates for these cases?

I think that would help. BTW we already have [`MXE_GET_GITHUB_SHA`](https://github.com/mxe/mxe/blob/master/Makefile#L120-L124).

index.html
```diff
@@ -2522,6 +2526,10 @@ <h2 id="packages">List of Packages</h2>
         <td class="website"><a href="http://sourceforge.net/projects/xmlwrapp/">xmlwrapp</a></td>
     </tr>
     <tr>
+        <td class="package">xorg-macros</td>
+        <td class="website"><a href="http://cgit.freedesktop.org/xorg/util/macros/">xorg-macros</a></td>
```
TimothyGu: X.org utility macros

index.html
```diff
@@ -1634,6 +1634,10 @@ <h2 id="packages">List of Packages</h2>
         <td class="website"><a href="https://dvdnav.mplayerhq.hu/">libdvdread</a></td>
     </tr>
     <tr>
+        <td class="package">libepoxy</td>
+        <td class="website"><a href="https://github.com/anholt/libepoxy">libepoxy/a></td>
```
TimothyGu: Broken `</a>`

TobiX:
That were stupid errors ;) - Updated.

TimothyGu:
Thanks!

PR #601, opened 2015-01-17
Make glib run on WinXP again

TobiX:
Recently, a Performous user notified me that Performous does not run under Windows XP anymore, because the function if_nametoindex was not found in iphlpapi.dll. This happens because glib in MXE is build in a way to use the system if_nametoindex instead of its own replacement.

This branch tries to fix that. Part of the reason why compilation of glib failed is a conflicting definition of if_nametoindex in the mingw-w64 headers and glib sources. I checked with mingw-w64 upstream and their trunk has this fixed: If the API version is lower then Vista, those functions are not exposed and there is no conflict in glib.

Unfortunately, the structure of those header files changed very much in mingw-w64 trunk, so cherry-picking patches is not really possible. Therefore, since this will be fixed eventually, this patch uses the minimal-invasive way: It only protects if_nametoindex and if_indextoname (because I know those are missing on Windows XP) and makes glib compile and run correctly.

This patch can be removed again when mingw-w64 is updated to 4.0...


TimothyGu:
Thanks for the explanation. LGTM.

PR #609, opened 2015-01-25
Add package termcap, and make readline use it instead of pdcurses.

marlam:
This fixes runtime problems with applications using readline.
For example, backspace does not work. This is because
pdcurses, which readline currently uses to get information about a
terminal, does not really implement tgetent() and friends; they are only
stubs. Google "pdcurses readline backspace" for more information.

First I tried using ncurses instead, but then the application crashes.
I did not try to debug.

I simply added libtermcap and made readline use that. Yes, it is
ancient, but it works fine...


TimothyGu:
Having something old is always better than having something broken, so merged, thanks!

PR #610, opened 2015-01-25
Fix 64 bit build of various packages

marlam:
This enables 64 bit builds for a number of packages: dcmtk, portablexdr, hdf4, netcdf, openexr, imagemagick.
The gdal package profits from this: since now hdf4 is available for 32 and 64 bit, the build rules do not have to be separate anymore.

### DIFF
--- a/src/netcdf.mk
+++ b/src/netcdf.mk
@@ -32,14 +32,10 @@ define $(PKG)_BUILD
         --enable-netcdf-4 \
         --enable-hdf4 \
         --prefix='$(PREFIX)/$(TARGET)' \
-        CPPFLAGS="-D_DLGS_H" \
+        CPPFLAGS="-DWIN32_LEAN_AND_MEAN -DH4_BUILT_AS_STATIC_LIB=1" \
         LIBS="-lmfhdf -ldf -lportablexdr -lws2_32"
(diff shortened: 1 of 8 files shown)


src/netcdf.mk
```diff
@@ -32,14 +32,10 @@ define $(PKG)_BUILD
         --enable-netcdf-4 \
         --enable-hdf4 \
         --prefix='$(PREFIX)/$(TARGET)' \
-        CPPFLAGS="-D_DLGS_H" \
+        CPPFLAGS="-DWIN32_LEAN_AND_MEAN -DH4_BUILT_AS_STATIC_LIB=1" \
```
TimothyGu: Woudl be better if you can hardcode `H4_BUILT_AS_STATIC_LIB` in hdf4 so that all apps using hdf4 can benefit from it.

marlam: OK, I split this pull request into to new ones, the first with imagemagick, openexr, dcmtk (should be uncontroversial) and the second one with portablexdr, hdf4, netcdf, gdal (which depend on one another). In the new version, H4_BUILT_AS_STATIC_LIB is added to the hdf4 header h4config.h, which is included by all hdf4 headers.

TimothyGu: Cool thanks.

PR #612, opened 2015-01-25
64bit fixes, part 2: portablexdr, hdf4, netcdf, gdal

marlam:


tonytheodore:
Thanks, these all seem to have been subsequently fixed.

PR #614, opened 2015-01-27
Add fix for Polarssl CVE-2015-1182

AndreHeinecke:
Temporary patch this until a new release of polarssl is made available. See:
https://polarssl.org/tech-updates/security-advisories/polarssl-security-advisory-2014-04


TimothyGu:
Thanks.

PR #622, opened 2015-02-05
Fixes for clean build environment

TobiX:
These are some fixes necessary in a (reasonably) clean build environment. Tested on an fresh Debian wheezy install.

- glib doesn't build without zlib any more, so provide a native version when building native glib tools
- sdl2_gfx tries to run aclocal-1.?? when running make since the timestamps between configure and Makefile.* are strange, so we rebuild everything with the installed automake tools...
- Somehow, sometimes, qtbase decides to link using gold, but gold only works for ELF. This can be fixed by disabling gold altogether...


TimothyGu:
Thanks for the PR. Sorry for forgetting about it.

PR #636, opened 2015-02-19
remove ncurses from libcaca

avih:
With the ffmpeg/caca patch and my earlier commits, mpv ( https://github.com/mpv-player/mpv ) can be built with MXE without any modifications and produce a useful binary.

### DIFF
--- a/src/libcaca.mk
+++ b/src/libcaca.mk
@@ -8,7 +8,7 @@ $(PKG)_CHECKSUM := ed138f3717648692113145b99a80511178548010
 $(PKG)_SUBDIR   := $(PKG)-$($(PKG)_VERSION)
 $(PKG)_FILE     := $(PKG)-$($(PKG)_VERSION).tar.gz
 $(PKG)_URL      := http://caca.zoy.org/raw-attachment/wiki/$(PKG)/$($(PKG)_FILE)
-$(PKG)_DEPS     := gcc freeglut ncurses zlib
+$(PKG)_DEPS     := gcc freeglut zlib
 
 define $(PKG)_UPDATE
     $(WGET) -q -O- 'http://caca.zoy.org/wiki/libcaca' | \
@@ -28,6 +28,7 @@ define $(PKG)_BUILD
         --disable-java \
         --disable-python \
         --disable-ruby \
+        --disable-ncurses \
         --disable-doc
     $(MAKE) -C '$(1)' -j '$(JOBS)'
     $(MAKE) -C '$(1)' -j 1 install


TimothyGu:
See https://github.com/mxe/mxe/pull/636/files#diff-a91d3ebe2abb2bb5617e6b7497e70574R53

You shouldn't need to disable caca because of that line which forces ffmpeg and its libraries to be considered console applications.

I have never heard of any issue with autoconf, so you should consider upgrading your Ubuntu distro.

avih:
This is on Ubuntu 14.04 LTS. When installing also autoconf 2.13 which might be required by some legacy projects (outside of MXE), it also installs the wrapper script to choose the autoconf version, which behaves according to the link at the commit message.

As for caca, IIRC I also tried to fix it without removing it but didn't succeed, but I'll follow your link, try again and post my results once I do.

Does MXE have non-console apps which use ffmpeg libs such that I can test how they behave when launched without a console?

avih:
> See https://github.com/mxe/mxe/pull/636/files#diff-a91d3ebe2abb2bb5617e6b7497e70574R53

What should I be looking at, and how should I use it to make mpv not spawn a console window when it's linked with libcaca?

FWIW, I think the issue is with libcaca (or its .pc file) rather than with ffmpeg.

mpv supports using libcaca output independently of ffmpeg (or rather, it doesn't use ffmpeg's capability to output using libcaca). Here are some results from various combinations of building mpv.

- mpv without caca linked to ffmpeg without caca -> doesn't spawn a console window.
- All other combination where either or both mpv and ffmpeg are have caca enabled do spawn a console window whenever running mpv.

So whenever libcaca is in play, it spawns an undesirable window. My approach to "fix" it was to remove it from ffmpeg since caca is mostly a gimmick (even if a very cute one) and not really a core feature of ffmpeg, and if there are cases where it harms the resulting binary, then I though it's better to have it disabled by default.

So far I don't have another solution for this issue.

avih:
I managed to narrow down the issue to caca built with ncurses. When i modified caca to build with `--disable-ncurses` and then built mpv with caca, it finally behaves as it should:

1. when dragging a video file in explorer onto mpv.exe it only opens a video window.
2. when  running from console it shows the normal console output and only opens a video window.
3. when running with --vo=caca it reuses the existing console.
4. when dragging a video file onto a shortcut to mpv.exe which has --vo=caca it does open a new console window for caca.

Would you accept a patch which disables ncurses for libcaca?

ncurses is optional for caca and caca appears to be able to perfectly reuse an existing console window or open a new console window if launched from explorer - also when ncurses is disabled.

avih:
I also tried to update ncurses to the current latest (c6cfd97), but it didn't fix the issue.

I think something in ncurses automatically spawns a console window if the executable is a GUI application, regardless if it's launched from a command prompt console or even if nothing initializes it explicitly.

I'd still recommend to build libcaca without ncurses. Really, libcaca is mostly a gimmick, and it seems to be fully functional on all cases which I tried also without ncurses.

avih:
I've tested it further, and I'm 99% sure it's a static initialization issue within ncurses - where whenever something in ncurses detects that it runs in a GUI exe, it spawns a console window - even if nothing calls any ncurses function. Just linking with it is enough to cause this bug.

The funny thing is that caca doesn't even use ncurses unless it's defined in `CACA_DRIVER` environment variable (or requested specifically by the caller - which I guess most callers don't). By default caca uses the win32 console "driver" on windows and when built with MXE.

This is the default caca choose order where it chooses the first which it supports (at `caca/caca.c` at `caca_select_driver`): cocoa, win32, conio, vga, x11, gl, ncurses, slang.

I tested different drivers by forcing caca to use them and then used mpv to output with caca:
- win32 (the default) - works fine.
- gl - fails to initialize.
- ncurses - fine if executed from a command prompt window but crashes if executed from a console wrapper (such as ConEmu).

Conclusion: it happens when caca is built with ncurses, and even on that case caca doesn't use ncurses by default (yet still spawns the console window when linked into a GUI exe).

caca doesn't need ncurses, and ffmpeg - which is a proper useful library which only uses caca as a gimmic - definitely doesn't need ncurses, and especially if it has undesirable side effects.

I'll send a PR to remove ncurses from caca dependencies.

avih:
PS.

Thanks to @wm4, we've identified the exact issue within ncurses:

- ncurses has a `__attribute__((constructor))` function at its win32 driver which executes whenever the library is loaded - no need to call any ncurses function in order for it to run.

- This function calls `AllocConsole` unconditionally: https://github.com/mirror/ncurses/blob/master/ncurses/win32con/win_driver.c#L2195 .

- `AllocConsole` spawns a new console window if the executable is a GUI application (or if a console application is launched from explorer): https://msdn.microsoft.com/en-us/library/windows/desktop/ms681944%28v=vs.85%29.aspx

As a result, any GUI application which links with MXE's ffmpeg libraries (which bring caca, which brings ncurses - but still ends up not using ncurses by default) ends up spawning a console window which the user never ask for.

avih:
Yes? no? comment?

wm4:
My 2 cents: building ffmpeg with caca enabled is bullshit. The way ncurses initializes on Windows is bullshit and extremely library-unsafe. Please save us from this nonsense.

avih:
I reported this to ncurses and they fixed it upstream:
- http://lists.gnu.org/archive/html/bug-ncurses/2015-02/msg00017.html
- http://lists.gnu.org/archive/html/bug-ncurses/2015-02/msg00021.html

So with this ncurses version, it no longer opens a new console window. I'll close this PR and open a new one to update ncurses.

I still think caca shouldn't build with ncurses, or even better, ffmpeg shouldn't build with caca. Summary of the reasons:

- even when built with ncurses, caca doesn't use it by default.
- win32 (default) and ncurses as caca backends seem to perform very similarly.
- ncurses as a caca backend crashes more than win32 backend (little testing).
- ncurses brings a relatively big dependency to any executable which links with caca, and since caca is mostly a gimmick output mode, it's not worth it.

wm4:
>I still think caca shouldn't build with ncurses, or even better, ffmpeg shouldn't build with caca,

enable all teh features!!11111

PR #638, opened 2015-02-26
index.html versions cache

TobiX:
Currentl, the index page pulls version information from all the *.mk files, which is quite a lot of data and requests. This introduces a cache for this data:

- A makefile target to generate the cache
- JavaScript code that reads the cache and falls back to the old style, if it isn't found
- A bit unrelated: Shorten long version numbers, like in build-matrix.

On my slow DSL connection, this is the difference for loading index.html:
Without cache: 372 Requests, 779,42 KB, 37,30 s
With cache: 6 Requests, 198,27 KB, 5,11 s

I don't know how useful this is for mxe.cc, but it may be possible to move the GitHub Pages hosting to another branch (gh-pages) and commit versions.json there...

Test page: http://buildbox.23.gs/mxe/#packages

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -507,7 +507,7 @@ show-upstream-deps-%:
 
 .PHONY: clean
 clean:
-	rm -rf $(call TMP_DIR,*) $(PREFIX) build-matrix.html
+	rm -rf $(call TMP_DIR,*) $(PREFIX) build-matrix.html versions.json
 
 .PHONY: clean-pkg
 clean-pkg:
@@ -663,3 +663,12 @@ build-matrix.html: $(foreach PKG,$(PKGS), $(TOP_DIR)/src/$(PKG).mk)
 	@echo '</table>'                        >> $@
 	@echo '</body>'                         >> $@
 	@echo '</html>'                         >> $@
+
+
+versions.json: $(foreach PKG,$(PKGS), $(TOP_DIR)/src/$(PKG).mk)
+	@echo '{'                         > $@
+	@{$(foreach PKG,$(PKGS),          \
+	    echo '    "$(PKG)":           \
+	        "$($(PKG)_VERSION)",';)} >> $@
+	@echo '    "": null'             >> $@
+	@echo '}'                        >> $@
--- a/index.html
+++ b/index.html
@@ -2607,8 +2607,12 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
             request.onreadystatechange = function reqCallback() {
                 if (request.readyState === 4) {
                     if (request.status === 200) {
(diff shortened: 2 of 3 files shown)


index.html
```diff
@@ -2601,6 +2601,21 @@ <h2 id="packages">List of Packages</h2>
                 })();
             }
         }
+        function loadVersionCache(doneCallback, errCallback) {
+            var request = new XMLHttpRequest();
+            request.open('GET', 'versions.json', true);
+            request.onreadystatechange = function reqCallback() {
+                if (request.readyState === 4) {
+                    if (request.status === 200) {
+                        var versions = JSON.parse(request.responseText);
```
TimothyGu: You should probably use a `try {} catch () {}` to fallback on `errCallback`.

Makefile
```diff
@@ -663,3 +663,12 @@ build-matrix.html: $(foreach PKG,$(PKGS), $(TOP_DIR)/src/$(PKG).mk)
...
 	@echo '</body>'                         >> $@
 	@echo '</html>'                         >> $@
+
+
+versions.json: $(foreach PKG,$(PKGS), $(TOP_DIR)/src/$(PKG).mk)
+	@echo '{'                         > $@
+	@{$(foreach PKG,$(PKGS),          \
+	    echo -e '\t "$(PKG)":         \
+	        "$($(PKG)_VERSION)",';)} >> $@
+	@echo -e '\t "": null'           >> $@
```
TimothyGu: Is empty key valid JSON?

Makefile
```diff
@@ -663,3 +663,12 @@ build-matrix.html: $(foreach PKG,$(PKGS), $(TOP_DIR)/src/$(PKG).mk)
 	@echo '</table>'                        >> $@
 	@echo '</body>'                         >> $@
 	@echo '</html>'                         >> $@
+
+
+versions.json: $(foreach PKG,$(PKGS), $(TOP_DIR)/src/$(PKG).mk)
+	@echo '{'                         > $@
+	@{$(foreach PKG,$(PKGS),          \
+	    echo -e '\t "$(PKG)":         \
```
TimothyGu: I'd rather use four spaces instead of `\t `

TimothyGu:
In general I like this idea. Note that mxe.cc is already hosted on GitHub Pages (see the `CNAME` file) on gh-pages branch.

Makefile
TobiX: It is. This is just an ugly workaround because the last comma would produce a syntax error instead... If you have a better idea...

TobiX:
Rebased to address tab-vs-space concerns ;)

TimothyGu:
Thanks!

PR #644, opened 2015-03-05
libcdio update to 0.93, and new package libcdio-paranoia

avih:


avih:
Force pushed the libcdio-paranoia patch - forgot to add the `index.html` updates.

PR #646, opened 2015-03-09
Jansson package

Mireiawen:
Hello, 
I have created a simple mxe package for Jansson JSON library; http://www.digip.org/jansson/ 
There is no PKG_UPDATE or test program for it. I'm not actually writing the code against the library myself, just building other code. So I'm not having good idea of the test program.


TimothyGu:
Thanks!

TimothyGu:
Actually this recipe looks very wrong. Is this supposed to be run on Windows or Linux? The current recipe is for the building machine, because it doesn't set host triplet or library mode properly.

TimothyGu:
The wrong*est* parts are fixed in ce68a9e850bd536f6b3ff040dd11030c52c39cfe. You should really test to see if it works.

PR #648, opened 2015-03-09
Updated mman-win32

0xACE:
This updates the the checksum and version to the latest commit...

If you're not willing to change to the latest commit, you could change the URL to:

    https://github.com/witwall/mman-win32/archive/3421c28e753c38d24a2e27c111b1c9b4601ebe7d.zip

Rather than keeping:

    https://github.com/witwall/mman-win32/archive/master.tar.gz

That master.tar.gz is probably always pointing at the latest commit.

### DIFF
--- a/src/mman-win32.mk
+++ b/src/mman-win32.mk
@@ -3,8 +3,8 @@
 
 PKG             := mman-win32
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 3421c28e753c38d24a2e27c111b1c9b4601ebe7d
-$(PKG)_CHECKSUM := c33e84043d49d0e33bc434bda3a16ce60432e789
+$(PKG)_VERSION  := 378ed6b69bb7220511dd9cd0973c22b3f6773ce7
+$(PKG)_CHECKSUM := f2e54393f530b35d1736fa98a5a29d7d3a0ce76b
 $(PKG)_SUBDIR   := mman-win32-master
 $(PKG)_FILE     := mman-win32-$($(PKG)_VERSION).tar.gz
 $(PKG)_URL      := https://github.com/witwall/mman-win32/archive/master.tar.gz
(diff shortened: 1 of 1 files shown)


TimothyGu:
Yes it would be best to change the URL. Thanks.

0xACE:
>Yes it would be best to change the URL. Thanks.
Alright. I reverted the changes made. Now it will download that specific commit.

TimothyGu:
No what I meant is updating *and* changing the URL.

src/mman-win32.mk
```diff
@@ -7,7 +7,7 @@ $(PKG)_VERSION  := 3421c28e753c38d24a2e27c111b1c9b4601ebe7d
 $(PKG)_CHECKSUM := c33e84043d49d0e33bc434bda3a16ce60432e789
 $(PKG)_SUBDIR   := mman-win32-master
 $(PKG)_FILE     := mman-win32-$($(PKG)_VERSION).tar.gz
-$(PKG)_URL      := https://github.com/witwall/mman-win32/archive/master.tar.gz
+$(PKG)_URL      := https://github.com/witwall/mman-win32/archive/3421c28e753c38d24a2e27c111b1c9b4601ebe7d.zip
```
TimothyGu: You should use the `$(PKG)_VERSION` variable here.

src/mman-win32.mk
```diff
@@ -7,7 +7,7 @@ $(PKG)_VERSION  := 3421c28e753c38d24a2e27c111b1c9b4601ebe7d
...
 $(PKG)_DEPS     := gcc
 
 define $(PKG)_UPDATE
```
TimothyGu: You might find `MXE_GET_GITHUB_SHA` useful. See how it's used: https://github.com/mxe/mxe/blob/4a758acf46f140ad50fb1114d16dfc01f313cd1b/src/librtmp.mk#L13

0xACE:
Thank you for your patience.

I've addressed your comments and pushed a new version.

TimothyGu:
Thank *you* for your contribution!

PR #649, opened 2015-03-10
Implementation for MXE_GET_GITHUB_TAG_SHA

0xACE:
This is identical to MXE_GET_GITHUB_SHA. This will grab the SHA based on the latest tag.

Example of usage:

    $(PKG)_UPDATE = $(call MXE_GET_GITHUB_TAG_SHA, owner/repository)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,6 +123,12 @@ define MXE_GET_GITHUB_SHA
     head -1
 endef
 
+define MXE_GET_GITHUB_TAG_SHA
+    $(WGET) -q -O- 'https://api.github.com/repos/$(strip $(1))/git/refs/tags/' | \
+    $(SED) -n 's#.*"sha": "\([^"]\{10\}\).*#\1#p' | \
+    tail -1
+endef
+
 # use a minimal whitelist of safe environment variables
 ENV_WHITELIST := PATH LANG MAKE% MXE% %PROXY %proxy
 unexport $(filter-out $(ENV_WHITELIST),$(shell env | cut -d '=' -f1))


TimothyGu:
Instead of printing the SHA-1, it is probably better to print the tag name instead, and download the tag tarball.

TimothyGu:
Can you test if the following works? If it does I'll commit my version, as it allows further post-processing of the version name.

Also, it is not always the best idea to choose the last tag, as it might not be the latest (e.g. `2.6-rc` is sorted after `2.6`).

```make
define MXE_GET_GITHUB_TAGS
    $(WGET) -q -O- 'https://api.github.com/repos/$(strip $(1))/git/refs/tags/' | \
    $(SED) -n 's#.*"ref": "refs/tags/\([^"]*\).*#\1#p' | \
    $(SORT) -V
endef
```

```make
define $(PKG)_UPDATE
    $(call MXE_GET_GITHUB_TAGS, akheron/jansson) | \
    $(SED) 's,^v,,g'
endef
```

(This won't work if the tag name contains `"`, but meh who cares at this point.)

0xACE:
>Also, it is not always the best idea to choose the last tag

My bad, you're right. Nice catch.

>Can you test if the following works? If it does I'll commit my version, as it allows further post-processing of the version name.

Yes, I tried it and it works.

The reason i chose the hash rather than the tagname was because i was worried someone would have weird tag names, not sure how this project handles spaces in filenames. As you pointed out tags containing `"` will cause issues. I thought about removing all but `[0-9.\-_]` but that also seemed like a poor hack as I imagined someone would use letters in the version number.

We could leave it to the person implementing `PKG_UPDATE` to create a sed pattern to deal with anamolies in tag names.

I'll leave the final decision to you.

starius:
MXE_GET_GITHUB_TAGS from https://github.com/mxe/mxe/pull/649#issuecomment-77987449 doesn't work for pire.
It joins multiple versions into one string.

Makefile:
```Makefile
define MXE_GET_GITHUB_TAGS
    $(WGET) -q -O- 'https://api.github.com/repos/$(strip $(1))/git/refs/tags/' | \
    $(SED) -n 's#.*"ref": "refs/tags/\([^"]*\).*#\1#p' | \
    $(SORT) -V
endef
```

src/pire.mk:
```Makefile
define $(PKG)_UPDATE
    $(call MXE_GET_GITHUB_TAGS, yandex/pire) | \
    $(SED) 's,^release-,,g'
endef
```

Change current version to some old version (e.g., 0.0.4) and try to update:
```bash
$ make update-package-pire
[using autodetected 4 job(s)]
OLD      pire  0.04 --> 0.0.3 0.0.4 0.0.5 ignoring
make: `update-package-pire' is up to date
```

TimothyGu:
Oh yeah I forgot the final `tail -1`

This should work:

```make
define MXE_GET_GITHUB_TAGS
    $(WGET) -q -O- 'https://api.github.com/repos/$(strip $(1))/git/refs/tags/' | \
    $(SED) -n 's#.*"ref": "refs/tags/\([^"]*\).*#\1#p' | \
    $(SORT) -V |
    tail -1
endef
```

starius:
I have created a pull request for this: #888

starius:
I think, this pull-request can be closed now, as #888 was merged.

PR #650, opened 2015-03-10
PKG_UPDATE for jansson.mk

0xACE:
I noticed how the author for jansson.mk in #646 mentioned the lack for resolving the SHA for the PKG_UPDATE. This patch along with #649 fixes this problem.

A side effect from this is that PKG_VERSION will have to be replaced by a checksum so that it resolves the correct github url. I'll leave it up to you how to deal with this.

### DIFF
--- a/src/jansson.mk
+++ b/src/jansson.mk
@@ -3,17 +3,14 @@
 
 PKG             := jansson
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 2.7
-$(PKG)_CHECKSUM := 51cd3b57f152132de686bb8c1f4471cc9ca7970e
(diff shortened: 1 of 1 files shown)


TimothyGu:
For this specific case, I think it is best not to use GitHub downloads. GitHub tarballs don't have proper configure scripts generated if the project doesn't check them in, plus the project has an official website.

TimothyGu:
I pushed a fix that checks its website. Can you check if it works? Thanks

0xACE:
I tried your patch:

    NEW      jansson  2.6 --> 2.7

It seems like it detected the correct version.

Good work!

PR #660, opened 2015-03-22
Fix compilation of GLEW

GuilhermeBernal:
1. pkg-config wrongly adds the flag `-I/usr/include`, creating conflict with system headers.
2. cflags contains a define on static builds, so it must come before sources when compiling the test.


TimothyGu:
Thanks!

PR #667, opened 2015-04-07
ffmpeg: update to 2.6.2

avih:


avih:
Rebased and updated ffmpeg to a newer version

PR #673, opened 2015-04-27
miniupnpc

BioMike:
This is an initial build script and patch for miniupnpc (used by many cryptocurrency wallets).

I made this to setup a buildchain for the new Windows wallet of Auroracoin.


src/miniupnpc.mk
```diff
@@ -0,0 +1,42 @@
...
+define $(PKG)_BUILD
+    cd '$(1)' && $(SH) ./updateminiupnpcstrings.sh
+    $(MAKE) -C '$(1)' -f Makefile.mingw -j '$(JOBS)' \
+        CC='$(TARGET)-gcc' \
+        AR='$(TARGET)-ar' \
+        RANLIB='$(TARGET)-ranlib' \
+        DLLWRAP='$(TARGET)-dllwrap' \
+        init libminiupnpc.a
+    $(INSTALL) -d '$(PREFIX)/$(TARGET)/lib'
+    $(INSTALL) -m644 '$(1)/libminiupnpc.a' '$(PREFIX)/$(TARGET)/lib/'
```
TimothyGu: So only static lib is supported? If this is the case, you'll need:

```make
$(PKG)_BUILD_SHARED =
```

to mark shared as not supported.

BioMike:
Shared  build is a DLL? I inactivated that as i though it was to no use for MXE.

TimothyGu:
> Shared build is a DLL? I inactivated that as i though it was to no use for MXE.

Yes. MXE does support shared build for about half of the packages.

BioMike:
Ok, in what place should I install the DLL then? In '$(PREFIX)/$(TARGET)/lib/' as well?

TimothyGu:
See https://github.com/mxe/mxe/blob/840c9a2435c08c7acd5f62527cc2d13794143c01/src/glfw3.mk#L33-L36

the dll is in bin/, the import lib is in lib/.

https://github.com/mxe/mxe/blob/master/tools/make-shared-from-static might be of interest if the build system doesn't support shared libs.

BioMike:
Ok, dll and lib files are now build installed as well.

BioMike:
Anything else holding this back?

tonytheodore:
The entry in `index.html` is missing.

Also, could you separate the build so the static and shared libraries are installed only for their respective targets?

BioMike:
I'll try to fix it asap. Could take a few days though.

tonytheodore:
Thanks!

[Rebased](https://github.com/mxe/mxe/commit/7696a567bf17c9c35cf162cef5905eadc99c9083) and [switched to cmake with updater](https://github.com/mxe/mxe/commit/eb20723302ed656122ce326129c1347e03b0460d).

PR #678, opened 2015-05-04
debian: Add dependency for libtool-bin

thewtex:
In Debian Jessie, the libtool binary was separated out
into the libtool-bin package:

  https://packages.debian.org/jessie/libtool-bin

### DIFF
--- a/index.html
+++ b/index.html
@@ -633,7 +633,7 @@ USE_OSGPLUGIN(&lt;plugin2&gt;)
     <!-- http://www.debian.org/distrib/packages#search_packages -->
     <pre>apt-get install \
     autoconf automake autopoint bash bison bzip2 cmake flex gettext \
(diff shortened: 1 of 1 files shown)


TimothyGu:
Can you please check which distributions don't have libtool-bin (include Mint and Ubuntu versions)?

thewtex:
Ubuntu has the libtool-bin package in utopic and vivid:

  http://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=libtool-bin&searchon=names

Linux Mint is based on Ubuntu Trusty, so I assume it does not have the libtool-bin package:

  http://www.linuxmint.com/oldreleases.php

starius:
This issue was fixed by 26ac245ffa69856248e519fde184c162f89b2f63

PR #691, opened 2015-05-21
Windows/MinGW extension: opening std::fstream with a wstring/wchar_t*

Ronald-Liu:
> I have implemented a small patch that mirrors Microsoft's STL extension, where one can use wide strings as an argument to std::(w)fstream. This is very useful on Windows, where, without this
extension, there is no way to open an fstream for a filename containing non-ASCII characters not in the Windows local codepage.

https://gcc.gnu.org/ml/libstdc++/2011-06/msg00066.html

I have changed the patch so that it can be used in this version of libstdc++.

### DIFF
--- a/src/gcc-3-fstream.patch
+++ b/src/gcc-3-fstream.patch
@@ -1,0 +1,605 @@
+This file is part of MXE.
+See index.html for further information.
+
+This patch has been taken from:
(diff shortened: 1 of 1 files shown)


TimothyGu:
I'm sorry but I'd like to close this issue. I don't feel like maintaining code that is rejected by upstream.

PR #698, opened 2015-06-04
Add Devil patch

saiarcot895:
Devil attempts to use the `restrict` keyword, even when building in C++ mode. Technically, `restrict` isn't defined in the C++ standard.


TimothyGu:
Thanks!

PR #709, opened 2015-06-13
Fix building dcmtk on 64-bit platforms.

saiarcot895:
The fix for building 64-bit dcmtk in #611 was to ignore the cast issue (an 8-byte pointer was being cast to a 4-bit integer). The proper fix is to fix the underlying issue.

In this case, the errors are in print-statements, so there should be no functionality change.


TimothyGu:
Thanks! You should consider submitting this patch to dcmtk upstream if they are still alive.

PR #710, opened 2015-06-13
Don't use git when building libgcrypt.

saiarcot895:
When reconfiguring libgcrypt using `autoconf`, libgcrypt calls `git` to determine what revision hash and number it's on. Most people won't see an issue, because the build directory is, by default, in a subdirectory of the mxe repo and `git` will use mxe's revision log and hashes. However, if the build directory is changed to `/dev/shm` for I/O performance reasons, then because there is no git repo there, the build will fail.

Specifically, `autoconf` will go through and ignore the error, but when building `versioninfo.rc` using `windres`, the version number will be listed as `1,6,3,` because revision number (hard-coded to 4091 below) will be blank, resulting in a syntax error.

### DIFF
--- a/src/libgcrypt-2-no-git.patch
+++ b/src/libgcrypt-2-no-git.patch
@@ -1,0 +1,34 @@
+This file is part of MXE.
+See index.html for further information.
+
+Contains ad hoc patches for cross building.
+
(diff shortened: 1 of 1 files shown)


TimothyGu:
Thanks!

PR #725, opened 2015-06-19
updated package of gd to 2.1.1 / include patch for deprecated defines

luckyhacky:


TimothyGu:
Does http://lists.nongnu.org/archive/html/mingw-cross-env-list/2015-06/msg00005.html fix the issue? It's a lot cleaner than this PR.

luckyhacky:
There are three options:
- my fix: use old defines
- fix mentioned above: use new defines
- use git repo with new code (newer than 2.1.1 - but not tagged)

1 and 2 do the same
3 uses the newest code but I don't know anything about stability.

I'm fine with all three.

TimothyGu:
I prefer actually fixing the code by using the new defines, especially since it is the fixed preferred by upstream. I'll merge the ML patch tomorrow.

TimothyGu:
Obsoleted by 9816446b2eee843f0a919ab9531d0381913ed8da.

PR #726, opened 2015-06-21
Ossim

bugbrains:
adding ossim


src/cmake/FindTIFF.cmake
```diff
@@ -1,13 +1,16 @@
...
-set(TIFF_INCLUDE_DIR ${SDL_INCLUDE_DIRS})
-set(TIFF_VERSION_STRING ${SDL_VERSION})
+INCLUDE(FindPackageHandleStandardArgs)
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(TIFF  DEFAULT_MSG  TIFF_LIBRARY  TIFF_INCLUDE_DIR)
+
+IF(TIFF_FOUND)
+  SET( TIFF_LIBRARIES ${TIFF_LIBRARY} )
+ENDIF(TIFF_FOUND)
+
+MARK_AS_ADVANCED(TIFF_INCLUDE_DIR TIFF_LIBRARY)
```
TimothyGu: This is wrong. The point of this file is to use pkg-config to bring in dependencies that are needed for static library linking.

bugbrains: pkgconfig is picking up my system library and without pkgconfig_executable in cmake, It gives me error

tonytheodore:
Can you use the OpenThreads that OpenSceneGraph provides? It seems to be a lot newer and has this [openthreads.pc.in issue fixed seven years ago](http://trac.openscenegraph.org/projects/osg/changeset/9700).

Also FindTiff.cmake has [subsequently been fixed](https://github.com/mxe/mxe/commit/e2cb6a302050a4377c65efc3159fec1e0b54e635) though I doubt it was the heart of the issue.

tonytheodore:
@bugbrains are these packages in working order on your system?

rashadkm:
@tonytheodore ,  do you mean to use openscenegraph as a dependency instead of openthreads?

tonytheodore:
> do you mean to use openscenegraph as a dependency instead of openthreads?

The simple answer is yes. Another way would be to use the source tarball from openscenegraph and only build the openthreads component (see bfd, gendef, and widl for examples of this approach).

> Also I am not clear what you mean by in working order. Can you explain a bit more?

I meant - do they actually build and produce the expected output? I ended up only [adding ossim](https://github.com/mxe/mxe/commit/7167339f483f62d92c638d2a17d5865a99888d11) with [various fixups](https://github.com/mxe/mxe/commit/b39c53e4e4f9cbdbd1711a73fe33cb07ec2d2793).

PR #737, opened 2015-06-23
add tool build-pkg.lua

starius:
This tool builds all packages for all targets and generates `*.tar.xz` and `*.deb` files. The latter ones can be uploaded to APT repository and installed to Debian machines.

build-pkg.lua was previously known as mxedeb.lua

See also:
 * http://mxe.redjohn.tk/
 * https://github.com/starius/mxedeb
 * https://lists.nongnu.org/archive/html/mingw-cross-env-list/2015-02/msg00010.html
 * https://lists.nongnu.org/archive/html/mingw-cross-env-list/2015-03/msg00001.html

### DIFF
--- a/tools/build-pkg.lua
+++ b/tools/build-pkg.lua
@@ -1,0 +1,356 @@
+#!/usr/bin/env lua
+
+-- This file is part of MXE.
+-- See index.html for further information.
+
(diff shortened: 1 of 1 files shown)


vog:
Great work!

TimothyGu:
:+1:

PR #750, opened 2015-07-04
GCC: Export codecvt_byname and codecvt symbols for mingw32.

hamzaalloush:
exports symbols neccessary for any application declaring the codecvt_byname facet, which affects Boost in MXE.

it has been fixed in GCC 5.2


TimothyGu:
Thank you!

PR #762, opened 2015-07-15
Add package libjpeg-turbo

tehnick:


TimothyGu:
Thanks! Next time you don't need to add the version.json entry – it's added automatically upon push to the repo.

PR #766, opened 2015-07-19
Add qhttpengine library.

nathan-osman:
QHttpEngine is a small, lightweight library that provides a set of classes for creating HTTP servers. It is based on Qt 5 and will be used in the next version of [NitroShare](https://github.com/nitroshare/nitroshare-desktop) to provide a local API.

The library can be built both as a static and shared library and I have tested both. Because the library uses the CMake `find_package()` command, it is necessary to provide CMake with the path to `Qt5NetworkConfig.cmake`. I have done this via:

    -DCMAKE_PREFIX_PATH='$(PREFIX)/$(TARGET)/qt5/lib/cmake'

### DIFF
--- a/index.html
+++ b/index.html
@@ -2122,6 +2122,10 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
         <td class="website"><a href="http://fallabs.com/qdbm/">QDBM</a></td>
     </tr>
     <tr>
+        <td class="package">qhttpengine</td>
+        <td class="website"><a href="https://github.com/nitroshare/qhttpengine">qhttpengine</a></td>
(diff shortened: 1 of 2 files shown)


TimothyGu:
Would you mind rebasing your patch against the current master? I've committed 9407cab3f312fdf77a998249e247ea468a430b12 which should hopefully make the `CMAKE_PREFIX_PATH` definition extraneous.

nathan-osman:
Okay, I've rebased the commit and pushed it. I also removed the `CMAKE_PREFIX_PATH` line which is, of course, no longer necessary.

TimothyGu:
Cool!

PR #775, opened 2015-07-27
add package pire

starius:
PIRE is Perl Incompatible Regular Expressions library.
This library is aimed at checking a huge amount of text against
relatively many regular expressions. Roughly speaking, it can
just check whether given text maches the certain regexp, but
can do it really fast (more than 400 MB/s on our hardware is
common). Even more, multiple regexps can be combined together,
giving capability to check the text against apx. 10 regexps in
a single pass (and mantaining the same speed).

Pire was developed in Yandex (http://company.yandex.ru/) as a
part of its web crawler.

Homepage: https://github.com/yandex/pire

### DIFF
--- a/src/pire.mk
+++ b/src/pire.mk
@@ -1,0 +1,21 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := pire
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 0.0.5
+$(PKG)_CHECKSUM := fc5f451043c6fc1034f104463a4b9c6c1c46c00c
+$(PKG)_SUBDIR   := pire-release-$($(PKG)_VERSION)
+$(PKG)_FILE     := pire-$($(PKG)_VERSION).tar.gz
+$(PKG)_URL      := https://github.com/yandex/pire/archive/release-$($(PKG)_VERSION).tar.gz
+$(PKG)_DEPS     := gcc
+
+define $(PKG)_BUILD
+    cd '$(1)' && autoreconf -fi -I'$(PREFIX)/$(TARGET)/share/aclocal'
+    cd '$(1)' && ac_cv_func_malloc_0_nonnull=yes ./configure \
+        $(MXE_CONFIGURE_OPTS) \
+        --enable-extra \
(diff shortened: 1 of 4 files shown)


src/pire.mk
```diff
@@ -0,0 +1,40 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := pire
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 0.0.5
+$(PKG)_SOVERS   := 0
```
TimothyGu: This is not needed

src/pire.mk
```diff
@@ -0,0 +1,40 @@
...
+$(PKG)_CHECKSUM := fc5f451043c6fc1034f104463a4b9c6c1c46c00c
+$(PKG)_SUBDIR   := pire-release-$($(PKG)_VERSION)
+$(PKG)_FILE     := release-$($(PKG)_VERSION).tar.gz
```
TimothyGu: Please use a more descriptive name like `pire-$($(PKG)_VERSION).tar.gz` (and change the URL accordingly)

src/pire.mk
```diff
@@ -0,0 +1,40 @@
...
+        --host='$(TARGET)' \
+        --build="`config.guess`" \
+        --enable-extra \
+        --enable-shared \
+        --disable-static \
+        --prefix='$(PREFIX)/$(TARGET)' \
+        PKG_CONFIG='$(PREFIX)/bin/$(TARGET)-pkg-config'
+    $(MAKE) -C '$(1)' -j '$(JOBS)'
+    $(MAKE) -C '$(1)' -j 1 install
+endef
```
TimothyGu: These two macros can be combined with the use of MXE_CONFIGURE_OPTS

src/pire.mk
starius: This URL was [produced](https://github.com/yandex/pire/releases) by Github: `https://github.com/yandex/pire/archive/release-0.0.5.tar.gz` for tag `release-0.0.5`.

TimothyGu: I understand. `$(PKG)_FILE` is the file to save to, not the URL file.

starius:
Fixed.

PR #776, opened 2015-07-27
Software updates

TobiX:
Updates:
- gcc to 5.2
- OpenEXR to 2.2.0
(I tried to fix OpenCV, but gave up :()


TobiX:
Rebased, since gcc was updated by 7882944eb99fa14bb82f3045896668a68f8c68a7

PR #787, opened 2015-08-11
update ocaml-native to 4.02.3

starius:
See https://lists.nongnu.org/archive/html/mingw-cross-env-list/2015-08/msg00003.html

### DIFF
--- a/src/ocaml-native-1-fixes.patch
+++ b/src/ocaml-native-1-fixes.patch
@@ -3,18 +3,17 @@ See index.html for further information.
 
 Contains ad hoc patches for cross building.
 
-From 3a3ddccba8fa90a42551fb4d529c921c0aac4f63 Mon Sep 17 00:00:00 2001
-From: MXE
-Date: Wed, 3 Oct 2012 09:25:11 +0200
-Subject: [PATCH] ocamlbuild : support for prefixed
- ocaml{c,opt,mklib,mktop,find} tools
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Boris Nagaev <bnagaev@gmail.com>
+Date: Wed, 12 Aug 2015 00:00:57 +0200
+Subject: [PATCH] ocamlbuild : support for prefixed ocaml tools
 
 
 diff --git a/ocamlbuild/findlib.ml b/ocamlbuild/findlib.ml
-index b5ef878..77454ed 100644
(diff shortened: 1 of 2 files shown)


src/ocaml-native-1-fixes.patch
```diff
@@ -24,78 +27,70 @@ index b5ef878..77454ed 100644
...
-+	   (ocamlmktop,"ocamlmktop")]
-   end;
- 
-   let reorder x y = x := !x @ (List.concat (List.rev !y)) in
+@@ -232,7 +235,7 @@ let spec = ref (
+    "-install-lib-dir", Set_string Ocamlbuild_where.libdir, "<path> Set the install library directory";
+    "-install-bin-dir", Set_string Ocamlbuild_where.bindir, "<path> Set the install binary directory";
+    "-where", Unit (fun () -> print_endline !Ocamlbuild_where.libdir; raise Exit_OK), " Display the install library directory";
+-   "-which", String (fun cmd -> print_endline (find_tool cmd); raise Exit_OK), "<command> Display path to the tool command";
++   "-which", String (fun cmd -> print_endline (find_tool cmd "FIXME"); raise Exit_OK), "<command> Display path to the tool command";
```
starius: Please review this line carefully. Is it OK to replace extra argument with string "FIXME"?

starius:
Package ocaml-core must be updated as well. It shares ocaml version with ocaml-native.

src/ocaml-native-1-fixes.patch
tonytheodore: I think so, I don't know anything about ocaml but an updated version with minor issues is better than an outdated version.

tonytheodore:
From #1003, this should also be moved to the native target so it's built only once. It's [building on OSX now](https://github.com/mxe/mxe/commit/a96fcf41143baa5aaf3c0e31a9fe204089fad3d9) so will be easier for me to test.

tonytheodore:
Also, project has moved:

http://ocaml.org/
https://github.com/ocaml/ocaml

I wonder if we should make this a plugin?

P.S. @starius, could you update the checksum on this and rebase?

starius:
Rebased, updated checksum, patch consolidated.

It doesn't work. I failed to update crosscompiled ocaml. Somebody who knows ocaml should look at it.

> I wonder if we should make this a plugin?

Why?

tonytheodore:
See #2512

PR #790, opened 2015-08-14
add package libid3tag

pavelvat:
problems:
1)libid3tag.la files builds even for static compilers.
2)option LDFLAGS='-no-undefined' enable build libid3tag-0.dll only for i686-w64-mingw32.shared, not for x86_64-w64-mingw32.shared, log for x86_64-w64-mingw32.shared: https://gist.github.com/pavelvat/be9a6958b390aa1908d7

### DIFF
--- a/index.html
+++ b/index.html
@@ -1665,6 +1665,10 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
         <td class="website"><a href="https://www.gnu.org/software/libiconv/">libiconv</a></td>
     </tr>
     <tr>
+        <td class="package">libid3tag</td>
+        <td class="website"><a href="http://sourceforge.net/projects/mad/files/libid3tag/">libid3tag</a></td>
+    </tr>
+    <tr>
         <td class="package">libidn</td>
         <td class="website"><a href="https://www.gnu.org/software/libidn/">Libidn</a></td>
     </tr>
--- a/src/libid3tag.mk
+++ b/src/libid3tag.mk
@@ -1,0 +1,23 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := libid3tag
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 0.15.1b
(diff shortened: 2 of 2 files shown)


ooxi:
I confirmed that your commit does build. Maybe you could add an `libid3tag-test.c` so others can verify the behaviour, too?

Nevertheless :+1: for merging.

pavelvat:
>I confirmed that your commit does build.

dll file don't build for target "x86_64-w64-mingw32.shared"

>Maybe you could add an libid3tag-test.c so others can verify the behaviour, too?

No.

tonytheodore:
> dll file don't build for target "x86_64-w64-mingw32.shared"

From the log:
> with libz and none of the candidates passed a file format test

It looks like this could be similar to [exiv2](https://github.com/mxe/mxe/blob/master/src/exiv2.mk) or [portaudio](https://github.com/mxe/mxe/blob/master/src/portaudio.mk) that have libtool cache variables set to ensure the file format test passes and the shared lib can be built.

>> Maybe you could add an libid3tag-test.c so others can verify the behaviour, too?

> No.

Tests are invaluable to catch linking issues early on and would be worth it just for that.

They are also instructional and provide guidance on how a library is meant to be used. I just had to [write a cmake-based test for `boost`](https://github.com/mxe/mxe/commit/8a2cbf2c289b46e2f3458d895f021ea79ac04d37) and without the original test to work from, would have had a much harder time of it and possibly given up.

Now we have two examples of how to use `boost` - I hope you reconsider adding a test.

pavelvat:
>It looks like this could be similar to exiv2 or portaudio that have libtool cache variables set to ensure the file format test passes and the shared lib can be built.

Thanks.

tonytheodore:
Thanks, merged in https://github.com/mxe/mxe/commit/60e944f6dedb7eba9cd5b05a752dc6df580e9199

Looks like it was a different libtool issue https://github.com/mxe/mxe/commit/a92b70622596cea7da193be9d3a9ba2cd6fa2b63, both shared targets are working now.

PR #791, opened 2015-08-19
new package berkeleydb

mezzovide:
tested. im not sure with build process, but i can use it.


TimothyGu:
I'd use `db` as the package name rather than `berkeleydb`

mezzovide:
done.

starius:
Timothy, why `db`?
It is usually called `Berkeley DB` or `BDB`, but not `db`.
`db` is too common name to delegate it to this particular database.

TimothyGu:
@starius said:
> `db` is too common name to delegate it to this particular database.

It's called `db` in Debian and in its tarball.

PR #803, opened 2015-08-27
Add plib shared and 64-bit builds

saiarcot895:
Add the ability to build plib for the 64-bit configuration and a shared build configuration.

### DIFF
--- a/src/plib-1-fixes.patch
+++ b/src/plib-1-fixes.patch
@@ -1,0 +1,42 @@
+This file is part of MXE.
+See index.html for further information.
+
+Contains ad hoc patches for cross building.
+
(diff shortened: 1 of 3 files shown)


starius:
Tested. Compiles for me for all 4 targets.
However file `test-plibc.exe` exists for `i686-w64-mingw32.static` only. Can you fix this?

saiarcot895:
Is that supposed to be modified as well? In building Qt, the `plibc` package wasn't built.

starius:
Sorry, wrong alert. File `test-plibc.exe` is a part of package `plibc`, not `plib`.

tonytheodore:
Thanks! Building on all four targets and installing shared libs correctly.

Issue #806, opened 2015-08-27
ncurses under OSX 10.10 does not build (fix proposal)


eruffaldi:
The build of ncurses crashes because during the installation phase the following command is executed:

 $(MAKE) -C '$(1)' -j 1 install TIC_PATH='$(1).native/progs/tic'

The TIC_PATH... is then passed as MAKEFLAGS and then in the ncurses Makefile it is used prepending a "-" because this is the standard behavior.

TOP_MFLAGS = -${MAKEFLAGS} DESTDIR="$(DESTDIR)" RPATH_LIST="$(RPATH_LIST)"

This result in invoking make with "-TIC_PATH" and then the OSX makefile refuses to build.

Solution: simply modify ncurses.mk adding a spurious -s before TIC_PATH. The silenti is a nop

I can make a Pull Request if needed.

PR #807, opened 2015-08-27
fixed ncurses bug in Makefile when specifying TIC_PATH in install (Issue #806)

eruffaldi:


src/ncurses.mk
```diff
@@ -50,6 +50,6 @@ define $(PKG)_BUILD
             --with-normal    --without-shared --with-static, \
             --without-normal --without-static --with-shared)
     $(MAKE) -C '$(1)' -j '$(JOBS)'
-    $(MAKE) -C '$(1)' -j 1 install TIC_PATH='$(1).native/progs/tic'
+    $(MAKE) -C '$(1)' -j 1 install -s TIC_PATH='$(1).native/progs/tic'
```
TimothyGu: Does this work:

```make
TIC_PATH='$(1).native/progs/tic' $(MAKE) -C '$(1)' -j 1 install
```

eruffaldi: Yes. This is more clean

tonytheodore: I can see the same issue on OSX 10.10 and while both of these allow the build to continue, I suspect they only do so if `tic` is already available.

tonytheodore:
Fixed in https://github.com/mxe/mxe/commit/a6aeeb03cfc0317233cd63621a62775bf7785341

PR #808, opened 2015-08-28
add package tor

starius:
Tor is free software and an open network that helps you defend
against traffic analysis, a form of network surveillance that
threatens personal freedom and privacy, confidential business
activities and relationships, and state security.

https://www.torproject.org/

### DIFF
--- a/src/tor.mk
+++ b/src/tor.mk
@@ -1,0 +1,27 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := tor
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 0.2.6.10
+$(PKG)_CHECKSUM := 4bf9689b311efce70db96c88d1f466caa41e0277
+$(PKG)_SUBDIR   := $(PKG)-$($(PKG)_VERSION)
+$(PKG)_FILE     := $(PKG)-$($(PKG)_VERSION).tar.gz
+$(PKG)_URL      := https://torproject.org/dist/$($(PKG)_FILE)
+$(PKG)_DEPS     := gcc libevent openssl zlib
+
+define $(PKG)_UPDATE
+$(WGET) -q -O- 'https://torproject.org/download/download' | \
+    $(SED) -n 's,.*tor-\([0-9][^"]*\)\.tar.*,\1,p' | \
+    head -1
+endef
+
+define $(PKG)_BUILD
+    cd '$(1)' && \
+        LIBS="`'$(TARGET)-pkg-config' --libs-only-l openssl`" \
+        ./configure \
+        $(MXE_CONFIGURE_OPTS)
+    $(MAKE) -C '$(1)' -j '$(JOBS)' install $(MXE_DISABLE_DOCS)
+endef
+
+$(PKG)_BUILD_SHARED =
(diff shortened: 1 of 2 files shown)


src/tor.mk
```diff
@@ -0,0 +1,29 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+https://www.torproject.org/dist/tor-0.2.6.10.tar.gz
```
TimothyGu: huh?

TimothyGu:
Does Tor have a library?

src/tor.mk
starius: fixed

starius:
Unfortunately there is no Tor library (at least official). That is why all applications using Tor (e.g., Tor Browser) are bundeled with Tor executable.

TimothyGu:
So… this package isn't a library? I'm afraid that we might not accept that: http://mxe.cc/#creating-packages

> The package should be a free software library that is really used by one of your applications.

starius:
Cross-compiling tools using MXE is also valueble. Can we create a separate place for packages that are not libraries?

TimothyGu:
tor isn't strictly a tool don't you think?

starius:
Tor is a tool working in background :-)
Even if it is not a library literally, it is used as if it is a library by many programs and libraries. For example, [Stem](https://stem.torproject.org/) is Python library using Tor.

Personally I would preffer to use Tor built in Linux (with MXE) instead of built in Windows, because I don't trust build toolchains operating in Windows. That is why I added it to MXE: to have a reliable place for getting Tor for Windows.

starius:
By the way, there are packages in MXE which are not libraries but programs:
 * wget
 * openmp-validation
 * nsis

tonytheodore:
> there are packages in MXE which are not libraries but programs:

MXE is targeted at developers and there's a class of handy tools (gdb and sqlite in addition to the ones you mention) that make it easier to test and debug your newly built app. I can see the argument that you'd want `tor` on hand to make sure your app can be easily configured to use it reliably.

> because I don't trust build toolchains operating in Windows

Not even the official ones from a project with a mindset like Tor? The logical conclusion of that line of reason is to abandon windows altogether, including developing for it. Too philosophical and not really relevant to it's inclusion.

> That is why I added it to MXE: to have a reliable place for getting Tor for Windows.

That is probably the exact reason we don't build most binaries - we don't want to become an unofficial source.

starius:
> Not even the official ones from a project with a mindset like Tor? The logical conclusion of that line of reason is to abandon windows altogether, including developing for it. Too philosophical and not really relevant to it's inclusion.

I want to built Tor manually instead of downloading a pre-built binary even if it was built by official maintainer. Building it on Windows is not reliable, as I said. And one more reason: builing on Windows is less convenient (at least for me) than just typing `make tor`. Using Tor on Windows is still needed and it is secure if Windows itself is constrained (for example it lives in VBox).

> That is probably the exact reason we don't build most binaries - we don't want to become an unofficial source.

I wish there is a source like MXE for tools I have to use on Windows. Tor is one of them. Building them with MXE seems more secure (and convenient!) than building them on Windows or downloading binaries (often bundeled with viruses).

tonytheodore:
> I wish there is a source like MXE for tools I have to use on Windows.

Indeed! Unfortunately "user" tools are (currently) outside the scope of MXE, but something like [MSYS2](https://msys2.github.io/) may be along the lines of what you're looking for.

I can see great value in a downstream MXE project that builds tools, demonstrates use-cases, maybe builds some complex apps (say VLC)...

PR #809, opened 2015-08-29
build-pkg: add release to mxe-requirements version

starius:
2.23 -> 2.23jessie | 2.23wheezy

Package mxe-requirements has different dependencies for
Wheezy and Jessie. APT server reprepro doesn't allow to
add two different packages with same name and version.


starius:
See [warnings](https://gist.githubusercontent.com/starius/b6ec1340229aeddc1847/raw/2680c6268b9b1d4d893e0c1217c322c49b574a99/build-pkg-warnings.log) produced by build-pkg.lua against MXE 0691f569eb5ed6f5b38d1f40df39b795ff5ae404

tonytheodore:
Thanks!

Issue #434, opened 2014-07-19
License?


mmucklo:
Do you have a license for this repo?

PR #810, opened 2015-08-29
add README.md and LICENSE files

starius:


starius:
see #434

tonytheodore:
Thanks, it's probably time we caught up to the present. Only concern is to keep this short so it doesn't get out of sync with `index.html`. Might also remove the travis-ci links if they become misleading.

PR #813, opened 2015-08-30
add packages lzma and upx

starius:
**UPX** is a free, portable, extendable, high-performance executable packer for several executable formats.
http://upx.sourceforge.net

The **LZMA SDK** provides the documentation, samples, header files, libraries, and tools you need to develop applications that use LZMA compression.
http://www.7-zip.org/sdk.html

UPX depends on source of [UCL](http://www.oberhumer.com/opensource/ucl/) and LZMA. Tarballs with source of these packages are unpacked when UPX is built.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2451,8 +2451,8 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
         <td class="website"><a href="http://www.twolame.org/">TwoLAME</a></td>
     </tr>
     <tr>
-         <td class="package">ucl</td>
-         <td class="website"><a href="http://www.oberhumer.com/opensource/ucl/">UCL</a></td>
+        <td class="package">ucl</td>
+        <td class="website"><a href="http://www.oberhumer.com/opensource/ucl/">UCL</a></td>
     </tr>
     <tr>
         <td class="package">unrtf</td>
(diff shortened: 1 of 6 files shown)


tonytheodore:
Leave this on hold for the time being, I'm thinking about some sort of `mxe-exe` sister project.

tonytheodore:
LZMA and UPX now seem to fall into the developer tools side - I think these can be merged (after rebasing to remove UCL).

starius:
Some cleanup is needed. I think about using `UNPACK_PKG_ARCHIVE` for unpacking `ucl` and `lzma` in `upx` source tree.

tonytheodore:
> I think about using UNPACK_PKG_ARCHIVE for unpacking ucl and lzma in upx source tree.

Use [`PREPARE_PKG_SOURCE`](https://github.com/mxe/mxe/blob/master/src/gcc.mk#L67) to ensure patches are applied.

starius:
* checksums updated (sha256 instead of sha1)
   * PREPARE_PKG_SOURCE

tonytheodore:
Thanks!

starius:
Should we also build `upx` natively?

tonytheodore:
Yes, that would be helpful for most developers.

PR #815, opened 2015-08-30
Wt: enable shared build and x86_64-w64-mingw32 target

starius:


tonytheodore:
Thanks!

I [consolidated the static/shared build rules](https://github.com/mxe/mxe/commit/73cfc5a8aacbee2d32af2b309bcd0ee3c3690a61) to reduce duplication.

PR #821, opened 2015-09-02
Berkeley DB: enable crypto, add updater script, update version, enable shared

starius:
Build passed


src/db.mk
```diff
@@ -20,10 +21,9 @@ define $(PKG)_BUILD
         $(MXE_CONFIGURE_OPTS) \
         --enable-mingw \
         --enable-cxx \
+        --enable-cryptography \
```
tonytheodore: Does it need openssl/gnutls as a dependency?

starius: No it doesn't.

tonytheodore:
Thanks! Merged [with some rebasing](https://github.com/mxe/mxe/compare/426a622...dd1ade7)

PR #812, opened 2015-08-30
add packages libtorrent-rasterbar and qbittorrent

starius:
**libtorrent** (aka libtorrent-rasterbar) is a feature complete C++ bittorrent implementation focusing on efficiency and scalability. It runs on embedded devices as well as desktops. It boasts a well documented library interface that is easy to use. It comes with a simple bittorrent client demonstrating the use of the library.

The main goals of libtorrent are:

  *  to be cpu efficient
  *  to be memory efficient
  *  to be very easy to use

http://www.libtorrent.org/

The **qBittorrent** project aims to provide a Free Software alternative to µtorrent. Additionally, qBittorrent runs and provides the same features on all major platforms (Linux, Mac OS X, Windows, OS/2, FreeBSD).

qBittorrent is based on Qt4 toolkit and libtorrent-rasterbar.

http://www.qbittorrent.org/

PR #825, opened 2015-09-03
add package libtorrent-rasterbar

starius:
**libtorrent** (aka libtorrent-rasterbar) is a feature complete C++ bittorrent implementation focusing on efficiency and scalability. It runs on embedded devices as well as desktops. It boasts a well documented library interface that is easy to use. It comes with a simple bittorrent client demonstrating the use of the library.

The main goals of libtorrent are:

  *  to be cpu efficient
  *  to be memory efficient
  *  to be very easy to use

http://www.libtorrent.org/

See #812


tonytheodore:
Thanks!

PR #827, opened 2015-09-04
new package subversion

mgerhardy:
Could need some help here - not yet compiling. I've seen some other apr based libs/tools and some related patches - maybe they could help here, too.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2451,6 +2451,10 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
         <td class="website"><a href="http://www.cise.ufl.edu/research/sparse/SuiteSparse/">SuiteSparse</a></td>
     </tr>
     <tr>
+        <td class="package">subversion</td>
+        <td class="website"><a href="http://subversion.apache.org">subversion</a></td>
+    </tr>
+    <tr>
         <td class="package">t4k_common</td>
         <td class="website"><a href="http://tux4kids.alioth.debian.org/">t4k_common</a></td>
     </tr>
--- a/src/subversion-1-fixes.patch
+++ b/src/subversion-1-fixes.patch
@@ -1,0 +1,98 @@
+--- subversion-1.9.2/configure.ac	2015-07-27 01:03:10.000000000 +0200
++++ subversion-1.9.2.new/configure.ac	2015-12-09 13:58:49.000000000 +0100
+@@ -467,9 +467,6 @@
+ 
+ dnl Mac OS specific features -------------------
+ 
+-SVN_LIB_MACHO_ITERATE
+-SVN_LIB_MACOS_PLIST
+-SVN_LIB_MACOS_KEYCHAIN
+ 
+ dnl APR_HAS_DSO -------------------
+ 
+@@ -1271,7 +1268,6 @@
+ # Python: Used for testsuite, and bindings
+ 
+ 
+-PYTHON="`$abs_srcdir/build/find_python.sh`"
+ if test -z "$PYTHON"; then
+   AC_MSG_WARN([Python 2.7 or later is required to run the testsuite])
+   AC_MSG_WARN([or to use the Subversion Python bindings])
+
+--- subversion-1.9.2/subversion/libsvn_subr/config_win.c	2015-04-17 06:00:18.000000000 +0200
++++ subversion-1.9.2.new/subversion/libsvn_subr/config_win.c	2015-12-10 09:49:11.645120955 +0100
+@@ -34,7 +34,7 @@
+ #define WIN32_LEAN_AND_MEAN
+ /* winsock2.h includes windows.h */
+ #include <winsock2.h>
+-#include <Ws2tcpip.h>
++#include <ws2tcpip.h>
+ 
+ #include <shlobj.h>
+ 
+--- subversion-1.9.2/subversion/libsvn_subr/win32_xlate.c	2015-02-03 13:56:57.000000000 +0100
++++ subversion-1.9.2.new/subversion/libsvn_subr/win32_xlate.c	2015-12-10 09:48:56.725192873 +0100
+@@ -36,7 +36,7 @@
+ 
+ /* winsock2.h includes windows.h */
+ #include <winsock2.h>
+-#include <Ws2tcpip.h>
++#include <ws2tcpip.h>
+ #include <mlang.h>
+ 
+ #include <apr.h>
+@@ -53,6 +53,8 @@
+ 
+ #include "svn_private_config.h"
+ 
++#define INITGUID 1
++DEFINE_GUID (IID_IMultiLanguage, 0x275c23e1,0x3747,0x11d0,0x9f,0xea,0x00,0xaa,0x00,0x3f,0x86,0x46);
+ static svn_atomic_t com_initialized = 0;
+ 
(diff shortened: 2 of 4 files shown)


tonytheodore:
Are you trying to get the command line tools? See #881 for some discussion, those currently aren't in scope for mxe.

mgerhardy:
no. i am trying to get libsubversion to compile. Will retry asap

tonytheodore:
Looks like a hard one, here's some mods to configure.ac
```
--- subversion-1.9.2/configure.ac	2015-07-27 09:03:10.000000000 +1000
+++ subversion-1.9.2.new/configure.ac	2015-09-25 16:13:02.000000000 +1000
@@ -467,9 +467,6 @@
 
 dnl Mac OS specific features -------------------
 
-SVN_LIB_MACHO_ITERATE
-SVN_LIB_MACOS_PLIST
-SVN_LIB_MACOS_KEYCHAIN
 
 dnl APR_HAS_DSO -------------------
@@ -1271,7 +1268,6 @@
 # Python: Used for testsuite, and bindings
 
 
-PYTHON="`$abs_srcdir/build/find_python.sh`"
 if test -z "$PYTHON"; then
   AC_MSG_WARN([Python 2.7 or later is required to run the testsuite])
   AC_MSG_WARN([or to use the Subversion Python bindings])
```
and extra options just to make it configure.
```
--disable-mod-activation \
--without-serf \
--without-apr_memcache \
--without-apxs \
--without-jdk \
--without-jikes \
--without-swig \
--with-sysroot=$(PREFIX)/$(TARGET) \
--disable-javahl \
PKG_CONFIG=$(PREFIX)/bin/$(TARGET)-pkg-config 
```

tonytheodore:
BTW, I ran `make update-package-subversion` first to get the latest version (1.9.2).

mgerhardy:
Thanks - finally I have some more time - will try with your patches.

mgerhardy:
```
configure:19257: checking for Mach-O dynamic module iteration functions
configure:19260: error: in `/home/mgerhardy/dev/mxe.mgerhardy/tmp-subversion-i686-w64-mingw32.static/subversion-1.9.2':
configure:19262: error: cannot run test program while cross compiling
See `config.log' for more details
```
Updated the package and fixed some more issues - but still not yet done.

mgerhardy:
configures, compiles and fixed all but one linking error:

```
/libsvn_subr-1.a(win32_xlate.o): In function `get_page_id_from_name':
/home/mgerhardy/dev/mxe.mgerhardy/tmp-subversion-i686-w64-mingw32.static/subversion-1.9.2/subversion/libsvn_subr/win32_xlate.c:141: undefined reference to `IID_IMultiLanguage'
```

it looks like mingw doesn't have full support for mlang - does anybody has any experience on this side?

mgerhardy:
I also tried to define the missing IID_IMultiLanguage on my own with:
```
DEFINE_GUID (IID_IMultiLanguage, 0x275c23e1,0x3747,0x11d0,0x9f,0xea,0x00,0xaa,0x00,0x3f,0x86,0x46);
```
but this also didn't help

mgerhardy:
Ok, right now I just disabled the code that uses IID_IMultiLanguage - without knowing the impact - it builds now... let's see whether I can use it properly.

tonytheodore:
I found this very similar patch https://github.com/onepremise/Mingle/blob/master/patches/subversion/1.7.9/svn-mingw.patch#L67. Adding the `#define INITGUID 1` allows the build to succeed. 

It also builds some extra stuff that should be disabled:

```diff
diff --git a/src/subversion-1-fix-configure.patch b/src/subversion-1-fix-configure.patch
index edf0821..52a705f 100644
--- a/src/subversion-1-fix-configure.patch
+++ b/src/subversion-1-fix-configure.patch
@@ -41,10 +41,11 @@
  #include <mlang.h>

  #include <apr.h>
-@@ -53,6 +53,7 @@
+@@ -53,6 +53,8 @@

  #include "svn_private_config.h"

++#define INITGUID 1
 +DEFINE_GUID (IID_IMultiLanguage, 0x275c23e1,0x3747,0x11d0,0x9f,0xea,0x00,0xaa,0x00,0x3f,0x86,0x46);
  static svn_atomic_t com_initialized = 0;

diff --git a/src/subversion.mk b/src/subversion.mk
index 7382096..e9c3ec5 100644
--- a/src/subversion.mk
+++ b/src/subversion.mk
@@ -34,5 +34,7 @@ define $(PKG)_BUILD
         PKG_CONFIG=$(PREFIX)/bin/$(TARGET)-pkg-config \
         --with-apr='$(PREFIX)/$(TARGET)' \
         --with-apr-util='$(PREFIX)/$(TARGET)'
-    $(MAKE) -C '$(1)' -j '$(JOBS)' install LDFLAGS="-lversion -lole32 -luuid"
+    $(MAKE) -C '$(1)' -j '$(JOBS)' \
+        install-fsmod-lib install-ramod-lib install-lib install-include \
+        LDFLAGS="-lversion -lole32 -luuid -no-undefined"
 endef
```

Can't figure out how to get the shared builds to actually create dlls though.

mgerhardy:
Thanks added your patches - but I still had to disable the CoCreateInstance call - otherwise I still have unresolved symbols for IID_IMultiLanguage

mgerhardy:
(disabled shared builds for now)

mgerhardy:
sidetrack: we should introduce a mirror list for the apache projects - all the related projects are using a different mirror.

Added a test program to test linking, but include dirs are not yet found... but we are getting closer more and more ;)

tonytheodore:
It's installing the *.pc files to the wrong location (`share` instead of `lib`), adding a `pkgconfig_dir="$(PREFIX)/$(TARGET)/lib/pkgconfig"` to the `$(MAKE)` invocation installs them in the right place.

However, the *.pc files themselves don't have the correct include dir with the `subversion-1` subdir.

tonytheodore:
There's still a `--disable-shared` in the configure options. Unfortunately, removing it doesn't fix the shared build.

mgerhardy:
What do you guys think, can we maybe add this and take care of the shared build later?

tonytheodore:
Yes, we can do the shared builds later but there's still an issue with:

> However, the *.pc files themselves don't have the correct include dir with the `subversion-1` subdir.

The test program fails with:

```
./src/subversion-test.c:1:25: fatal error: svn_cmdline.h: No such file or directory
```

and I can't see a way fix it (without patching every `libsvn_*.pc.in` file).

tonytheodore:
Actually, the test has a few more issues, especially all libs are named `libsvn_*-1.a` and that isn't reflect in the `*.pc` file.

mgerhardy:
http://svn.apache.org/viewvc?view=revision&revision=r1592547

Just linking here the pkg-config support patch for libsubversion - looks like there are still some upstream errors - will try to fix them

mgerhardy:
ole32 is missing in the linkage of the test. and the Requires.private lib still doesn't contain the -1 suffix

mgerhardy:
works for me now. If nothing stops us from adding it to master - someone could please merge this in.

mgerhardy:
merged e11179655df86ddde4ccae92ed0c9f6fbcc9698e

PR #831, opened 2015-09-06
qhttpengine: decrease required cmake ver

starius:
Fix build for CMake 2.8.9, which is used in Debian Wheezy.


tonytheodore:
Isn't Jessie the new stable version? We'd normally only support the current stable, though I checked this is forward compatible with cmake 3. Thanks!

Issue #826, opened 2015-09-04
libpano13 is broken


starius:
Build script is defined for static targets only. Both of them are broken.

```
[using autodetected 6 job(s)]
make[1]: Entering directory `/home/mxe/mxe'
uname -a
Linux sd-68075 3.2.0-4-amd64 #1 SMP Debian 3.2.63-2+deb7u1 x86_64 GNU/Linux
git show-branch --list --reflog=1
  [refs/heads/master@{0}] (2 hours ago) clone: from https://github.com/mxe/mxe
lsb_release -a 2>/dev/null || sw_vers 2>/dev/null || true
Distributor ID:	Debian
Description:	Debian GNU/Linux 7.8 (wheezy)
Release:	7.8
Codename:	wheezy
rm -rf   '/home/mxe/mxe/tmp-libpano13-i686-w64-mingw32.static'
mkdir -p '/home/mxe/mxe/tmp-libpano13-i686-w64-mingw32.static'
cd '/home/mxe/mxe/tmp-libpano13-i686-w64-mingw32.static' &&  tar xzf '/home/mxe/mxe/pkg/libpano13-2.9.18.tar.gz'
cd '/home/mxe/mxe/tmp-libpano13-i686-w64-mingw32.static/libpano13-2.9.18'
sed -i 's,WINDOWSX\.H,windowsx.h,'                                                  '/home/mxe/mxe/tmp-libpano13-i686-w64-mingw32.static/libpano13-2.9.18/sys_win.h'
sed -i 's,\${WINDRES-windres},i686-w64-mingw32.static-windres,'                                  '/home/mxe/mxe/tmp-libpano13-i686-w64-mingw32.static/libpano13-2.9.18/build/win32/compile-resource'
sed -i 's,m4 -DBUILDNUMBER=\$buildnumber,sed "s/BUILDNUMBER/\$buildnumber/g",' '/home/mxe/mxe/tmp-libpano13-i686-w64-mingw32.static/libpano13-2.9.18/build/win32/compile-resource'
sed -i 's,\(@HAVE_MINGW_TRUE@am__objects_4 = .*\),\1 ppm.lo,'                       '/home/mxe/mxe/tmp-libpano13-i686-w64-mingw32.static/libpano13-2.9.18/Makefile.in'
[...]

PR #834, opened 2015-09-06
libjpeg-turbo: install DLL to bin/libjpeg-turbo

starius:
Enable shared build. It was disabled because DLLs were installed to lib/bin.

see #826

### DIFF
--- a/src/libjpeg-turbo-2-install-DLL-to-bin-libjpeg-turbo.patch
+++ b/src/libjpeg-turbo-2-install-DLL-to-bin-libjpeg-turbo.patch
@@ -19,7 +19,7 @@ index a968835..6acc338 100644
  	  tdlname=$dlname
  	  case $host,$output,$installed,$module,$dlname in
 -	    *cygwin*,*lai,yes,no,*.dll | *mingw*,*lai,yes,no,*.dll) tdlname=../bin/$dlname ;;
-+	    *cygwin*,*lai,yes,no,*.dll | *mingw*,*lai,yes,no,*.dll) tdlname=../../bin/libjpeg-turbo/$dlname ;;
++	    *cygwin*,*lai,yes,no,*.dll | *mingw*,*lai,yes,no,*.dll) tdlname=../../bin/$dlname ;;
  	  esac
  	  $echo > $output "\
  # $outputname - a libtool library file
--- a/src/libjpeg-turbo-2-install-DLL-to-bin-libjpeg-turbo.patch
+++ b/src/libjpeg-turbo-2-install-DLL-to-bin-libjpeg-turbo.patch
@@ -1,0 +1,28 @@
+This file is part of MXE.
+See index.html for further information.
+
+From 703326874bf235adb392b9748ff0a56a08991690 Mon Sep 17 00:00:00 2001
+From: Boris Nagaev <bnagaev@gmail.com>
+Date: Sun, 6 Sep 2015 17:45:40 +0100
+Subject: [PATCH] install DLL to bin/libjpeg-turbo not to lib/bin
+
+---
+ ltmain.sh | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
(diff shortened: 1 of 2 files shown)


src/libjpeg-turbo-2-install-DLL-to-bin-libjpeg-turbo.patch
```diff
@@ -0,0 +1,28 @@
...
+diff --git a/ltmain.sh b/ltmain.sh
+index a968835..6acc338 100644
+--- a/ltmain.sh
++++ b/ltmain.sh
+@@ -5229,7 +5229,7 @@ fi\
+ 	  # place dlname in correct position for cygwin
+ 	  tdlname=$dlname
+ 	  case $host,$output,$installed,$module,$dlname in
+-	    *cygwin*,*lai,yes,no,*.dll | *mingw*,*lai,yes,no,*.dll) tdlname=../bin/$dlname ;;
++	    *cygwin*,*lai,yes,no,*.dll | *mingw*,*lai,yes,no,*.dll) tdlname=../../bin/libjpeg-turbo/$dlname ;;
```
TimothyGu: Shouldn't it be ../../bin/$dlname? We don't want DLLs in a subdirectory of bin/.

starius: DLL libjpeg-62.dll in bin/ can overwrite a file from package `jpeg`.

Can you explain why DLLs are written to `bin`, please?

TimothyGu: > DLL libjpeg-62.dll in bin/ can overwrite a file from package jpeg.

At least http://mxe.redjohn.tk/repos/apt/debian/pool/main/m/mxe-i686-w64-mingw32.shared-jpeg/mxe-i686-w64-mingw32.shared-jpeg_9a_amd64.deb doesn't have the file you are describing.

> Can you explain why DLLs are written to bin, please?

To be honest, I do not know. My guesses are:

1. dynamically linked exes can be directly executed if the DLLs are in the same directory as they are
2. historically you can't "link" to a DLL – only to its import library – which is the reason why import libs is installed to `lib/`

starius: > At least http://mxe.redjohn.tk/repos/apt/debian/pool/main/m/mxe-i686-w64-mingw32.shared-jpeg/mxe-i686-w64-mingw32.shared-jpeg_9a_amd64.deb doesn't have the file you are describing.

I mean this could happen. Package `jpeg` has file `i686-w64-mingw32.shared/bin/libjpeg-9.dll`. If I understand correctly, this `9` is something like SONAME in Linux. `libjpeg-turbo` has file `libjpeg-62.dll`. If `9` becomes `62`, the file would be overwritten.

> dynamically linked exes can be directly executed if the DLLs are in the same directory as they are

Good reason.

I don't know where the right place for `libjpeg-62.dll` is.

BTW, jack installs DLLs to `bin/jack`. It may be related to e6b645eaaa3f28527993eb544ec4666c06fe40a9 though it changes `/lib/` to `/lib/jack`. Maybe Jack's build system is too clever to decide to install DLLs to `bin/jack`... I don't know.

tonytheodore: The "right" place for dlls is `bin/`, putting them anywhere else imposes extra work for people wanting to use them. It's not so bad for static libs as they only need to be found at build-time and pkg-config can help.

The real issue isn't so much where they are installed, but which package is installed. I've added a note for the [goals of customisable packages](https://github.com/mxe/mxe/wiki/Customisable-Packages)

starius:
I have updated this patch. Now DLLs are installed to bin/

tonytheodore:
Thanks, I'm still not sure about the best way to handle these but we can wait till there's someone really trying to use it.

PR #836, opened 2015-09-07
Refactor MXE configuration and cmake toolchain

tonytheodore:
See #823

PR #843, opened 2015-09-11
build-pkg: add env. var. MXE_DIR and MXE_NO_DEBS

starius:
See https://github.com/mxe/mxe/pull/836#issuecomment-139034608

### DIFF
--- a/tools/build-pkg.lua
+++ b/tools/build-pkg.lua
@@ -11,7 +11,15 @@
 -- Packages are written to `*.tar.xz` files.
 -- Debian packages are written to `*.deb` files.
 
+-- To prevent build-pkg from creating deb packages,
+-- set environment variable MXE_NO_DEBS to 1
+-- In this case fakeroot and dpkg-deb are not needed.
+
+-- To limit number of packages being built to x,
+-- set environment variable MXE_MAX_PACKAGES to x,
+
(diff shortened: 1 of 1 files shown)


TimothyGu:
Isn't the whole point of build-pkg producing distributable formats of packages (i.e. tarballs or debs)?

tonytheodore:
@TimothyGu Yes, that is it's main purpose. It also [detects various inconsistencies](https://gist.githubusercontent.com/starius/2a60589ae3fccc7a9306/raw/dc7efb5d78f42a6a7db4a5d4e98a0d8f6272d182/build-mxe-conf-cmake-09Sep2015.log) and is useful to run for that alone (say on OSX).

@starius would these be better as command line options or is that overkill at this stage?

starius:
> would these be better as command line options or is that overkill at this stage?

Lua provides command line options only as an array of arguments (global variable `arg`), similar to C's `argc` and `argv`. To parse them properly, external Lua module is required (I prefer [argparse](https://github.com/mpeterv/argparse)). I think, requiring external Lua modules for `build-pkg` is overkill.

starius:
> Isn't the whole point of build-pkg producing distributable formats of packages (i.e. tarballs or debs)?

Tarballs are produced even with option `MXE_NO_DEBS`. Useful on OSX, which doesn't have `dpkg-deb`.

tonytheodore:
Thanks!

PR #846, opened 2015-09-13
build-pkg: no build mxe-requirements with fakeroot

### LINKED ISSUE #840: build-pkg: fix "fakeroot: database file `deb.fakeroot' does not exist."

starius:
`mxe-requirements` package doesn't install any files therefore fakeroot is not required. Because no files were created with fakeroot, database file `deb.fakeroot` did not exist.

close #840


tonytheodore:
Thanks!

Issue #386, opened 2014-04-21
Wrong checksum of package agg


KaiMartin:
This is my first encounter with mxe. So I followed the tutorial at http://mxe.cc/#tutorial
apt-get and git clone were fine. 
But on make I get: 

```
$ make
[target]   i686-pc-mingw32 with 276 goals from src/*.mk
[download] agg

Wrong checksum of package agg!
- - --------------------------------------------------------- -
Resolving www.antigrain.com (www.antigrain.com)... 208.91.197.132
Connecting to www.antigrain.com (www.antigrain.com)|208.91.197.132|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 271 [text/html]
Saving to: ‘STDOUT’
     0K                                                       100% 21.2M=0s
2014-04-21 03:59:22 (21.2 MB/s) - written to stdout [271/271]
- - --------------------------------------------------------- -
[log]      /usr/local/src/mxe/mxe/log/agg-download

make: *** [/usr/local/src/mxe/mxe/usr/i686-pc-mingw32/installed/agg] Error 1
```

The host system is Debian/testing  

---<)kaimartin(>---

PR #848, opened 2015-09-13
Revert "package agg: use mxe pkg mirror until project relocates"

starius:
This reverts commit c9afbabda56635bb1f2f49090a0c2091d4ab1340.

AGG project's site was restored.

See #386


tonytheodore:
Thanks!

PR #795, opened 2015-08-24
Add gtk3

mgehre:
This updates glib, atk and pango to the minimal versions required by gtk3.
And adds the gtk3 package.

In librsvg, there was an error due to typedef redefinition caused by the glib update. I removed '-pedantic' from the librsvg test flags to work around that issue.

PR #850, opened 2015-09-13
add packages gtk3 and gtkmm3

starius:
Pango has been updated to 1.37.4 in 24cf97dbdd62ece0ea0b39cf33694a8804352bd1

Side effects:
  * mute function g_win32_run_session_bus of glib (see the thread)
  * add new external dependency: gdk-pixbuf

Original thread:
https://lists.nongnu.org/archive/html/mingw-cross-env-list/2015-08/msg00009.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -612,6 +612,10 @@ USE_OSGPLUGIN(&lt;plugin2&gt;)
         <td><a href="http://tukaani.org/xz/">XZ Utils</a></td>
         <td></td>
     </tr>
+    <tr>
+        <td><a href="http://www.gdk-pixbuf.org/">gdk-pixbuf</a></td>
+        <td></td>
+    </tr>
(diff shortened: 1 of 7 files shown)


tonytheodore:
These build successfully on OSX (though the test programs are huge).

Is there a reason this is an older version? #795 has 3.16.6 and there seem to have been [many releases](http://git.gnome.org/browse/gtk+/refs/tags) in the 3.14 series since 3.14.4.

tonytheodore:
> Is there a reason this is an older version?

Seems it was waiting on a [glib update](https://github.com/mxe/mxe/commit/06a18c9af8fc8eac73429b69be8d0793ee81e051).

tonytheodore:
Okay, let's use this as the base for future updates.

Thanks!

PR #851, opened 2015-09-13
Qt: test more Qt tools (moc, rcc)

starius:
Qt header is translated to C++ code by `moc` tool.
Qt resource file (.qrc) is translated to C++ code by `rcc` tool.


tonytheodore:
Thanks!

PR #853, opened 2015-09-14
nsis: increase NSIS_MAX_STRLEN to 8192

starius:
> Currently you can't change NSIS_MAX_STRLEN without re-compiling NSIS.
> This was a problem because some NSIS projects would require the special
> build and some wouldn't, and sometimes we'd compile a project that
> required the special build, with the normal MakeNSIS build.

(from http://nsis.sourceforge.net/SetReqStrLen:_Allow_compile_w/_8192_special_build_only)

See also http://nsis.sourceforge.net/Special_Builds#Large_strings


tonytheodore:
Thanks!

Issue #852, opened 2015-09-13
 MXE + Qt + CMake = linking error (lcms2) 


starius:
Qt [can be used](http://www.cmake.org/Wiki/CMake/Tutorials/Qt) with cmake as well as with qmake.

Using cmake with Qt, I get linking error because of missing `-llcms2`. I created minimal example reproducing the error as a patch for MXE. Apply  pull request #851 and [the patch](https://gist.github.com/starius/3dc64ab7c23c4457fd98) and type `make qt`. It builds Qt example with cmake.

I get the following errors:

```
Linking CXX executable qt-test-cmake.exe
/usr/bin/cmake -E cmake_link_script CMakeFiles/qt-test-cmake.dir/link.txt --verbose=1
/usr/bin/cmake -E remove -f CMakeFiles/qt-test-cmake.dir/objects.a
/home/mxe/mxe-conf-cmake/usr/bin/x86_64-w64-mingw32.static-ar cr CMakeFiles/qt-test-cmake.dir/objects.a @CMakeFiles/qt-test-cmake.dir/objects1.rsp
/home/mxe/mxe-conf-cmake/usr/bin/x86_64-w64-mingw32.static-g++  -O3 -DNDEBUG    -Wl,--whole-archive CMakeFiles/qt-test-cmake.dir/objects.a -Wl,--no-whole-archive  -o qt-t
est-cmake.exe -Wl,--out-implib,libqt-test-cmake.dll.a -Wl,--major-image-version,0,--minor-image-version,0  /home/mxe/mxe-conf-cmake/usr/x86_64-w64-mingw32.static/qt/lib/l
ibQtGui.a -Wl,-Bstatic -lpng -ljpeg -Wl,-Bdynamic -ltiff -llzma -ljpeg -lz -Wl,-Bstatic -lmng -Wl,-Bdynamic -limm32 -lwinmm /home/mxe/mxe-conf-cmake/usr/x86_64-w64-mingw3
2.static/qt/lib/libQtCore.a -Wl,-Bstatic -lz -Wl,-Bdynamic -lws2_32 -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 
[...]

PR #857, opened 2015-09-15
update vtk6

muellni:


tonytheodore:
Thanks Martin, this is a very welcome update!

Do you have any insight on the [static linking errors in qt](https://github.com/muellni/mxe/commit/dc2526a2d89b83a3398a503e2b1e78ed2f0c6dac). There's an open issue at #852 and it would be great to fix the root cause.

tonytheodore:
P.S. The issue is [nearly five years old](http://lists.gnu.org/archive/html/mingw-cross-env-list/2010-11/msg00069.html) and all help is appreciated.

tonytheodore:
There's no reason to hold this up for #852, could you move the [qt fix](https://github.com/muellni/mxe/commit/dc2526a2d89b83a3398a503e2b1e78ed2f0c6dac) to libmng (it already has a pkg-config hack)?

tonytheodore:
> could you move the qt fix to libmng

On second thoughts, it's okay where it is. Thanks!

PR #864, opened 2015-09-18
report downloading from backup

starius:
Found following problems:
  * MXE Warning! Downloading muparser from backup.
  * MXE Warning! Downloading openscenegraph from backup.
  * MXE Warning! Downloading suitesparse from second URL.


tonytheodore:
Nice!

PR #866, opened 2015-09-18
muparser: update upstream URL and version to 2.2.5

starius:
[Original muparser location](http://sourceforge.net/projects/muparser/) forwards to [Google Code](https://code.google.com/p/muparser/), which in turn forwards to [Github](https://github.com/beltoforion/muparser).

Version 2.2.3 can't be found on Github, that is why it was updated to 2.2.5.

Updater script updated as well.

Build passes on Debian Jessie.

### DIFF
--- a/src/muparser.mk
+++ b/src/muparser.mk
@@ -3,17 +3,16 @@
 
 PKG             := muparser
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 2.2.3
-$(PKG)_CHECKSUM := dfe831b69392ab0b1eb59d3601b7b1575554a85057cf2f234f64f930c4148902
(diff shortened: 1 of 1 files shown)


tonytheodore:
See #649 and https://github.com/mxe/mxe/commits/master/src/qtserialport_qt4.mk. It would be good to standardise these.

starius:
Rebased. `MXE_GET_GITHUB_TAGS` from https://github.com/mxe/mxe/pull/649#issuecomment-77987449 doesn't work as tag name for muparser is prefixed with 'v': 'v2.2.5'.

starius:
Rewritten using MXE_GET_GITHUB_TAGS
Depends on #888

PR #868, opened 2015-09-18
fix suitesparse download URL

starius:
Old URL is 404. Parent URL redirects to http://faculty.cse.tamu.edu/davis/research.html


tonytheodore:
Thanks!

PR #874, opened 2015-09-20
build-pkg: pre-download all packages and check *.a with nm

starius:


tonytheodore:
Why call it testCommand? That implies it's only a check and no intended to do anything.

starius:
Shall it rename it to `execute`?

tonytheodore:
`execute` sounds good.

starius:
Done

tonytheodore:
Thanks!

PR #880, opened 2015-09-22
Add pthreads.cmake file generation

ChristianFrisson:
A workaround for the TRY_RUN error in FindThreads.cmake when cross-compiling with mxe and CMake.

Similarly to:
https://github.com/osrf/rba_scripts/issues/5

### DIFF
--- a/src/pthreads.mk
+++ b/src/pthreads.mk
@@ -24,4 +24,8 @@ PTHREADS_TEST = \
     '$(TARGET)-gcc' \
         -W -Wall -Werror -ansi -pedantic \
         '$(TOP_DIR)/src/pthreads-test.c' -o '$(PREFIX)/$(TARGET)/bin/test-pthreads.exe' \
-        `'$(TARGET)-pkg-config' --libs pthreads`
+        `'$(TARGET)-pkg-config' --libs pthreads` && \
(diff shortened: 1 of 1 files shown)


tonytheodore:
Hmmm, I'm not sure. I'd say there's something wrong with `FindThreads.cmake` as a test for pthreads shouldn't involve a `TRY_RUN`.

tonytheodore:
I've added a [minimal cmake test] (https://github.com/mxe/mxe/commit/fd257c1335663f9452a9f360a3fcc72108053da9) to see if `find_package(Threads REQUIRED)` works. I suspect something else is the issue.

ChristianFrisson:
Hi,

I've tried your test and it works for me too. I just had to add `cmake_minimum_required(VERSION 3.0)`as first generated line to the cmake test CMakeLists.txt.

So I double checked the project I was trying to cross-compile, it was forking its own FindThreads.cmake... Removing it solved the issue!

This pull request can be solved for me.

Sorry for wasting your time.

Best regads,
Christian

tonytheodore:
I'll leave it open for the moment, the test needs tidying up and later versions of cmake can use `Threads::Threads` instead of `${CMAKE_THREAD_LIBS_INIT}`.

PR #882, opened 2015-09-23
New package: enet library

ooxi:
ENet's purpose is to provide a relatively thin, simple and robust network communication layer on top of UDP (User Datagram Protocol).The primary feature it provides is optional reliable, in-order delivery of packets. ENet omits certain higher level networking features such as authentication, lobbying, server discovery, encryption, or other similar tasks that are particularly application specific so that the library remains flexible, portable, and easily embeddable.

Patch contains an `enet-test.c` program and was compiled and tested on Debian Jessie with the following targets

 * x86_64-w64-mingw32.static
 * i686-w64-mingw32.static
 * x86_64-w64-mingw32.shared
 * i686-w64-mingw32.shared

### DIFF
--- a/src/enet.mk
+++ b/src/enet.mk
@@ -1,0 +1,53 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := enet
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 5f476546edabdf37509cd3448d1a616f5eca535d
+$(PKG)_CHECKSUM := 1c49f074d611b3e00c79abd94cb1df8f1b4bb013ebd20e0d03015ddac955ff16
+$(PKG)_SUBDIR   := $(PKG)-$($(PKG)_VERSION)
+$(PKG)_FILE     := $($(PKG)_SUBDIR).tar.gz
+$(PKG)_GITHUB   := https://github.com//lsalzman/enet
+$(PKG)_URL      := $($(PKG)_GITHUB)/archive/$($(PKG)_VERSION).tar.gz
+$(PKG)_DEPS     := gcc
+
+define $(PKG)_UPDATE
+    echo 'TODO: write update script for $(PKG).' >&2;
+    echo $($(PKG)_VERSION)
+endef
+
+define $(PKG)_BUILD
+
+    # Configure
+    mkdir '$(1).build'
+    cd '$(1).build' && cmake \
+      -DCMAKE_TOOLCHAIN_FILE='$(CMAKE_TOOLCHAIN_FILE)' \
+      '$(1)'
+
+
+    # Build
+    $(MAKE) -C '$(1).build' -j '$(JOBS)'
+
+
+    # Install include files
+    $(INSTALL) -D '$(1)/include/enet/callbacks.h' '$(PREFIX)/$(TARGET)/include/enet/callbacks.h'
+    $(INSTALL) -D '$(1)/include/enet/enet.h'      '$(PREFIX)/$(TARGET)/include/enet/enet.h'
+    $(INSTALL) -D '$(1)/include/enet/list.h'      '$(PREFIX)/$(TARGET)/include/enet/list.h'
+    $(INSTALL) -D '$(1)/include/enet/protocol.h'  '$(PREFIX)/$(TARGET)/include/enet/protocol.h'
+    $(INSTALL) -D '$(1)/include/enet/time.h'      '$(PREFIX)/$(TARGET)/include/enet/time.h'
+    $(INSTALL) -D '$(1)/include/enet/types.h'     '$(PREFIX)/$(TARGET)/include/enet/types.h'
+    $(INSTALL) -D '$(1)/include/enet/unix.h'      '$(PREFIX)/$(TARGET)/include/enet/unix.h'
+    $(INSTALL) -D '$(1)/include/enet/utility.h'   '$(PREFIX)/$(TARGET)/include/enet/utility.h'
+    $(INSTALL) -D '$(1)/include/enet/win32.h'     '$(PREFIX)/$(TARGET)/include/enet/win32.h'
+
+    # Install library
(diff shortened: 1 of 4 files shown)


src/enet.mk
```diff
@@ -0,0 +1,53 @@
...
+
+    # Install library
+    $(INSTALL) -D '$(1).build/libenet.a'            '$(PREFIX)/$(TARGET)/lib/libenet.a'
+
+
+    # Build test executable
+    '$(TARGET)-gcc' \
+        -W -Wall -Werror -ansi -pedantic \
+        '$(2).c' -o '$(PREFIX)/$(TARGET)/bin/test-enet.exe' \
+        -lenet -lws2_32 -lwinmm
```
starius: use pkg-config to get a list of libraries to link against.

src/enet.mk
```diff
@@ -0,0 +1,53 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := enet
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 5f476546edabdf37509cd3448d1a616f5eca535d
+$(PKG)_CHECKSUM := 1c49f074d611b3e00c79abd94cb1df8f1b4bb013ebd20e0d03015ddac955ff16
+$(PKG)_SUBDIR   := $(PKG)-$($(PKG)_VERSION)
+$(PKG)_FILE     := $($(PKG)_SUBDIR).tar.gz
+$(PKG)_GITHUB   := https://github.com//lsalzman/enet
```
starius: replace // with /

src/enet.mk
```diff
@@ -0,0 +1,53 @@
...
+$(PKG)_URL      := $($(PKG)_GITHUB)/archive/$($(PKG)_VERSION).tar.gz
+$(PKG)_DEPS     := gcc
+
+define $(PKG)_UPDATE
```
starius: use MXE_GET_GITHUB_TAGS, see #649

src/enet.mk
```diff
@@ -0,0 +1,53 @@
...
+    $(INSTALL) -D '$(1)/include/enet/list.h'      '$(PREFIX)/$(TARGET)/include/enet/list.h'
+    $(INSTALL) -D '$(1)/include/enet/protocol.h'  '$(PREFIX)/$(TARGET)/include/enet/protocol.h'
+    $(INSTALL) -D '$(1)/include/enet/time.h'      '$(PREFIX)/$(TARGET)/include/enet/time.h'
+    $(INSTALL) -D '$(1)/include/enet/types.h'     '$(PREFIX)/$(TARGET)/include/enet/types.h'
+    $(INSTALL) -D '$(1)/include/enet/unix.h'      '$(PREFIX)/$(TARGET)/include/enet/unix.h'
+    $(INSTALL) -D '$(1)/include/enet/utility.h'   '$(PREFIX)/$(TARGET)/include/enet/utility.h'
+    $(INSTALL) -D '$(1)/include/enet/win32.h'     '$(PREFIX)/$(TARGET)/include/enet/win32.h'
+
+    # Install library
+    $(INSTALL) -D '$(1).build/libenet.a'            '$(PREFIX)/$(TARGET)/lib/libenet.a'
```
starius: You install .a file only. What about shared build and DLLs? I have checked, that it builds `.a` instead of `.dll` for shared targets as well.

versions.json
```diff
@@ -44,6 +44,7 @@
     "devil": "1.7.8",
     "dlfcn-win32": "1.0.0",
     "eigen": "3.2.5",
+    "enet": "5f476546edabdf37509cd3448d1a616f5eca535d",
```
starius: Don't do it manually. The bot will do it for you.

src/enet.mk
ooxi: I was not familiar with `pkg-config`, I'll try to learn and update the patch

src/enet.mk
ooxi: Of course, I did not see that, thanks for the catch!

src/enet.mk
ooxi: Unfortunately that's not possible since libenet does not tag releases. [It's requested, though](https://github.com/lsalzman/enet/issues/46) but for the time being I think it's necessary to use the commit id instead. I could of course fork enet and provide the tags myself, what's your opinion on that?

src/enet.mk
ooxi: I will update the build script accordingly!

versions.json
ooxi: I thought the bot was just as last net if you forget do do it manually. Won't do it :)

ooxi:
Thanks @starius for the feedback! I will try to address all issues :)

src/enet.mk
starius: MXE prefers original sources, so don't fork it just to provide tags. If enet doesn't have releases, then don't provide update script.

ooxi:
@starius I got shared compilation working but I don't know how to proceed with `pkg-config`. The upstream sources do not provide such a `pkg-config` file and I don't know how to generate one. Could you give me a hint or guide me to a tutorial?

starius:
`libenet` provides a file [libenet.pc.in](https://github.com/lsalzman/enet/blob/master/libenet.pc.in) from which `libenet.pc` is produced.

enet's README [says](https://github.com/lsalzman/enet/blob/master/README) the following:
> If you obtained this package from github, the quick description on how to build
is:
> Generate the build system.
> autoreconf -vfi
> Compile and install the library.
> ./configure && make && make install

`configure` generates `libenet.pc` from `libenet.pc.in`

Add cross-compiling options `$(MXE_CONFIGURE_OPTS)` to `configure`.

Use `make install` instead of manually copying each file. I hope `make install` installs `libenet.pc` to `usr/<target>/lib/pkgconfig/`.

By the way you can use official download with version umber from official site:
http://enet.bespin.org/Downloads.html

ooxi:
Thanks, I will issue a new pull request based on your input!

PR #885, opened 2015-09-23
llvm: enable shared build

starius:


tonytheodore:
I'm not sure this does what it would appear to do. From memory, llvm is designed to be always static.

starius:
Changes:
  * set BUILD_SHARED_LIBS to `SHARED` for shared targets
  * copy DLLs to bin (closes #886 )

From [llvm docs](http://llvm.org/docs/CMake.html#frequently-used-cmake-variables):
> BUILD_SHARED_LIBS:BOOL
    Flag indicating if shared libraries will be built. Its default value is OFF. This option is only recommended for use by LLVM developers. On Windows, shared libraries may be used when building with MinGW, including mingw-w64, but not when building with the Microsoft toolchain.

tonytheodore:
Thanks!

Issue #881, opened 2015-09-23
Building handy tools and applications


tonytheodore:
Following on from #808, #812, and #813 it is indeed very handy to have a source for tools that are outside the scope of mxe. As an example, I've always turned a blind eye to `ffmpeg.exe` that we build since it's great to have a static and recent version on hand.

Some sort of sibling or downstream project (mxe-exe?) could be more liberal in what packages it accepts, and would allow the core mxe to be stricter about focussing on libs and headers. We could also possibly use it as place to gather build recipes and examples.

Any thoughts?

starius:
Maybe name `mxe-apps` is better than `mxe-exe`?

Sister project `mxe-exe` is fine, though for me it is more convenient to have all packages in one place. Having some packages in other repo some kind of synchronization between them is required. Maybe adding executable packages to MXE in disabled state is better. Without `BUILD_APPS` option enabled, these packages do not present. Or something like this.

Technical issue: how to use MXE from mxe-exe? I like the idea of Makefile and xxx.mk per package. Can it be used in mxe-exe without copy-paste from MXE? Ideally, it should work transparently as if mxe-exe's packages are MXE packages. Dependencies should work. For example, building qbittorrent in mxe-exe should cause building qt in MXE.

Another proposal. Add "plugins" for MXE. It would work as follows: a user sets MXE variable `MXE_PLUGINS` to "/path/to/mxe-exe;/path/to/other/plugin" and has all packages from those paths as a part of MXE.

PS. #813 has two libs (lzma and ucl) and one executable (upx), which can be considered as a part of a compiler chain. Many people apply it after stripping a binary to make it even smaller.

PS 2. I have one more candidate package keepassx: 6c561c5f3307944d7b6d7ec3de732b25bf69ed00

tonytheodore:
> Maybe name mxe-apps is better than mxe-exe?

Agreed, apps is a pretty universal term these days and it conveys the idea well.

> it is more convenient to have all packages in one place. 

Indeed, but I'd like to convey the idea that these aren't particularly supported. Maybe having them in a separate place will actually cause more work (issue tracker, repo, etc.) but it's hard to say. 

> Technical issue: how to use MXE from mxe-exe?

It's the old library vs framework debate - does one include MXE from their code and execute it, or does MXE include your code and execute it? I was mostly thinking of the former with some sort of git submodule approach to keep things synced, however ...

> Another proposal. Add "plugins" for MXE.

[...]

PR #889, opened 2015-09-26
new package: unrtf

mabrand:
This is a tool for converting RTF (rich text format) to several usable formats. As far as I know, there is no portable library that does this, so unrtf is useful for applications that need this functionality.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2459,6 +2459,10 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
         <td class="website"><a href="http://www.twolame.org/">TwoLAME</a></td>
     </tr>
     <tr>
+        <td class="package">unrtf</td>
+        <td class="website"><a href="https://www.gnu.org/software/unrtf/">unRTF</a></td>
+    </tr>
+    <tr>
         <td class="package">vamp-plugin-sdk</td>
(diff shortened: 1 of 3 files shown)


starius:
These two unrelated changes (empty line addition and deletion) undo each other. IMHO they should be removed from the patch.

starius:
This package installs `usr/<target>/bin/unrtf.exe` and a directory `usr/<target>/share/unrtf/`. The directory has the following files in it:
  * html.conf  
  * latex.conf  
  * rtf.conf  
  * SYMBOL.charmap  
  * text.conf  
  * troff_mm.conf  
  * vt.conf

Does `unrtf` use them if it is compiled in MXE?

starius:
IMHO, in light of #881 this package is a candidate for `mxe-exe`, because it provides a tool, not a library, and this tool is not a part of build-chain.

mabrand:
Thanks for the reviw. I cleaned up the patch.
unrtf needs the config files to run. It has switch -P to specify configuration directory.

tonytheodore:
@starius This would still be an edge case even with #881 as it's not really a handy or end user tool, more something a developer would bundle with their app to get the functionality that can't be provided in another way. Probably better to think in terms of use case of developers vs users rather than the implementation artefacts, but I'll try and clarify it further.

PR #893, opened 2015-09-27
protobuf: install prefixed protoc tool

starius:
This tool is needed by some packages using protobuf (e.g., mosh).


tonytheodore:
Thanks!

PR #894, opened 2015-09-28
add package speexdsp

Hopper262:
The Speex developers split the project into two separate repositories in 1.2rc2. This adds the spun-off SpeexDSP library to MXE, so projects like mine (Aleph One) can continue to use the functionality previously covered by the "speex" package.


TimothyGu:
Thanks!

PR #895, opened 2015-09-29
widl: generation of widl.cmake 

ChristianFrisson:
with CMAKE_WIDL indicating the path of the widl executable


tonytheodore:
Thanks!

PR #896, opened 2015-09-29
pkgconf: added option CACHE PATH for PKG_CONFIG_EXECUTABLE 

ChristianFrisson:
in pkgconf.cmake so that pkgconf can be found through cmake at first run.


tonytheodore:
Do [other entries in the toolchain file](https://github.com/mxe/mxe/blob/master/src/mxe-conf.mk) also need the `CACHE PATH` option?

ChristianFrisson:
I have not had issues with other entries from the cmake toolchain so far.

Issue #10, opened 2012-04-12
Use different msvcrt


tonytheodore:
See:
http://lists.nongnu.org/archive/html/mingw-cross-env-list/2012-01/msg00008.html

PR #897, opened 2015-09-29
mingw-w64: enable secure API

ChristianFrisson:
In that case MINGW_HAS_SECURE_API is defined and mingw can compile against additional versions of strcpy, strcat... suffixed with '_s'.

### DIFF
--- a/src/mingw-w64.mk
+++ b/src/mingw-w64.mk
@@ -23,7 +23,8 @@ define $(PKG)_BUILD_mingw-w64
         --host='$(TARGET)' \
         --prefix='$(PREFIX)/$(TARGET)' \
         --enable-sdk=all \
-        --enable-idl
+        --enable-idl \
+        --enable-secure-api
     $(MAKE) -C '$(1).headers-build' install
 endef
 


tonytheodore:
Does this break WinXP compatibility?

ChristianFrisson:
From https://chromium.googlesource.com/v8/v8/+/3.31.0/src/base/platform/platform-win32.cc 
> Secure API functions are not available using MinGW with msvcrt.dll on Windows XP.

Then this code could be used on WinXP:
```
#ifdef MINGW_HAS_SECURE_API
#undef MINGW_HAS_SECURE_API
#endif  // MINGW_HAS_SECURE_API
```

tonytheodore:
Unfortunately, this will need some more thought, investigation, and consultation. I was ready to [drop XP support](http://lists.nongnu.org/archive/html/mingw-cross-env-list/2014-12/msg00014.html) a while ago but we don't want to create additional maintenance by deliberately disabling it in the runtime then fixing everything downstream.

One of our oldest issues is selecting different `msvcr*` runtimes (#10) and that's possibly where we should start. I suspect some of the secure API is header-only, but I'd like to be certain.


P.S. Lately, I was shocked (shocked!) to realise that I'm still running XP VMs after all this time.

ChristianFrisson:
Searching for MINGW_HAS_SECURE_API and one of the secure version of the functions (for instance strcat_s) in the mingw-w64 source code (a github fork) https://github.com/Alexpux/mingw-w64/search?q=MINGW_HAS_SECURE_API https://github.com/Alexpux/mingw-w64/search?q=strcat_s&type=Code seems to indicate it might be header-only and just bindings to MS dlls.

tonytheodore:
mingw-w64 is now built inline with gcc (#925). I'll roll this in with the next iteration on #1019 which introduces various extra customisation capabilities.

ChristianFrisson:
Great news thanks!

Issue #890, opened 2015-09-27
build-pkg builds mxe-conf and flex as one package


starius:
```
[using autodetected 6 job(s)]
[check requirements]
[build]     mxe-conf               i686-w64-mingw32.static                                                 
[done]      mxe-conf               i686-w64-mingw32.static                                  312 KiB        0m5.343s
[no-build]  flex                   i686-w64-mingw32.static                                                 
[using autodetected 6 job(s)]
[build]     gcc-gmp                i686-w64-mingw32.static                                                 
[done]      gcc-gmp                i686-w64-mingw32.static                                  28352 KiB      0m25.476s
```

```
$ cat i686-w64-mingw32.static-flex.list
usr/bin/config.guess
usr/bin/i686-w64-mingw32.static-cmake
usr/i686-w64-mingw32.static/installed/flex
usr/i686-w64-mingw32.static/installed/mxe-conf
usr/i686-w64-mingw32.static/lib/pkgconfig/gl.pc
usr/i686-w64-mingw32.static/lib/pkgconfig/glu.pc
usr/i686-w64-mingw32.static/share/cmake/mxe-conf.cmake
usr/i686-w64-mingw32.static/share/config.site
```

Probably `build-pkg` should analyze output of `make <package>` and produce an error message if there are more than one build.

PR #898, opened 2015-09-29
build-pkg: add MXE_CONF_PKGS to deps of packages

starius:
`mxe-conf` is a special package added to dependencies of other packages implicitly, not via `$(PKG)_DEPS`. Tool `build-pkg` used to find dependencies in `$(PKG)_DEPS` only. Now it looks in $(MXE_CONF_PKGS) as well. This code was adopted from `PKG_TARGET_RULE`.

see #890

### DIFF
--- a/tools/build-pkg.lua
+++ b/tools/build-pkg.lua
@@ -180,7 +180,9 @@ print-deps:
 	@$(foreach pkg,$(PKGS),echo \
 		for-build-pkg $(pkg) \
 		$(subst $(SPACE),-,$($(pkg)_VERSION)) \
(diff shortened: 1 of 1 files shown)


tonytheodore:
This will work in the short term, but "target" dependencies will likely break in the same way down the track (though now I'll keep it in mind and provide some better logic in the Makefile).

The way to get an ordered list of what `make` is going to build is a bit ugly:
```
make --debug=b -n --always-make -j1 | grep 'Must remake target' | grep -v download-only | sed -n "s,.*/usr/\(.*\)/installed/\(.*\)'\.,\1 \2,p"
```

PR #907, opened 2015-10-06
Add patch to openexr to fix pointer conversion.

saiarcot895:
This allows building on x86_64-w64-mingw32.shared (at least).


tonytheodore:
Thanks!

PR #912, opened 2015-10-09
libass: update to 0.13.0

avih:


tonytheodore:
Thanks!

avih:
np.

FWIW, this version of libass doesn't require fontconfig anymore, except on windows XP. It can still be built with fontconfig, but by default it will only use it if it can't DirectWrite at runtime.

So when MXE decides than it can drop XP support, libass can drop fontconfig as a dependency.

tonytheodore:
[Noted, thanks](https://github.com/mxe/mxe/commit/8300f39b75e64896bbd9e15ced2dd1867feac5f2).

avih:
This version also drops ENCA support, so the `--disable-enca` can/should be removed as well (though it doesn't complain if it's there). I verified this locally.

I also just read the libass changelog and MXE's libass.mk, and as far as I can tell, beyond enca,  there shouldn't be more stuff to do here.

Apologies for not having the XP comment and the ENCA thingy as part of the PR.

Issue #921, opened 2015-10-11
MXE recongizes file src/lua-rapidjson-1-foo.patch as a patch for lua


starius:
MXE recongizes file `src/lua-rapidjson-1-use-unix-file-functions.patch` as a patch for `lua`, however it was intended to be a patch for package `lua-rapidjson`.

```
$ ls src/*.mk | grep -c -- -
38
```

There are 38 packages with `-` in names. Prefixes of some of them are equal to names of other packages: `gcc`, `gcc-isl`.

starius:
Testing commit 466c7841a4e6e8b05904a32c72d52c48b6447b39 with `build-pkg.lua`.

PR #923, opened 2015-10-12
require patch number to start with a digit

starius:
MXE recongized file `src/lua-rapidjson-1-use-unix-file-functions.patch` as a patch for `lua`, however it was intended to be a patch for `lua-rapidjson`. Patch filename pattern was `src/$(1)-*.patch`. `*` captured "rapidjson-1-use-unix-file-functions".

Currently `*` was replaced with `[0-9]*` to require patch name starting with a digit.

Tested on all packages of `i686-w64-mingw32.static` target.

see #921


tonytheodore:
Thanks!

PR #924, opened 2015-10-13
fix travis builds for pull requests

starius:


tonytheodore:
Thanks!

starius:
Please, enable Travis builds for pull requests.
By the way a build of `master` branch fails if there is nothing to commit in `versions.json`

tonytheodore:
@TimothyGu could you look at this?

TimothyGu:
@starius, I'll take a look at reenabling Travis for PRs, but why? The only purpose for Travis right now is for updating the gh-pages branch and versions.json, neither of which is applicable for PRs.

On the other hand, the reason for current build failure is because you removed `|| true` in the script (unless I'm missing something).

starius:
@TimothyGu, I want Travis to check Makefile syntax of Pull Requests.

Currently it does fail in `make versions.json` if something is wrong in `Makefile` or `src/*.mk` (it is a good part), but it can also fail in `git commit` if it has nothing to commit (bad part). If nothing changes, it should not fail in `git commit`. But it should fail if `git push` failed (that is why that `|| true` was too permissive).

TimothyGu:
No, I meant `|| true` just for the commit command.

tonytheodore:
@starius, I think Travis uses Github for permissions - I've never logged in to it before but can see a switch for pull requests.

starius:
I also see it. Shall I switch "Build pull requests" to ON?

tonytheodore:
Yes, it's the best way to see what happens.

starius:
Done.

tonytheodore:
Working! https://travis-ci.org/mxe/mxe/pull_requests

Issue #919, opened 2015-10-11
Toolchain re-org


tonytheodore:
See:
 - http://lists.nongnu.org/archive/html/mingw-cross-env-list/2015-09/msg00026.html
 - http://lists.nongnu.org/archive/html/mingw-cross-env-list/2015-09/msg00019.html
 - #852
 - #898
 - #918

ooxi:
I'll be glad to give you a helping hand, is there anything I can do to support this issue? :)

starius:
Can `src/cmake/modules` path (or it's installed copy in `usr/`) be included in the package `mxe-conf`?

tonytheodore:
It could be installed by `mxe-conf`, it would work and even `git` wouldn't notice that we installed it multiple times. This may be the best short-term solution to #918 though we still need do get the rest right.

As an aside, I'd lean towards including all mxe sources for reference (and possibly as a legal requirement of changes made).

ooxi:
I strongly agree with installing the mxe sources from which the binary packages were build. This would also eliminate possible incompatibilities between new commits and existing binary packages.

starius:
I'm going to add new common package `mxe-src` to `build-pkg`, which includes `src` directory. All other packages depend on `mxe-src` as well as on `mxe-requirements`.

> possibly as a legal requirement of changes made

I have overlooked this issue. GPL'ed packages require distributing source along with binary packages. Is it worth adding package `mxe-pkg` with directory `pkg`?

tonytheodore:
> GPL'ed packages require distributing source along with binary packages

IANAL, but I believe you only need to make the sources available on request. Otherwise, no modern OS could ship on (say) a CD.

> Is it worth adding package mxe-pkg with directory pkg?

I would say no, especially not as a requirement. Maybe as a `recommends` package, but no other normal debian packages do that.

starius:
AFAIK `deb-src` is what binary packages are built from. You can say `apt-get source mxe && cd mxe* && debuild` and it build all binary packages for you.

tonytheodore:
> Can src/cmake/modules path (or it's installed copy in usr/) be included in the package mxe-conf?

@starius, indeed, having the native mxe-conf install these modules in usr/ is a very good idea that I've included in the pull request. (Installing the mxe sources is mostly orthogonal to this).

> I'll be glad to give you a helping hand, is there anything I can do to support this issue? :)

@ooxi, test out #925 and see if it behaves as expected.

[...]

Issue #516, opened 2014-10-02
Add mechanism for depending on build system packages


TimothyGu:
Right now, packages that depend on libraries on build system are done in a very inefficient manner. For example, `gcc` depends on gcc-cloog, etc. This means that these packages that are shared across targets are rebuilt when a new target is specified.

One way to make it more efficient is using the `build-requirements` mechanism, and enable native build for regular cloog package. Then, implement a dependency system where something like this is possible:

``` make
$(PKG)_DEPS := cloog_$(BUILD)
```

tonytheodore:
> `$(PKG)_DEPS := cloog_$(BUILD)`

Down the track we may add a mechanism for explicit inter-target deps for packages, but at the moment deps at a target level seem sufficient (#925).

tonytheodore:
Target deps aren't sufficient as they cause rebuilds for unrelated packages and may cause unnecessary packages to be built.

tonytheodore:
Also, `make download-foo` doesn't traverse target deps:
http://lists.nongnu.org/archive/html/mingw-cross-env-list/2015-12/msg00005.html

tonytheodore:
Implemented in #1990

PR #925, opened 2015-10-13
Toolchain re-org - target dependencies

tonytheodore:
Following on from #919...

This implements a fairly straightforward two-phase build where cross targets depend on the native target. Visibility is lost about `gcc` depending on `gmp` et. al. (#516), but at some point in the toolchain you just have to rely on implied dependencies.

The ordering dependencies like `binutils --> pkgconf` and `<most pkgs> --> gcc` remain, these can't be removed until the use cases for `phases` are fully understood (requires attempts at clang and glibc).

Native builds are enabled by specifying:
```
$(PKG)_TARGETS  := $(BUILD) [$(MXE_TARGETS)]
```
but otherwise there's no change in normal usage or *.mk files.

There's no simple way to alter some `gcc` install directories, so it's `mv` and `rm`.

The main impact will be on `build-pkg.lua` where an ordered build is required. I've added an initial 
```
make print-build-order [MXE_TARGETS=...]
```
target that is the most reliable way (I know) to determine what `make` thinks it needs to do. @starius, I'd recommend moving the logic from `getPkgs` and `sortForBuild` into the main `Makefile` (if not the creation of tarballs and manifests altogether).


starius:
> `<most pkgs> --> gcc`

typo? Should be `gcc --> <most pkgs>`

starius:
Package `gcc` is compiled with the following log message:
```
[build]     gcc                    i686-w64-mingw32.static
```

Can you print phase as a part of this log message?

tonytheodore:
> typo? Should be gcc --> <most pkgs>

Probably just a matter of context. If talking about dependencies I'd go left to right in the sense "most pkgs depend on gcc", build order would be the reverse of that "gcc is built before most pkgs".

> Can you print phase as a part of this log message?

There aren't explicit phases at the moment, just targets. The native phase is `x86_64-unknown-linux-gnu`, other targets (`*-w64-*`) depend on that and can be thought of as the second or cross phase.

> There's no simple way to alter some gcc install directories, so it's mv and rm.

Only one package (libgsf shared) seems to have a problem with this - I'm looking at it.

tonytheodore:
> Only one package (libgsf shared) seems to have a problem with this - I'm looking at it.

Unrelated [`sed` issue] (https://github.com/mxe/mxe/commit/a530b79d5a9257588debf6fe0a17fdccc442dc68).

starius:
> [build]     gcc                    i686-w64-mingw32.static

 ,

> There aren't explicit phases at the moment, just targets. The native phase is x86_64-unknown-linux-gnu, other targets (*-w64-*) depend on that and can be thought of as the second or cross phase.

So `gcc` belongs to cross-phase, doesn't it?

starius:
Native packages create files `usr/<target>/installed/isl` as well as `usr/x86_64-unknown-linux-gnu/installed/isl`.

```
$ ls usr/*/installed/isl                           
usr/i686-w64-mingw32.static/installed/isl
usr/x86_64-unknown-linux-gnu/installed/isl
usr/x86_64-w64-mingw32.static/installed/isl
```

Is it a bug or a feature?

tonytheodore:
> So gcc belongs to cross-phase, doesn't it?

Yes, though that can be misleading, think of it as phase two. As I mentioned, phases don't have names yet, just target dependencies.

> Is it a bug or a feature?

A feature as the native isl will only be built once instead of multiple times and clobbering the previous build when it was simply under `usr/`. It also gives us a place to do things like install `config.guess` and the cmake modules only once in the native `mxe-conf`.

Running:
```
$ make isl MXE_TARGETS=`ext/config.guess`
[using autodetected 6 job(s)]
[check requirements]
[build]     mxe-conf               x86_64-apple-darwin14.5.0
[done]      mxe-conf               x86_64-apple-darwin14.5.0                              312 KiB        0m4.637s
[build]     gmp                    x86_64-apple-darwin14.5.0
[done]      gmp                    x86_64-apple-darwin14.5.0                              26840 KiB      0m34.185s
[build]     isl                    x86_64-apple-darwin14.5.0
[done]      isl                    x86_64-apple-darwin14.5.0                              29416 KiB      0m21.104s
```
and there is no reference to the cross targets. Running `make isl` will build all of `gcc` and it's dependents and cross versions of `gmp` and `isl`.

In `build-pkg.lua`, you can remove most of the COMMON_FILES:
```diff
diff --git a/tools/build-pkg.lua b/tools/build-pkg.lua
index 7a5df56..2ac4928 100755
--- a/tools/build-pkg.lua
+++ b/tools/build-pkg.lua
@@ -49,37 +49,9 @@ local BLACKLIST = {
 }
 
 local COMMON_FILES = {
-    ['gcc-isl'] = {
-        '^usr/include/isl/',
-        '^usr/lib/libisl%.',
-        '^usr/lib/pkgconfig/isl.pc$',
-    },
-    ['gcc-mpc'] = {
-        '^usr/include/mpc.h$',
-        '^usr/lib/libmpc%.',
-    },
-    ['gcc-gmp'] = {
-        '^usr/include/gmp.h$',
-        '^usr/lib/libgmp%.',
-    },
-    ['gcc-mpfr'] = {
-        '^usr/include/mpf2mpfr.h$',
-        '^usr/include/mpfr.h$',
-        '^usr/lib/libmpfr%.',
-    },
-    ['gcc'] = {
-        '^usr/lib/libcc1%.',
-    },
-    ['yasm'] = {
-        '^usr/include/libyasm',
-        '^usr/lib/libyasm.a$',
-    },
     ['ncurses'] = {
         '^usr/lib/pkgconfig/',
     },
-    ['pkgconf'] = {
-        '^usr/bin/config.guess$',
-    },
 }
 
 local ARCH_FOR_COMMON = 'i686-w64-mingw32.static'
```

Hope that makes sense.

tonytheodore:
> This implements a fairly straightforward two-phase build

Maybe I should elaborate on this. Phase one (the "native" phase) consists of seven packages: 
 - the `gcc` deps (`gmp`, `isl`, `mpc`, and `mpfr`)
 - our custom `pkgconf` (with `Cflags.private` support)
 - `yasm` (the assembler)
 - the implied `mxe-conf` (for once-off setup)

All these packages currently also exist in phase two:
 - `gmp`, `mpc`, and `mpfr` as requirements of other cross-packages
 - `isl` (why not?)
 - `pkgconf` (to create wrapper scripts)
 - `yasm` (to create wrapper scripts for "yasm the assembler", and for cross-`libyasm`)
 - the implied `mxe-conf` (for target-specific setup)

Arguably, `isl` and `libyasm` can be removed from phase two until someone or some package requires it. All other packages behave the same as before.

One package that (seemingly) will exist only in phase one is (the future) `cmake`. Surely `mxe-conf` in phase two will configure it correctly?

It does so at the moment, but if we added `cmake`, I would move the configuration from `mxe-conf` into a phase two `cmake` - probably adding a "fake" dependency in `binutils` (as with `pkgconf`).

This is far from ideal, but packages should configure themselves. We're not planning on delivering a `cmake.exe` in the near future - that would require an n-phase build.

> A feature as the native isl will only be built once instead of multiple times

In an n-phase build, we could also build `gcc` only once in a multi-lib configuration. We don't need to build it four times, but the subtleties and complexities of doing that are staggering. I'd like to see how far we can go with a simple two-phase build.

starius:
Changes for build-pkg required:
  * remove most of [common packages](https://github.com/mxe/mxe/pull/925#issuecomment-147992049)
  * fix getting dependency information. `make print-build-order` is not sufficient, because it provides only build order, while `build-pkg` needs a list of dependencies to write them to Debian package. `build-pkg` generates temporary mk-file to get dependencies. This workaround was written when `build-pkg` was not a part of MXE. Now the content of this temporary mk-file can be merged into Makefile as new target.

I tried `build-pkg` with this pull request as is and it works. The only bug I see is that all native packages are squashed into `mxe-conf` package. It generated 4 shared packages and all of them are empty:
  * common-gcc.list
  * common-ncurses.list
  * common-pkgconf.list
  * common-yasm.list

tonytheodore:
Something like [`print-deps-for-build-pkg`](https://github.com/tonytheodore/mxe/commit/c7c16495a76d0848d6947209e8a61da7d6570106) may work in the short term. It lets you refer to packages as `gmp~x86_64-unknown-linux-gnu`, I've used a tilde but the separator can be anything unlikely to be in a package name.

starius:
To track cross-target dependencies (`i686~foo` depends on `native-bar`), target information is required in output of `print-deps-for-build-pkg`. I like the idea of adding a target and `~`. Can you modify `print-deps-for-build-pkg` to output dependency graph for all targets at once and add targets to names, please? A package could appear multiple times, with different targets. `build-pkg.lua` would process such full-target dependencies instead of target agnostic dependencies and build all items in one loop.

tonytheodore:
Okay, [now it prints `target~pkg`](https://github.com/tonytheodore/mxe/commit/d3e66480da3b818d7897a6d1df67fc7a0a07fa0a), but calling `make` with `pkg~target` will also work. It's still based on `MXE_TARGETS` in case you want a subset but full deps on the native target will always be shown.

starius:
Patches for build-pkg.lua after toolchain re-org:
https://gist.github.com/starius/deb0902a35bf5ef1a6b2
This patch was applied to tonytheodore:toolchain merged with mxe:master (5c61e160beb2b10ca501c4b2e87d4c467f313b3a).
I'm building all packages with patched build-pkg. Please review the code.

tonytheodore:
Nice work!, I added a line to keep the Debian control files:

```diff
--- a/tools/build-pkg.lua
+++ b/tools/build-pkg.lua
@@ -435,6 +435,8 @@ local function makeDeb(item, list_path, deps, ver)
         -- make .deb file
         local cmd = 'fakeroot -i deb.fakeroot dpkg-deb -b %s'
         os.execute(cmd:format(dirname))
+    else
+        os.execute(('cp %s %s.deb-control'):format(control_fname, dirname))
     end
     -- cleanup
     os.execute(('rm -fr %s deb.fakeroot'):format(dirname))
```

and those along with the `*list` files all look good to me (using MXE_MAX_ITEMS=100). A couple of comments.

https://gist.github.com/starius/deb0902a35bf5ef1a6b2#file-mxe-reorg-build-pkg-patch-L662
> return not isInString('unknown', target)

A check against the output of `ext/config.guess` would probably be better.

https://gist.github.com/starius/deb0902a35bf5ef1a6b2#file-mxe-reorg-build-pkg-patch-L684
> because it seems to work faster with lesser MXE_TARGETS.

I've been working on some cleanup and speedups for this in another branch that should be ready in the next few days.

https://gist.github.com/starius/deb0902a35bf5ef1a6b2#file-mxe-reorg-build-pkg-patch-L710
> Subject: [PATCH] build-pkg: log cross-target dependencies
>
>Don't log dependencies on native target.

not sure about this last patch, it prints a lot of lines and I'm not sure if they are supposed to be warnings. What is it trying to do?

tonytheodore:
I'm currently installing the cmake modules in `usr/share/cmake` but `usr/share` is blacklisted. Should they be installed somewhere else or change the blacklist?

starius:
**Warning**: New toolchain builds `usr/i686-w64-mingw32.static/qt5/lib/libQt5Bootstrap.a` natively (detected by build-pkg). I'll post full build log when it finishes.

New patches: https://gist.github.com/starius/931118a42fb7793b8b74

> I added a line to keep the Debian control files:

You have not pushed this change. I have included it to part 2 of the patch.
I like the idea of keeping Debian control files. Why do you keep them only if `not no_debs`? 

> I'm currently installing the cmake modules in usr/share/cmake but usr/share is blacklisted. Should they be installed somewhere else or change the blacklist?

I prefer changing the blacklist. Path `usr/share` was unlisted and paths `usr/share/{doc,gcc*,info,man}` were listed.

> A check against the output of ext/config.guess would probably be better.

Done

> I've been working on some cleanup and speedups for this in another branch that should be ready in the next few days.

`make pkg MXE_TARGETS=target` keeps working so I prefer this variant because it is more stable.

> not sure about this last patch, it prints a lot of lines and I'm not sure if they are supposed to be warnings. What is it trying to do?

It produces a warning when one cross-target (e.g. `i686-w64-mingw32.static`) depends on another cross-target (e.g. `i686-w64-mingw32.shared`), which is obviously an error. It doesn't produce a warning when a cross-target depends on a native target (`x86_64-unknown-linux-gnu`).

My build process is in progress but it has produced no such warnings.

starius:
The patch 2 was updated (forgot to `trim` output of `shell`).
https://gist.github.com/starius/931118a42fb7793b8b74

tonytheodore:
> Warning: New toolchain builds usr/i686-w64-mingw32.static/qt5/lib/libQt5Bootstrap.a natively 

I noticed that in the [last log also](https://gist.githubusercontent.com/starius/825d9d81d46f4186fa75/raw/624599880e2d16d1503e9eb352c8775fdf559b2c/build-17Sep2015-issues.log). I imagine it's for bootstrapping the native `qmake` etc. and can likely be deleted if those tools are static (which they appear to be).

> I like the idea of keeping Debian control files. Why do you keep them only if not no_debs?

Mostly because I'm testing on my fast machine (running OSX) without creating debs. I agree, they are very useful and can always be kept - I've changed the commit.

> make pkg MXE_TARGETS=target keeps working so I prefer this variant because it is more stable.

Yes, that will always work.

> It doesn't produce a warning when a cross-target depends on a native target (x86_64-unknown-linux-gnu).

Fixed now with using `config.guess` for the native target.

> The patch 2 was updated (forgot to trim output of shell).

Thanks! I've pushed those along with some `.gitignore` and `make clean` additions.

tonytheodore:
> "make target~pkg" doesn't work if target is not in MXE_TARGETS

I've [removed the phony `pkg~target` and `target~pkg` targets](https://github.com/tonytheodore/mxe/commit/5a310127c4526e4d768feb97750941156f4523d5). 

```
make pkg MXE_TARGETS=target
```

Is again the only recommended way.

starius:
Logs:
Wheezy: https://gist.github.com/starius/e9e32a4d206fd268fb0f
Jessie: https://gist.github.com/starius/c242cd70b967ba8f6baf
Wheezy's log doesn't have `[downloads]` entries because packages had been already downloaded.
No packages are broken.

tonytheodore:
Thanks! `gendef` looks like the last one related to this and [it's an easy fix](https://github.com/tonytheodore/mxe/commit/ded1936b32ad158ad25513c2224ee013e65f694e).

tonytheodore:
> I've been working on some cleanup and speedups for this in another branch

Should I add the [speedups](https://github.com/tonytheodore/mxe/commit/4eb8955acaa5cc29516617764e9f5489e0cf1788) to this or is this ready for merging and I'll open a new request?

starius:
I think speeding up the Makefile is an unrelated change and this pull request is ready for merging.

I've built MXE twice: without this pull request and with it.
Here are changes of the file list made by #925: https://gist.github.com/starius/3b5ae26b804f9820e97b
"<" stands for removed files, ">" stands for added files.
Please analyse this list before merging.

tonytheodore:
> Please analyse this list before merging.

Everything is as intended in that list, just a couple of comments:
 - log4cxx is using a hash for filenames that would be expected to change 
 - `libcc1.so` has other issues as it is also an `*.so` on OSX
 - `vstasm` and `ytasm` are different front-ends for `yasm` that aren't currently used
 - doc exclusion generally needs some focus

> I think speeding up the Makefile is an unrelated change

I've removed the speedup commit and rebased, it could do with some re-factoring for readability. Having `print-deps-for-build-pkg` is a good sanity check.

> and this pull request is ready for merging

Thanks for all your help with this! It gives us a lot of flexibility having an additional dependency mechanism and the packaging is all nicely self-contained.

Issue #922, opened 2015-10-12
blas-3.5.0.tgz file changed on upstream


starius:
See http://lists.nongnu.org/archive/html/mingw-cross-env-list/2015-10/msg00009.html

The upstream was notified with the same message.
As for commit be0ce5f6f60759657333472a4b87c41f39b6e8a6 package `blas` can't be downloaded. For new version `SUBDIR` was changed from `BLAS` to `BLAS-3.5.0`. The following patch gets old (working) version of `blas.tgz` to be downloaded:
```diff
diff --git a/src/blas.mk b/src/blas.mk
index af92d2a..cdc8e88 100644
--- a/src/blas.mk
+++ b/src/blas.mk
@@ -7,8 +7,8 @@ $(PKG)_VERSION  := 1
 $(PKG)_CHECKSUM := ea87df6dc44829ee0a1733226d130c550b17a0bc51c8dbfcd662fb15520b23b5
 $(PKG)_SUBDIR   := BLAS
 $(PKG)_FILE     := $(PKG).tgz
-$(PKG)_URL      := http://www.netlib.org/$(PKG)/$($(PKG)_FILE)
-$(PKG)_URL_2    := ftp://ftp.eq.uc.pt/pub/software/math/netlib/$(PKG)/$($(PKG)_FILE)
+#$(PKG)_URL      := http://www.netlib.org/$(PKG)/$($(PKG)_FILE)
+$(PKG)_URL      := ftp://ftp.eq.uc.pt/pub/software/math/netlib/$(PKG)/$($(PKG)_FILE)
 $(PKG)_DEPS     := gcc
 
 define $(PKG)_UPDATE
```

I propose to find HTTP version of old version and replace `$(PKG)_URL` with it (because [FTP doesn't work](http://lists.nongnu.org/archive/html/mingw-cross-env-list/2014-07/msg00002.html) on some machines). After getting answer from the upstream, switch to new version.

TimothyGu:
For reference:

```diff
diff -Naurx '._*' BLAS/Makefile BLAS-3.5.0/Makefile
--- BLAS/Makefile	2007-04-05 13:59:57.000000000 -0700
+++ BLAS-3.5.0/Makefile	2015-10-07 17:50:12.000000000 -0700
@@ -88,7 +88,7 @@
 #  Level 2 and Level 3 BLAS.  Comment it out only if you already have
 #  both the Level 2 and 3 BLAS.
 #---------------------------------------------------------------------
[...]

PR #926, opened 2015-10-13
update & fix blas

starius:
Transient fix for #922
see http://lists.nongnu.org/archive/html/mingw-cross-env-list/2015-10/msg00009.html

Tested against whole `i686-w64-mingw32.static` target.

### DIFF
--- a/src/blas.mk
+++ b/src/blas.mk
@@ -3,12 +3,11 @@
 
 PKG             := blas
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 1
-$(PKG)_CHECKSUM := ea87df6dc44829ee0a1733226d130c550b17a0bc51c8dbfcd662fb15520b23b5
-$(PKG)_SUBDIR   := BLAS
+$(PKG)_VERSION  := 3.5.0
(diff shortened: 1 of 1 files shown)


tonytheodore:
Thanks!

kolewu:
shouldn't there be a change in version.json, too?

```diff
--- a/versions.json
+++ b/versions.json
@@ -15,7 +15,7 @@
     "bfd": "2.25.1",
     "binutils": "2.25.1",
     "bison": "2.7.1",
-    "blas": "1",
+    "blas": "3.5.0",
     "boost": "1.57.0",
     "box2d": "2.3.0",
     "bullet": "2.82-r2704",
```

tonytheodore:
The MXEBot does that automatically https://github.com/mxe/mxe/commit/efdf84d620ce88843363c23adc6e67f092110428

kolewu:
Ahh, so when making a patch with a different version I must not touch `version.json`?

tonytheodore:
I don't think it matters if you do, but you shouldn't go out of your way to do so. It's essentially derived metadata and you don't (or shouldn't) need to think about it.

PR #929, opened 2015-10-14
add variable to define directory for tmp-directories

kolewu:
sometimes it's necessary to not build inside the mxe folder. The
variable MXE_TMP can be used to change the base directory for creating
tmp directories that are used for building.

Rationale: when using virtualbox to build the cross-compile environment
and the mxe project is in a shared folder, at least stripping is not
working because of an internal error of the shared filesystem.  So the
solution is to change the tmp-directory to a place inside the VM.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,8 @@ PKG_CDN            := d1yihgixbnrglp.cloudfront.net
 PWD        := $(shell pwd)
 SHELL      := bash
 
+MXE_TMP := $(PWD)
+
 BUILD_CC   := $(shell (gcc --help >/dev/null 2>&1 && echo gcc) || (clang --help >/dev/null 2>&1 && echo clang))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -223,6 +225,12 @@ else
         echo '# within one package ("intra-package parallelism").'; \
         echo '#JOBS := $(JOBS)'; \
         echo; \
+        echo '# This variable controls where intermediate files are created'; \
+        echo '# this is necessary when compiling inside a virtualbox shared'; \
+        echo '# directory. Some commands like strip fail in there with »Protocol error'; \
```
starius: Fix quotes near `»`

starius:
:+1: 
Let me build all packages with `MXE_TMP=/tmp` to make sure nothing was broken.

PR #931, opened 2015-10-14
use better type for pointer cast

kolewu:
there is an integer type that is guaranteed to be big enough to hold a pointer (but it's only available since C++11 and C99):
* [Fixed width integer types (since C++11) - cppreference.com](http://en.cppreference.com/w/cpp/types/integer])
* [Fixed width integer types (since C99) - cppreference.com](http://en.cppreference.com/w/c/types/integer])

### DIFF
--- a/src/openscenegraph-3-fix-pointer-to-int-conversion-on-64bit.patch
+++ b/src/openscenegraph-3-fix-pointer-to-int-conversion-on-64bit.patch
@@ -14,12 +14,12 @@ diff --git a/src/osgPlugins/osgjs/WriteVisitor.cpp b/src/osgPlugins/osgjs/WriteV
 index 9f2b3c7..98e0c19 100644
 --- a/src/osgPlugins/osgjs/WriteVisitor.cpp
 +++ b/src/osgPlugins/osgjs/WriteVisitor.cpp
-@@ -225,7 +225,7 @@ JSONObject* createImage(osg::Image* image, bool inlineImages, int maxTextureDime
+@@ -225,8 +225,7 @@ JSONObject* createImage(osg::Image* image, bool inlineImages, int maxTextureDime
              // no image file so use this inline name image and create a file
              std::stringstream ss;
              ss << osgDB::getFilePath(baseName) << osgDB::getNativePathSeparator();
 -            ss << (long int)image << ".inline_conv_generated.png"; // write the pointer location
-+            ss << (long long int)image << ".inline_conv_generated.png"; // write the pointer location
++            ss << (intptr_t)image << ".inline_conv_generated.png"; // write the pointer location
              std::string filename = ss.str();
              if (osgDB::writeImageFile(*image, filename)) {
                  image->setFileName(filename);


saiarcot895:
Wouldn't `uintptr_t` be better, since pointers are unsigned?

TimothyGu:
@saiarcot895 pointers themselves are not signed or unsigned (see C11 6.5.8.5). Plus, the original code uses signed.

starius:
> pointers themselves are not signed or unsigned

But `intptr_t` is signed and can be negative, printing as a negative number. I would prefer a file with name `123.png` rather than `-123.png`.

```c++
#include <stdint.h>
#include <iostream>

int main() {
    void* ptr = 0;
    ptr -= 1;
    std::cout << "intptr_t: " << (intptr_t)ptr << std::endl;
    std::cout << "uintptr_t: " << (uintptr_t)ptr << std::endl;
}
```

[Output](https://ideone.com/TOHvs5):
```
intptr_t: -1
uintptr_t: 4294967295
```

> Plus, the original code uses signed

The original code uses wrong type and doesn't even compile. The type it uses should not be taken into account.

tonytheodore:
I don't have a strong opinion on this, but pointers look like ordinals, and long/int look like cardinals.

starius:
Interesting topic :)

[Ordinal](https://en.wikipedia.org/wiki/Ordinal_number) in math means something like "index" and [cardinal](https://en.wikipedia.org/wiki/Cardinal_number) means something like "size", doesn't it?

In modern C++ type `int` is intended to be used as an index and type `size_t` as a size. The latter was introduced exactly not to use `int` as both index and size.

In this case, a number generated from a pointer means neither "index" nor "size", it is used as a random token to add it to a filename. It is the same that PID is used. Using positive numbers is more convenient.

TimothyGu:
```c
    void* ptr = 0;
    ptr -= 1;
```

I'm pretty sure this is undefined behavior…

TimothyGu:
That code snippet seems to have two wrongs:

1. you can't do pointer arithmetic on `void *`: it's illegal in C (but legal in GCC): http://stackoverflow.com/a/3524270
2. Subtracting a number from a pointer without knowing there's something before it is undefined: see C11 6.5.6.8 (unchanged from C99):

 > If both the pointer operand and the result point to elements of the same array object, or one past the last element of the array object, the evaluation shall not produce an overflow; otherwise, the behavior is undefined.

kolewu:
But I think the subtraction is not meant as some pointer arithmetic, but rather as a means to get the binary representation of the type with all bits set. Because of the two's complement the biggest unsigned number is just `-1` for the corresponding signed representation (and fortunately this is working with gcc).

Back to the original problem: I think `uintptr_t`is a better fit here in theory. When really working on this problem upstream, I would suggest to use the hex representation of the corresponding pointer to get a unique file name -- but to only make it compile `uintptr_t`is good enough.

TimothyGu:
> But I think the subtraction is not meant as some pointer arithmetic, but rather as a means to get the binary representation of the type with all bits set. Because of the two's complement the biggest unsigned number is just `-1` for the corresponding signed representation

Realistically, yes it does work this way. But in C standard, a pointer is not an integer, and therefore the rules with unsigned integers don't work with pointers. Plus, the provided C file explicitly uses an arithmetic operator against a pointer, so the argument that it is not pointer arithmetic doesn't stand.

But this is all moot. I'll change the type to `uintptr_t` to prevent theoretical undefined behaviors on 32-bit platforms.

starius:
> you can't do pointer arithmetic on void *: it's illegal in C (but legal in GCC): http://stackoverflow.com/a/3524270 

Good catch! Clang produces "error: arithmetic on a pointer to void". This can be fixed by replacing `void` with `char`.

> Subtracting a number from a pointer without knowing there's something before it is undefined

Is the following code valid?

```c++
    intptr_t p1 = -1;
    void* p2 = reinterpret_cast<void*>(p1);
    char* ptr = reinterpret_cast<char*>(p2);
```

TimothyGu:
TBH, I don't know. The C standard (6.3.2.3.5) says:

> An integer may be converted to any pointer type. Except as previously specified, the result is implementation-defined, might not be correctly aligned, might not point to an entity of the referenced type, and might be a trap representation.

So it is implementation-defined.

tonytheodore:
> But this is all moot.

Good call @TimothyGu.

Issue #932, opened 2015-10-14
cmake wrapper doesn't work with --build option


starius:
`cmake` has option `--build` which can be used instead of calling `make`:
```
$ cmake . -Bbuild-dir
-- The C compiler identification is GNU 4.7.2
-- The CXX compiler identification is GNU 4.7.2
-- Check for working C compiler: /usr/bin/gcc
...
$ cmake --build build-dir/
[  5%] Generating qrc_qfoo.cxx
[ 10%] Generating src/view/moc_MainWindow.cxx
...
OK
```

MXE CMake wrapper calls `cmake` with option `-DCMAKE_TOOLCHAIN_FILE`. This doesn't work with `--build` options, since `cmake` reconfigures build directory instead of building.

```
$ i686-w64-mingw32.static-cmake  . -Bbuild-dir
-- The C compiler identification is GNU 5.2.0
-- The CXX compiler identification is GNU 5.2.0
-- Check for working C compiler: /mxe/usr/bin/i686-w64-mingw32.static-gcc
-- Check for working C compiler: /mxe/usr/bin/i686-w64-mingw32.static-gcc -- works
...
$ i686-w64-mingw32.static-cmake --build build-dir/
-- Configuring done
-- Generating done
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_TOOLCHAIN_FILE


-- Build files have been written to: .../build-dir
```

PR #933, opened 2015-10-14
cmake wrapper: don't pass -DCMAKE_TOOLCHAIN_FILE to cmake when  with --build

starius:
see #932

### DIFF
--- a/src/mxe-conf.mk
+++ b/src/mxe-conf.mk
@@ -46,9 +46,21 @@ define $(PKG)_BUILD_COMMON
      ) > '$(CMAKE_TOOLCHAIN_FILE)'
 
     #create prefixed cmake wrapper script
-    (echo '#!/bin/sh'; \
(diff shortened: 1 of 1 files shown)


tonytheodore:
Looks good, just two points:
 - use `#!/usr/bin/env bash` for portability
 - the `-P` mode also doesn't use the toolchain file (and probably others)

starius:
`-P` works with the toolchain file.
```
$ cat 111.cmake 
message("123")
$ cmake -P 111.cmake 
123
$ /mxe/usr/bin/i686-w64-mingw32.static-cmake -P 111.cmake 
123
```

I am checking [other options](https://cmake.org/cmake/help/v2.8.8/cmake.html#section_Options).

starius:
Fixed.

tonytheodore:
Thanks!

PR #934, opened 2015-10-15
Fix OpenSceneGraph build for 64-bit shared.

saiarcot895:
The include header case and the external library linking should apply to all build types. The pointer type applies to 64-bit builds.

Tested on 64-bit shared.


saiarcot895:
Never mind. I just saw this was fixed a few days back.

PR #935, opened 2015-10-15
MXE Plugins

tonytheodore:
Following on from #881 and #925, this removes the optional native builds and provides a new way to handle such things.

Basic usage is to drop some `*.mk` files in a directory and set `MXE_PLUGIN_DIRS='path1 path2 ...'`. Building and updates work as expected, but I haven't tested heavily enough to expose any edge cases.

```
$ make keepassx MXE_PLUGIN_DIRS='plugins/apps' -j4
[using autodetected 6 job(s)]
[check requirements]
[build]     mxe-conf               x86_64-apple-darwin14.5.0
[done]      mxe-conf               x86_64-apple-darwin14.5.0                              312 KiB        0m7.361s
[build]     gmp                    x86_64-apple-darwin14.5.0
[build]     pkgconf                x86_64-apple-darwin14.5.0
...
[done]      freetds                i686-w64-mingw32.static                                138528 KiB     1m39.181s
[done]      postgresql             i686-w64-mingw32.static                                228680 KiB     1m6.607s
[build]     qt                     i686-w64-mingw32.static
[done]      qt                     i686-w64-mingw32.static                                1593572 KiB    11m48.140s
[build]     keepassx               i686-w64-mingw32.static
[done]      keepassx               i686-w64-mingw32.static                                71752 KiB      0m16.743s
```

Plugins are self-contained and should have the following lines (depending on what they're doing):

```
$(PKG)_WEBSITE  := http://www.gnu.org/software/automake
$(PKG)_MAKEFILE := $(lastword $(MAKEFILE_LIST))
$(PKG)_OWNER    := https://github.com/tonytheodore
```

Since the plugins are just makefiles, users will have almost unlimited flexibility to modify rules and keep those changes under source control (without having merge conflicts etc.). For the plugins in our tree, I've [tried to highlight that they aren't supported](https://github.com/tonytheodore/mxe/tree/plugins/plugins).

Naming, of course, can be changed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,12 +170,18 @@ UNPACK_ARCHIVE = \
 UNPACK_PKG_ARCHIVE = \
     $(call UNPACK_ARCHIVE,$(PKG_DIR)/$($(1)_FILE))
 
-PATCHES = $(sort $(wildcard $(TOP_DIR)/src/$(1)-[0-9]*.patch))
+# some shortcuts for awareness of MXE_PLUGIN_DIRS
+# plugins will need to set their own $(PKG)_MAKEFILE for updates
+# all files for extension plugins will be considered for outdated checks
+PKG_MAKEFILE  = $(realpath $(or $($(1)_MAKEFILE),$(TOP_DIR)/src/$(1).mk))
+PKG_MAKEFILES = $(realpath $(sort $(wildcard $(addsuffix /$(1).mk, $(TOP_DIR)/src $(MXE_PLUGIN_DIRS)))))
+PKG_TESTFILES = $(realpath $(sort $(wildcard $(addsuffix /$(1)-test*, $(TOP_DIR)/src $(MXE_PLUGIN_DIRS)))))
+PKG_PATCHES   = $(realpath $(sort $(wildcard $(addsuffix /$(1)-[0-9]*.patch, $(TOP_DIR)/src $(MXE_PLUGIN_DIRS)))))
 
 define PREPARE_PKG_SOURCE
     cd '$(2)' && $(call UNPACK_PKG_ARCHIVE,$(1))
     cd '$(2)/$($(1)_SUBDIR)'
-    $(foreach PKG_PATCH,$(PATCHES),
+    $(foreach PKG_PATCH,$(PKG_PATCHES),
         (cd '$(2)/$($(1)_SUBDIR)' && $(PATCH) -p1 -u) < $(PKG_PATCH))
 endef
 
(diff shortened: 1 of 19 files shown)


tonytheodore:
> Since the plugins are just makefiles

Added [an example of overriding rules and deps](https://github.com/tonytheodore/mxe/blob/17b25213f2c0e9ddf90d18d05b8c6b0a6a6f2c49/plugins/custom-qt-min/overrides.mk). The uses cases are old, but it would be hard to offer this level of customisation in any other way.

tonytheodore:
> `$(PKG)_MAKEFILE := $(lastword $(MAKEFILE_LIST))`

Changed to `$(realpath $(lastword $(MAKEFILE_LIST)))`

starius:
Can plugins redefine properties of normal packages?

The following use cases are of interest for me:

  * change gcc version https://github.com/mxe/mxe/pull/965#issuecomment-152827660
  * use LuaJIT instead of PUC-Lua as Lua implementation ([luabind issue](https://github.com/mxe/mxe/pull/579))
  * change the set of Qt components installed https://github.com/mxe/mxe/issues/967#issuecomment-153550595

These package customizations need changing the set of patches.
Can a plugin redefine the set of patches applied to the package?

tonytheodore:
> Can plugins redefine properties of normal packages?

Yes, they are read after all normal packages so can extend or override those.

> * change gcc version #965 (comment)
> * use LuaJIT instead of PUC-Lua as Lua implementation (luabind issue)

These may be the same underlying use case and the usage mentioned in #579 seems likely:
```
make MXE_CC=[gcc-49 | gcc-52 | clang] MXE_LUA=[lua | luajit]
```
Though lua seems more like it belongs in the normal set of packages.

> change the set of Qt components installed #967 (comment)

See [example](https://github.com/tonytheodore/mxe/blob/3f0cbd596fa091bf3249e58a58e4194e39c19f7e/plugins/custom-qt-min/overrides.mk) from [earlier comment](https://github.com/mxe/mxe/pull/935#issuecomment-148458362). Qt comes up often over the years and this would allow people to keep their modifications in their own source tree and version control.

> Can a plugin redefine the set of patches applied to the package?

Yes, plugins have their own set of patches that can be either additions or replacements. For the most part though, plugins would have different names (say `gcc-52`) and only use those.

tonytheodore:
Added `gcc52` overlay style plugin. This style is useful when multiple packages are affected, though it has the downside of not receiving updates. For the packages involved in this case, that isn't an issue.

I'll merge this for the time being to get some feedback on usage - there are many ways it can be used and t's only a single commit to revert if there are any major issues (I haven't seen any yet).

starius:
Thanks! I'm going to rebase my [luarocks](https://github.com/LuaAndC/mxe/tree/luarocks) branch using plugins.

tonytheodore:
Great! Time to revisit #808 and #812 also.

starius:
How to build all packages from a particular plugin?

tonytheodore:
You can't really, other than creating a [virtual package for deps like qt5](https://github.com/mxe/mxe/blob/master/src/qt5.mk) that uses `ls -1` instead of `grep`.

starius:
LuaRocks plugin: https://github.com/mxe/mxe/pull/1017

rr-:
@tonytheodore this really needs better documentation - there are no instructions how to use plugins on neither http://mxe.cc or any of the READMEs.

tonytheodore:
> this really needs better documentation

Agreed, it's actually at the top of my (wish)list.

PR #940, opened 2015-10-16
Decrease cmake version and check for policies used

starius:


tonytheodore:
Thanks, should we revert https://github.com/mxe/mxe/commit/f1dd70802a4e20c0f8c849aa44f24651895c7cfc also?

starius:
> Thanks, should we revert f1dd708 also?

I have tried reverting f1dd708 and it works. Please revert.

tonytheodore:
Done https://github.com/mxe/mxe/commit/d08148a0da54811afb71c767dc4b6f225f038917.

PR #714, opened 2015-06-14
Reenable TCL/Tk by disabling SQLLite dependency

highperformancecoder:
TCL was originally disabled because of a compiler build error in the SQL lite component.
SQLLite functionality is not an important feature of TCL, and can be disabled via a configuration variable.
TCL/Tk is an important component technology for projects of mine that use MXE, so I will be maintaining TCL/Tk indefinitely, at least within my branch.

Only commits 848c83ea3fb99101196a8c14409bc5bcef2ee5ee and dbe181a01dc796a9ca19b6c0611a2130d86a4f4b are needed for this pull request. I can't seem to specify which commits to include in a pull request via Github's pull request create form.

5dbde9debfb897a8e40352531eb109018e30e0c4 is for a compile problem I found in qdbm.

PR #941, opened 2015-10-16
Reenable TCL/Tk by disabling SQLLite dependency

tonytheodore:
See #714


tonytheodore:
A couple of quick observations:
 - seems to require native `tcl` to build
 - build on all four targets succeeds
 - no option to disable shared builds or easily control installation directories.
 - not sure removing the sqlite dependency does anything, it seems to build it's own:
```
configure: WARNING: unrecognized options: --enable-static, --without-sqlite
Building package 'sqlite3.8.0'
```
 - has options for postgres and mysql - these should be enabled.
 - most likely is outside the scope of what would be included in MXE.

tonytheodore:
See #948

PR #942, opened 2015-10-16
add package libmemcached

starius:
libMemcached is an open source C/C++ client library and tools for the memcached server (http://danga.com/memcached). It has been designed to be light on memory usage, thread safe, and provide full access to server side methods.


src/libmemcached.mk
```diff
@@ -0,0 +1,27 @@
...
+define $(PKG)_UPDATE
+    $(WGET) -q -O- 'https://launchpad.net/libmemcached/+download' | \
+    $(SED) -n 's,.*libmemcached-\([0-9][^>]*\)\.tar.*,\1,p' | \
+    head -1
+endef
+
+define $(PKG)_BUILD
+    cd '$(1)' && ./configure \
+        $(MXE_CONFIGURE_OPTS) \
+        '--with-mysql=$(PREFIX)/$(TARGET)/bin/mysql_config'
```
tonytheodore: Small nitpick - why quote the whole option?

starius: Fixed

starius:
Problems:
  * `-lmemcachedutil` is included to *.pc file, but doesn't exist
  * tests and examples don't build

tonytheodore:
Looks like the windows build is supposed to be minimal, from `Makefile.am:40`:

```
# Build targets for Windows
if BUILD_WIN32
include libhashkit-1.0/include.am
include libhashkit/include.am
include libmemcached/include.am
include libmemcached-1.0/include.am
else
```
It even excludes the include `win32/include.am` which sets `LDFLAGS+=-no-undefined`. Could be a lot of fiddling to tidy it up.

PR #944, opened 2015-10-18
change Wt homepage to http://www.webtoolkit.eu

starius:
witty.sourceforge.net redirects to www.webtoolkit.eu


tonytheodore:
Thanks!

PR #945, opened 2015-10-19
add CMake module FindFriBiDi.cmake

starius:
Now the following CMake code finds FriBiDi:

```cmake
  find_package(FriBiDi)
```


tonytheodore:
Thanks!

Issue #331, opened 2014-02-21
openmp


jboulanger:
Hi,

I had hard time to cross compile code using openmp.

I found discussion (2009) about the support of openmp on the mailing list but the current status was unclear. 

By default the gcc.mk has openmp disabled (--disable-libgomp)  and the default thread model is win32 (which i read is incompatible with gcc  implementation of libgomp). I found using 
        --enable-threads=posix \
        --enable-libgomp \
- adding pthreads as a dependency working (although a cyclic dependency warning is triggered).

An enhancement would be to have the possibility to enable openmp as an global option... if this makes sense. At least document these steps.

Otherwise many thanks for this great project.

PR #958, opened 2015-10-27
Building pthreads inline with gcc and enable optional posix threads in libgcc

tonytheodore:
Firstly, build pthreads inline and enable libgomp:
* fixes "~winpthreads changes ~pthread_signal.h" etc. reported by `build-pkg`.
  mingw-w64 installs dummy headers if winpthreads isn't built
* enables libgomp to avoid double-build (see #331)
* no change in openmp-validation (still 20 failures - taken with a grain of
  salt as the batch file test runner isn't a reliable perl substitute)
* pthreads virtual package kept for future testing

Then enbale `MXE_GCC_THREADS` and allow target spec `<triplet>.<link>.<threads>`:
* enables optional c++11 `<thread>, <mutex>, and <future>` in libgcc
* avoid [double build advice for posix threads](http://sigrok.org/gitweb/?p=sigrok-util.git;a=blob;f=cross-compile/mingw/README;h=248af546f939a81855dda33142d834eecf2979c2;hb=HEAD#l44)
* not heavily tested but should be released for feedback


tonytheodore:
See [the Qt wiki](https://wiki.qt.io/MinGW-64-bit#GCC_Threading_model_.28posix_vs_win32.29) and [stack overflow question](http://stackoverflow.com/questions/17242516/mingw-w64-threads-posix-vs-win32) for some discussion on threading.

Maybe we should do the exception variants at the same time?

starius:
Thank you! Let me build all packages to make sure that nothing was broken.

> Maybe we should do the exception variants at the same time?

These changes seem to be unrelated.

tonytheodore:
> Let me build all packages to make sure that nothing was broken.

Thanks, I saw some issues with cegui, ffmpeg, and opencsg but can't replicate them - seem to be affected by some other package being built earlier.

>> Maybe we should do the exception variants at the same time?

> These changes seem to be unrelated.

Agreed, enabling them isn't hard but we have no way of testing every combination. It's also not yet clear if threads should be options or defaults.

starius:
opencv is [broken](https://gist.githubusercontent.com/starius/4e7f7999ba416d7f6bc0/raw/12b904bddffd146b9b1f447b935519babc0daadb/opencv_i686-w64-mingw32.static)

starius:
The build is completed.
`grep build-pkg log` > https://gist.github.com/starius/01db21c22331a2701f7c
Only opencv and cimg (which depends on opencv) are broken.

tonytheodore:
> opencv is broken

It was [missing the `mingww-64` patch](https://github.com/tonytheodore/mxe/commit/55e9c2451d626daf0ae6738d4d5216f5ed605215).

starius:
Let me build all packages again.

starius:
All packages were built successfully.

tonytheodore:
Thanks! I'll merge this for now so we can start experimenting with target specs.

Issue #946, opened 2015-10-19
build-pkg should install mxe sources


starius:
see #918
see https://github.com/mxe/mxe/issues/919#issuecomment-147727375

tonytheodore:
I wonder if we should create corresponding git tags for every packaging release?

starius:
> I wonder if we should create corresponding git tags for every packaging release?

Good idea. Maybe even Github releases to have a place where to upload log and maybe some other information about a release.

PR #959, opened 2015-10-27
build-pkg: install source, refactoring and other improvements

starius:
see #946


tools/build-pkg.lua
```diff
@@ -552,36 +563,57 @@ local function makeMxeRequirementsDeb(release)
...
+        'Makefile',
+        'README.md',
+        'assets',
+        'doc',
+        'ext',
+        'index.html',
+        'settings.mk',
+        'src',
+        'tools',
+        'usr',
```
tonytheodore: `settings.mk` and `usr` should be excluded.

starius: fixed

tonytheodore:
Thanks!

PR #963, opened 2015-10-29
build-pkg: print progress

starius:
Example:
 
```
[  4/100] The build is expected to complete in 0.3 hours, on Thu Oct 29 23:03:27 2015
```

### DIFF
--- a/tools/build-pkg.lua
+++ b/tools/build-pkg.lua
@@ -59,8 +59,12 @@ local TARGETS = {
     'x86_64-w64-mingw32.shared',
 }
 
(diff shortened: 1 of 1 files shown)


starius:
Please review.

tonytheodore:
Looks good! The handling of "huge" packages is a nice approach. If it becomes hard to maintain, we can add a `build-stats.json` (or similar) as a basis for [standard build units](http://www.linuxfromscratch.org/lfs/view/development/chapter04/aboutsbus.html).

Issue #964, opened 2015-10-30
g++ produces broken executables with -O2 (pubsetbuf)


starius:
Code:
```cpp
#include <fstream>

const int BUFFER_SIZE = 4096;
char buffer_[BUFFER_SIZE];

struct BufferedIfstream : public std::ifstream {
    BufferedIfstream() {
        rdbuf()->pubsetbuf(buffer_, BUFFER_SIZE);
    }
};

int main() {
    BufferedIfstream fff;
    return 0;
}
```

Compilation command:
```bash
$ ./usr/bin/i686-w64-mingw32.static-g++ mingw-pubsetbuf-bug.cpp -O2 -o mingw-pubsetbuf-bug.exe
```

Run:
```bash
$ wine mingw-pubsetbuf-bug.exe
Application tried to create a window, but no driver could be loaded.
Make sure that your X server is running and that $DISPLAY is set correctly.
wine: Unhandled page fault on write access to 0x7edd0000 at address 0x4951b8 (thread 0009), starting debugger...
Application tried to create a window, but no driver could be loaded.
Make sure that your X server is running and that $DISPLAY is set correctly.
Unhandled exception: page fault on write access to 0x7edd0000 in 32-bit code (0x004951b8).
Register dump:
 CS:0023 SS:002b DS:002b ES:002b FS:0063 GS:006b
 EIP:004951b8 ESP:00d0fc00 EBP:00d0fd80 EFLAGS:00010202(  R- --  I   - - - )
 EAX:ffffffff EBX:7edd0000 ECX:00d0fd5c EDX:00d0fd14
 ESI:00111e38 EDI:00000015
Stack dump:
0x00d0fc00:  00000000 00d0fbc0 0000000c 00000001
0x00d0fc10:  00111e38 00000001 00000001 004afca1
0x00d0fc20:  004bc844 0000ffff 00d0fc58 7bc3ef72
0x00d0fc30:  00000040 7bc36596 00d0fc60 00000000
0x00d0fc40:  7bc36686 0000000d 00000000 7bc36686
0x00d0fc50:  00d0fc70 004af400 004b01a0 00d0fd70
Backtrace:
=>0 0x004951b8 in mingw-bug (+0x951b8) (0x00d0fd80)
  1 0x004afca1 in mingw-bug (+0xafca0) (0x00d0fd80)
  2 0x7b83a4c6 SetUnhandledExceptionFilter+0x5() in kernel32 (0x00000000)
[...]

Issue #961, opened 2015-10-28
Build gcc once


starius:
See https://github.com/mxe/mxe/pull/925#issuecomment-147992049

tonytheodore:
Agreed, this is the right time to implement such a change. If we want to do threading and exception variants, we'll end up building gcc sixteen times when large parts are common.

The implementation I see is based around multi-lib compilers, wrapper scripts, and spec files. The spec files may also provide an entry point for customisation.

tonytheodore:
The multi-lib configuration ends up being fairly complicated and reduces flexibility of varying threading/exception between architectures. It also doesn't really save much in the way of build time, only a third is common, the rest is built twice.

Spec files may still be useful, though again the extra complexity seems to outweigh the benefits. I think these should only be used for minor tweaks.

tonytheodore:
See also #1133, arch-independent files should be disabled or configured to install correctly.

tonytheodore:
See also https://github.com/mxe/mxe/pull/1165#issuecomment-177636523, `nsis` should also be built once and possibly linked across i686 and x86_64 targets (until it can be built for x86_64).

PR #965, opened 2015-10-31
Downgrade gcc to 4.9

starius:
See #964
See da82d07dbeca23087aa4351914cf4b890c1d3972
See c96a007c6a4b03ca8d43c2ed848c47832f9e2baf
See 7882944eb99fa14bb82f3045896668a68f8c68a7

### DIFF
--- a/index.html
+++ b/index.html
@@ -1155,6 +1155,10 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
         <td class="website"><a href="http://acoustid.org/chromaprint">Chromaprint</a></td>
     </tr>
     <tr>
+        <td class="package">cloog</td>
+        <td class="website"><a href="http://www.cloog.org/">CLooG Code Generator</a></td>
+    </tr>
+    <tr>
         <td class="package">cimg</td>
         <td class="website"><a href="http://cimg.eu/">CImg Library</a></td>
     </tr>
--- a/src/cloog.mk
+++ b/src/cloog.mk
@@ -1,0 +1,34 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := cloog
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 0.18.1
+$(PKG)_CHECKSUM := 02500a4edd14875f94fe84cbeda4290425cb0c1c2474c6f75d75a303d64b4196
+$(PKG)_SUBDIR   := $(PKG)-$($(PKG)_VERSION)
+$(PKG)_FILE     := $(PKG)-$($(PKG)_VERSION).tar.gz
+$(PKG)_URL      := http://www.bastoul.net/cloog/pages/download/$($(PKG)_FILE)
+$(PKG)_URL_2    := ftp://gcc.gnu.org/pub/gcc/infrastructure/$($(PKG)_FILE)
+$(PKG)_TARGETS  := $(BUILD) $(MXE_TARGETS)
+$(PKG)_DEPS     := gcc gmp isl
+
+$(PKG)_DEPS_$(BUILD) := gmp isl
+
+# stick to tested versions from gcc
+define $(PKG)_UPDATE
+    $(WGET) -q -O- 'ftp://gcc.gnu.org/pub/gcc/infrastructure/' | \
+    $(SED) -n 's,.*cloog-\([0-9][^>]*\)\.tar.*,\1,p' | \
+    $(SORT) -V |
+    tail -1
+endef
+
+define $(PKG)_BUILD_$(BUILD)
+    cd '$(1)' && ./configure \
+        $(MXE_CONFIGURE_OPTS) \
+        --with-gmp-prefix='$(PREFIX)/$(TARGET)' \
+        --with-isl-prefix='$(PREFIX)/$(TARGET)'
+    $(MAKE) -C '$(1)' -j '$(JOBS)'
+    $(MAKE) -C '$(1)' -j '$(JOBS)' install
+endef
+
+$(PKG)_BUILD_SHARED =
--- a/src/gcc.mk
+++ b/src/gcc.mk
@@ -3,8 +3,8 @@
 
 PKG             := gcc
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 4.9.2
-$(PKG)_CHECKSUM := 2020c98295856aa13fda0f2f3a4794490757fc24bcca918d52cc8b4917b972dd
+$(PKG)_VERSION  := 4.9.3
+$(PKG)_CHECKSUM := 2332b2a5a321b57508b9031354a8503af6fdfb868b8c1748d33028d100a8b67e
 $(PKG)_SUBDIR   := gcc-$($(PKG)_VERSION)
 $(PKG)_FILE     := gcc-$($(PKG)_VERSION).tar.bz2
 $(PKG)_URL      := http://ftp.gnu.org/pub/gnu/gcc/gcc-$($(PKG)_VERSION)/$($(PKG)_FILE)
--- a/src/gcc.mk
+++ b/src/gcc.mk
@@ -3,8 +3,8 @@
 
 PKG             := gcc
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 5.2.0
-$(PKG)_CHECKSUM := 5f835b04b5f7dd4f4d2dc96190ec1621b8d89f2dc6f638f9f8bc1b1014ba8cad
+$(PKG)_VERSION  := 4.9.2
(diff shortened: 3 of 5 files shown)


tonytheodore:
I'm not sure this is a good approach as we'll never know when we can upgrade. I think it's better to investigate further and submit upstream reports.

If someone wants to lock a gcc version, [plugins](https://github.com/mxe/mxe/pull/935) are a better way to manage that, and if we can isolate a specific issue (say `-fno-devirtualize`) then we can control that with [spec files](https://github.com/mxe/mxe/issues/961#issuecomment-152827095).

starius:
`libmysqlclient` is broken for this pull request ([log](https://gist.githubusercontent.com/starius/71d5f276d62e58f36d2f/raw/c564fc4fcce5777a5ca0af338b32606983873af2/libmysqlclient_i686-w64-mingw32.static)).
It can be interesting even if we won't downgrade to 4.9.

starius:
> as we'll never know when we can upgrade

When gcc-5 becomes stable. Stable releases of Linux distributions (Debian, OpenSuse, gentoo) use gcc <= 4.9. A bug in compiler is a bug in every other package. Branch 5 of GCC has [too many bugs](https://gcc.gnu.org/bugzilla/buglist.cgi?cf_known_to_fail=5.2.1&cf_known_to_fail_type=allwords&cf_known_to_work_type=allwords&product=gcc&query_format=advanced). Bugs in gcc-5 keep being detected and fixed. If we want gcc-5, we should track and backport all reported bugs carefully, as Debian maintainers do. Gcc 4.9 is less buggy than gcc 5. Does any package in MXE use features of gcc-5 not available in gcc-4.9?

tonytheodore:
> When gcc-5 becomes stable. Stable releases of Linux distributions (Debian, OpenSuse, gentoo) use gcc <= 4.9

It's a bit of chicken-and-egg, how will it become stable if the major distros don't give it exposure? gcc 4.9 also [had issues](http://lkml.iu.edu//hypermail/linux/kernel/1407.3/00650.html) earlier in it's cycle.

> Does any package in MXE use features of gcc-5 not available in gcc-4.9?

Very unlikely, it would be the latest versions if so and many packages need updating. We've spent more time chasing down bugs than looking at those ;)

From the [gcc homepage](https://gcc.gnu.org/) it looks like gcc-6 is already in the works and the 5 series may not have a very long life. If no one objects in the next day or so I'll merge this and see about making the 5 series a plugin.

mabrand:
My doubts are in line with Tony's first comments on this. Ubuntu 15.10 uses gcc 5.2.1 which suggests that gcc 5 is at least somewhat reliable.

https://github.com/mxe/mxe/issues/964 mentions the possibility that gcc-1.fixes could be to blame. Shouldn't that be investigated first anyway?

> Does any package in MXE use features of gcc-5 not available in gcc-4.9?

I'm not sure this should be the deciding criterion. Whether or not MXE packages already exploit gcc 5 features, users of MXE might very well be interested in the features and improvements of gcc 5.

I  think the criterion for downgrading should be whether programs are much more likely to be broken by gcc 5 than gcc 4. The fact that Ubuntu 15.10 has adopted gcc 5 makes it seem doubtful that this is the case. Are the problems with gcc 5 much more likely to affect mingw than linux?

starius:
> #964 mentions the possibility that gcc-1.fixes could be to blame. Shouldn't that be investigated first anyway?

The reason is [the gcc bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=68175). The MXE patches do not affect this bug.

This bug was erroneously marked as a duplicate of [another similar bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67056), but the fix proposed for that bug doesn't fix compilation of my code. The bug has not been updated for a week. It is typical for a development branch not for stable.

There is another bug mentioned in [the mailing thread](http://lists.nongnu.org/archive/html/mingw-cross-env-list/2015-10/msg00020.html), which seems to be an unrelated bug in gcc.

> Ubuntu 15.10 uses gcc 5.2.1 which suggests that gcc 5 is at least somewhat reliable.

Unfortunately Ubuntu often updates packages to unstable version too early. New Ubuntu releases (e.g., Ubuntu 15) are based on Debian Sid (=unstable). [Stable](https://packages.debian.org/jessie/gcc) and [testing](https://packages.debian.org/stretch/gcc) Debian releases use gcc 4.9.

As an example, Travis CI uses old Ubuntu with gcc 4.6 (or optionally Ubuntu 14 with gcc 4.8).

> I think the criterion for downgrading should be whether programs are much more likely to be broken by gcc 5 than gcc 4.

There are at least [two](https://github.com/mxe/mxe/issues/964) [real](http://lists.nongnu.org/archive/html/mingw-cross-env-list/2015-10/msg00020.html) cases when a program of MXE user is broken because of not fixed gcc-5 bug. I know no such cases for gcc-4.9

mabrand:
I'm satisfied with that explanation. Writing software is hard enough when
you can trust the compiler.

starius:
libmysqlclient fixed

starius:
PR rebased

starius:
Help needed with https://github.com/mxe/mxe/commit/c96a007c6a4b03ca8d43c2ed848c47832f9e2baf#commitcomment-14323213

starius:
I'm building commit 44ad791

starius:
Commit 44ad79155db61170fb55d931ee3d030a14a8f5fc was built. No errors occurred.
https://gist.github.com/starius/d23568a542766398bc6f

tonytheodore:
Great! Thanks for all your work on this - in the absence of a stable branch, the core toolset will benefit from a slower cycle and will give us some leeway to work on infrastructure and packages to make the overall MXE experience better.

tm604:
Note that gcc < 5.0 means some C++14 features will no longer be available:

https://gcc.gnu.org/projects/cxx1y.html

of these, the one I use most is the [variable templates](https://isocpp.org/files/papers/N3651.pdf) support. Supporting a stable version of gcc seems like a sensible goal, just concerned that rolling back to 4.9 after so long with 5.0+ being available is quite a big step backwards.

Having gcc5 available via config, separate package or plugin support would be great if that's possible?

tonytheodore:
> Having gcc5 available via config, separate package or plugin support would be great if that's possible?

Yes, it's certainly possible and on the short-term agenda - the main issue is which of those options is the most suitable?

tm604:
> > Having gcc5 available via config, separate package or plugin support would be great if that's possible?
> Yes, it's certainly possible and on the short-term agenda - the main issue is which of those options is the most suitable?

I don't think gcc5 is ABI-compatible with 4.9, so I'd guess that a separate package would introduce some problems of its own. Perhaps either a variable like MXE_GCC_VERSION (similar to MXE_GCC_THREADS) or a separate MXE_TARGET would be more appropriate?

avih:
> Great! Thanks for all your work on this - in the absence of a stable branch, the core toolset will benefit from a slower cycle and will give us some leeway to work on infrastructure and packages to make the overall MXE experience better.

Hmm... maybe this would be a good opportunity to create a stable branch again? Or approach it by using "subject-specific experiments" branches (in this case - GCC 5 branch)?

Possibly with some well defined policy to keep the branches as close to identical as possible, except for the subject being experimented with (GCC 5 here, maybe something else in the future too), with a stated goal to merge them as soon as possible?

tonytheodore:
> Perhaps either a variable like MXE_GCC_VERSION (similar to MXE_GCC_THREADS) or a separate MXE_TARGET would be more appropriate?

At the moment, I'm leaning towards both a variable and a pseudo target in relation to #961. My WIP so far has normal MXE targets depending on a toolchain target along the lines of `i686-w64-minw32.multi`. I'm no longer using the multi-lib approach, so these can be along the lines of:
```
i686-w64-mingw32.gcc49
i686-w64-mingw32.gcc52
i686-w64-mingw32.clang
```
Normal use would be a variable like MXE_USE_CC, people building side-by-side (say for testing) would use `i686-w64-mingw32.static.posix.gcc52` etc.

> Possibly with some well defined policy to keep the branches as close to identical as possible, except for the subject being experimented with (GCC 5 here, maybe something else in the future too), with a stated goal to merge them as soon as possible?

This is much easier to do with selectively overriding rules via a "plugin" mechanism (#935). There's no merge conflicts and experiments can be very long-lived. You can also do multiple experiments simultaneously which requires a lot of mental housekeeping to do with `git`.

tonytheodore:
See https://github.com/mxe/mxe/commit/5e63e416b166d4e7855e88176a21126bc381bb3d.

It's the simplest thing that works, no side-by-side, alternate targets, or spec files. You can test this with:
```
make MXE_PLUGIN_DIRS=plugins/gcc52
```

PR #974, opened 2015-11-08
add package ucl

starius:
See also #813

**UCL** is a portable lossless data compression library written in ANSI C.
http://www.oberhumer.com/opensource/ucl/


tonytheodore:
Thanks!

Issue #969, opened 2015-11-05
disable WGET in builds


starius:
Set `WGET=false` while building to prevent occasional downloads, which are not verified with checksums. Better solution is to switch off all networking for the building process (as we [did](https://github.com/mxe/mxe/commit/8e2d0e85c5dcdecbe858ffbca391dcc5ca0c3527) for wine) but I don't know how.

tonytheodore:
> Set WGET=false while building to prevent occasional downloads, which are not verified with checksums. 

What downloads are these? I can't say I've noticed them.

starius:
> What downloads are these?

I [used](https://github.com/mxe/mxe/pull/812/files#diff-f14c34aab0433118ea71788dfc3d5223R21) this approach to download GeoIP.dat.gz needed for qbittorrent (rejected pull request #812). I think that breaking network access for build commands helps to prevent such errors in future.

BTW what would be the right solution for GeoIP.dat.gz file? Normal file could be downloaded as a package, but this file is updating frequently.

tonytheodore:
Creating a package for GeoIP data seems reasonable, the license is suitable and monthly updates aren't an issue. It may also be useful to other packages.

starius:
Another source of unattended downloads are smart package building tools. Some of them could detect missing dependency and download it (instead of failing the build). For example. `luarocks` downloads and installs missing dependencies of the rock being built. It is easy to oversee such a problem, if the new version of the rock being built gets a new dependency.

tonytheodore:
I'm not sure I fully understand the problem, unattended downloads sound like a feature. Is the issue that these downloads don't have checksums? Would dropping a `tar` wrapper script in `usr/bin` (first in PATH) be better than disabling networking?

starius:
> unattended downloads sound like a feature

These downloads may be non-verified (e.g., `luarocks` doesn't verify downloads generally), so the build result (success of fail) is unpredictable. For example, luarocks rock `foo` depends on luarocks rock `bar`. If `foo` is a part of MXE, than foo's sources are verified, but bar's sources are not. The source of bar is just downloaded from latest version of bar on the luarocks server. So installing `foo` causes implicit installation of **latest published version** of bar. If the source is not frozen, than the build result is unpredictable. This situation may be the result of updating foo, if previous version of foo had no dependency of bar and it was overseen when updating.

starius:
> Would dropping a tar wrapper script in usr/bin (first in PATH) be better than disabling networking?

[...]

PR #977, opened 2015-11-10
disable networking with LD_PRELOAD when build

starius:
see #969


tonytheodore:
Nice approach to disabling networking, it provides a clear separation of activities. I think the equivalent on OSX is `DYLD_INSERT_LIBRARIES` - I'll see if I can make it more portable.

I do use `git` in build rules sometimes - mostly since there isn't another way to do such things at the moment.

starius:
> I do use git in build rules sometimes - mostly since there isn't another way to do such things at the moment.

Do you mean `git clone`? I can't find it with `grep`.
This pull request breaks no packages.

tonytheodore:
> Do you mean git clone? I can't find it with grep.

`clone` and/or `pull`. I use the `Makefile` as a general purpose automator from time to time and it's unrelated to this. We could handle those cases by recognising the `git://` url scheme in `$(PKG)_URL` or by adding a `$(PKG)_REPO` variable to handle `svn`, `hg`, `bzr` etc.

I'll create a separate issue for OSX.

PR #980, opened 2015-11-11
build-pkg: add EOL after last line in *.list file

### LINKED ISSUE #979: build-pkg doesn't print EOL after last item of *.list file
`wc` prints wrong number of lines for *.list files
See https://github.com/mxe/mxe/pull/975#issue-115985014

starius:
fix #979


tonytheodore:
Thanks!

PR #984, opened 2015-11-12
new package: x265

darealshinji:


src/x265.mk
```diff
@@ -0,0 +1,65 @@
...
+        -DENABLE_ASSEMBLY=$$X265_ASM \
+        -DEXTRA_LIB="x265_main10.a;x265_main12.a" \
+        -DEXTRA_LINK_FLAGS=-L. \
+        -DLINKED_10BIT=ON \
+        -DLINKED_12BIT=ON
+
+    $(MAKE) -C '$(1)/8bit' -j '$(JOBS)' install
+    $(INSTALL) '$(1)/8bit/libx265_main12.a' '$(PREFIX)/$(TARGET)/lib/libx265_main12.a'
+    $(INSTALL) '$(1)/8bit/libx265_main10.a' '$(PREFIX)/$(TARGET)/lib/libx265_main10.a'
+    $(SED) -i 's|-lx265|-lx265 -lx265_main10 -lx265_main12|' '$(PREFIX)/$(TARGET)/lib/pkgconfig/x265.pc'
```
TimothyGu: No, the `-DEXTRA_LIB` line above already added the 10- and 12-bit symbols to the libx265.a archive.

src/x265.mk
```diff
@@ -0,0 +1,65 @@
...
+    $(WGET) -q -O- https://ftp.videolan.org/pub/videolan/x265/ | \
+    $(SED) -n 's,.*">x265_\([0-9][^<]*\)\.t.*,\1,p' | \
+    tail -1
+endef
+
+define $(PKG)_BUILD
+    cd '$(1)' && mkdir -p 8bit 10bit 12bit
+
+    # 12 bit
+    test '$(TARGET)' = 'x86_64-w64-mingw32.static' && X265_ASM=ON || X265_ASM=OFF; \
```
TimothyGu: Why? I believe x265's assembly optimization works on 32-bit as well.

darealshinji: Unfortunately it doesn't work anymore.
https://bitbucket.org/multicoreware/x265/issues/150/linux-x86-32-bit-yasm-error

src/x265.mk
darealshinji: Are you sure? Afaik that only adds them to the CLI.

src/x265.mk
```diff
@@ -0,0 +1,65 @@
...
+# See index.html for further information.
+
+PKG             := x265
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 1.8
+$(PKG)_CHECKSUM := d41a3f0cc06dfc3967f6c47f458cb30b6aaa518f86c56c147946395bfe22b6f2
+$(PKG)_SUBDIR   := x265_$($(PKG)_VERSION)
+$(PKG)_FILE     := x265_$($(PKG)_VERSION).tar.gz
+$(PKG)_URL      := https://ftp.videolan.org/pub/videolan/$(PKG)/$($(PKG)_FILE)
+$(PKG)_DEPS     := gcc cmake yasm
```
tonytheodore: `cmake` is built automatically and shouldn't be listed in DEPS

src/x265.mk
tonytheodore: The HIGH_BIT_DEPTH options are not supported on 32-bit, do we really want to enable them?

http://hg.videolan.org/x265/file/f8b8ebdc5457/source/CMakeLists.txt#l302

src/x265.mk
tonytheodore: > Afaik that only adds them to the CLI.

That seems correct (checking with `nm`), all three need to be installed. Note that we don't build executables, just the libraries.

tonytheodore:
I've got the shared builds (mostly) working and I'll use this as an example of how we separate the core mxe (libs only) with the relaxed "handy tools" available via plugins. Leave it with me.

src/x265.mk
darealshinji: Disabling everything which is not fully supported on 32 bit builds sounds reasonable to me.

tonytheodore: I've got a new `option(USE_HIGH_BIT_DEPTH_SUFFIX "Name libraries according to bit depth" OFF)` etc. working so we can just install them and let people decide if they want to use them.

darealshinji:
I've updated the files and added a test. Now shared libraries are build too. The library name however is `libx265.dll` without the API.
I'm not so sure on how to rebase this because of the master branch changes I've merged in. I could make a new PR if you really want only 1 commit.

src/x265.mk
```diff
@@ -0,0 +1,74 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := x265
+$(PKG)_IGNORE   :=
```
tonytheodore: add WEBSITE and use single line `# This file is part of MXE. See LICENSE.md for licensing information.`

src/x265.mk
```diff
@@ -0,0 +1,74 @@
...
+    cd '$(BUILD_DIR)' && mkdir -p 10bit 12bit
+
+    # 12 bit
+    cd '$(BUILD_DIR)/12bit' && $(TARGET)-cmake '$(SOURCE_DIR)/source' \
+        -DHIGH_BIT_DEPTH=ON \
+        -DEXPORT_C_API=OFF \
+        -DENABLE_SHARED=OFF \
+        -DENABLE_ASSEMBLY=$(if $(findstring x86_64,$(TARGET)),ON,OFF) \
+        -DENABLE_CLI=OFF \
+	    -DWINXP_SUPPORT=ON \
```
tonytheodore: extra indent

darealshinji:
File updated.

tonytheodore:
Thanks, normally would prefer to avoid `sed` but there's not much that can be done in this case.

PR #992, opened 2015-11-14
Update to OS X requirements instructions

stjosh:
- update Xcode version to 7
- added 'cmake' and 'gsed' to the list of ports to install
- mark 'build-requirements' step as optional

### DIFF
--- a/index.html
+++ b/index.html
@@ -728,7 +728,7 @@ USE_OSGPLUGIN(&lt;plugin2&gt;)
 
     <p>
     Install
-    <a href="http://developer.apple.com/xcode/">Xcode 5</a>
+    <a href="http://developer.apple.com/xcode/">Xcode 7</a>
     </p>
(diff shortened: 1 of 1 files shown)


starius:
> update Xcode version to 7

Does Xcode version depend on OSX version? Can we just require "Xcode" without a version? Or "Xcode >= 5"?

> mark 'build-requirements' step as optional

I am not sure that Step 2 is optional for older OSX versions. It is optional if `cmake` is installed.

Could you squash all changes to one commit, please? Also, avoid merges in a pull request.

tonytheodore:
> Can we just require "Xcode" without a version?

This would be better.

> It is optional if cmake is installed.

Actually, at the moment, `cmake` is the only package that should be built. The version from macports breaks `cgal` and `vmime`. I'll take a look at it tomorrow and update the docs.

stjosh:
ok cool, thanks @tonytheodore .

PR #999, opened 2015-11-15
disable wine

### LINKED ISSUE #995: dlfcn-win32 finds wine
`dlfcn-win32` detects `wine` with the following command:
```
$ wine --version
```
This command succeeds even if WINEPREFIX is set to an empty or read-only directory:

```
$ WINEPREFIX=/ wine --version
wine-1.6.2
```

I think we should create shell script `wine` returning exit code 1 and add it to `PATH`.

starius:
Now dlfcn-win32 can't detect wine.
From log/dlfcn-win32_i686-w64-mingw32.static:

    static: yes
    shared: no
    wine:

Previously wine was set to "yes wine".

fix #995


tonytheodore:
Nice!

PR #998, opened 2015-11-15
MXE_GET_GITHUB_TAGS: new argument, stripped prefix

starius:

PR #1002, opened 2015-11-17
Standardise github packages

tonytheodore:
Following on from #998, a first pass at standardising the handling of github packages. I had to use the `sha` for downloads as tag/file naming conventions vary too much.

Thoughts?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,9 +130,38 @@ MAKE_SHARED_FROM_STATIC = \
 	--libdir '$(PREFIX)/$(TARGET)/lib' \
 	--bindir '$(PREFIX)/$(TARGET)/bin'
 
+# convenience snippet to standardise github files and updates
+# uses the following package metadata:
+#   GH_REPO    - formatted as owner/repo
+#
+# form one - track branch:
+#   GH_BRANCH  - uses latest sha from branch for updates
+#
+# form two - track tags:
+#   GH_TAG_PFX - optional tag prefix to strip from tags for version sort
+#   GH_TAG_SHA - short sha used in SUBDIR (automatically updated)
+#
+# both forms are exclusive and fetch files based on sha as
+# conventions for tag and file naming vary considerably
+define MXE_SETUP_GITHUB
+    $$(if $$(and $$($$(PKG)_GH_BRANCH),$$($$(PKG)_GH_TAG_SHA)),\
+        $$(error $$(PKG) specifies both GH_BRANCH and GH_TAG_SHA))
+    $$(PKG)_SHA_REF := $$(or $$($$(PKG)_GH_TAG_SHA),$$($$(PKG)_VERSION))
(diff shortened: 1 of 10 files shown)


tonytheodore:
Updated to group the github stanzas.

For releases, this may mean derived/generated artefacts are not present (different checksums), but it will need more input to get a feel for usage, naming conventions, and additional metadata variables.

tonytheodore:
> Does a commit to upstream break MXE build?

In theory, yes - but that's the same for any update. It's also sometimes used to stay on a particular commit (say [pkgconf](https://github.com/mxe/mxe/blob/master/src/pkgconf.mk)).

starius:
`pkgconf` uses a particular commit (da179fd), but no commit is specified if `GH_BRANCH` is used, if I understand correctly. In this case new commit to upstream should be treated as new package version. MXE should not break when new version of an upstream package is released.

tonytheodore:
In the case of `pkgconf`, any commit or subsequent release version breaks MXE (it's something I have to follow up but that's a different issue).

> In this case new commit to upstream should be treated as new package version. 

Yes, that's the mechanism it uses - the short sha is the version. As these don't sort numerically, any given commit may not be seen as new by `make update` but in the scope of things that's not really worth worrying about.

Projects (like [vmime](http://www.vmime.org/downloads.html)) that recommend the use of development branches probably have their own polices that correspond to a "release" policy of other projects.

> MXE should not break when new version of an upstream package is released.

That would be ideal, but we can't do much about it other than update and test - there's no real difference if it's a tarball release or checkout. [`x264` uses snapshots as releases](https://github.com/mxe/mxe/blob/master/src/x264.mk) but nobody has updated it for nearly a year.

tonytheodore:
Note I didn't change all (12?) packages currently using github branches, I wanted to get a feel for usage and naming etc. before doing that.

tonytheodore:
> For GH_BRANCH form, MXE downloads latest commit of the branch.

Now I see, not reading properly. It will always download the (tested) version, it only fetches the latest commit when `make update` is run. Then that will be tested and committed to MXE if successful.

starius:
> It will always download the (tested) version

Ah, I see. I thought `$(PKG)_VERSION` was removed, but it remains.

tonytheodore:
All this does is replace some boilerplate with metadata. I'll look at the other packages to see if any other fields are required.

tonytheodore:
> I had to use the sha for downloads as tag/file naming conventions vary too much.

I'll probably remove this functionality, it only works for lightweight tags, not annotated tags, and I think I've figured out github's file naming convention.

> For releases, this may mean derived/generated artefacts are not present

Typically autotools packages that don't use github for distribution (say [libffi](https://sourceware.org/libffi/)). Hopefully there's not many of these otherwise the exceptions will be outweigh the standardisation.

tonytheodore:
> I'll probably remove this functionality, it only works for lightweight tags, not annotated tags, and I think I've figured out github's file naming convention.

The tag sha is too cumbersome, I think I've got a better way.

PR #1004, opened 2015-11-19
suppress git commit error in Travis

starius:
See #924
See https://travis-ci.org/mxe/mxe/builds/91781012#L99


tonytheodore:
Seems to be [working again with or without version updates](https://travis-ci.org/mxe/mxe/builds). Thanks!

PR #1005, opened 2015-11-19
Coin3D: do not disable features by default

fiesh:
Per https://github.com/mxe/mxe/pull/966#issuecomment-157644451, I have removed the disabling of features.  I tested it with a clean MXE installation that only had gcc installed.

### DIFF
--- a/src/coin.mk
+++ b/src/coin.mk
@@ -23,12 +23,6 @@ define $(PKG)_BUILD
         --disable-debug \
         --disable-symbols \
         --enable-compact \
-        --without-openal \
-        --without-fontconfig \
-        --without-spidermonkey \
-        --without-freetype \
-        --without-zlib \
-        --without-bzip2 \
         --without-x \
         COIN_STATIC=$(if $(BUILD_STATIC),true,false)
 


tonytheodore:
Following on from https://github.com/mxe/mxe/pull/966#issuecomment-158004510, it's better to have explicit deps and `--enable` features.

We don't want to rely on auto-detection, it causes various bugs with build ordering and is (very) hard to maintain in the longer term.

fiesh:
Since coin does everything with run time linking, even with a bare MXE/gcc installation, all features are enabled by default, without any explicit `--enable`.  Is it really meaningful to still add them?

tonytheodore:
Have you got an example of how the run-time linking works? Is it run-time loading of plugins or something else? I'm certain I don't understand what `coin` is.

fiesh:
Sorry, maybe my explanation was sloppy. What I mean is that `coin` uses `dlopen()` (or the windows likes) to dynamically open libraries to support additional features (like supporting bzip2).  Therefore it doesn't require these libraries at build time.

tonytheodore:
Okay, thanks!

Issue #565, opened 2014-11-24
qt (qtassistant?) fails to build with gcc thread model posix


uwehermann:
We're using Qt4 from MXE currently to build the sigrok / PulseView installers for Windows. We also need to use the gcc thread model "posix" (not "win32") since PulseView uses C++11 (and C++11 std::thread).

However, "make qt" fails to build when gcc's thread model is set to "posix". Apparently the qtassistant part has issues. If you comment out the snippet building that in src/qt.mk the rest of Qt seems to build fine.

```
-    cd '$(1)/tools/assistant' && '$(1)/bin/qmake' assistant.pro
-    # can't figure out where -lQtCLucene comes from so use
-    # sed on the output instead of patching the input
-    $(MAKE) -C '$(1)/tools/assistant' sub-lib-qmake_all
-    $(SED) -i 's,-lQtCLucene$$,-lQtCLucene4,' '$(1)/tools/assistant/lib/Makefile.Release'
-    $(MAKE) -C '$(1)/tools/assistant' -j '$(JOBS)' install
+    ## cd '$(1)/tools/assistant' && '$(1)/bin/qmake' assistant.pro
+    ## # can't figure out where -lQtCLucene comes from so use
+    ## # sed on the output instead of patching the input
+    ## $(MAKE) -C '$(1)/tools/assistant' sub-lib-qmake_all
+    ## $(SED) -i 's,-lQtCLucene$$,-lQtCLucene4,' '$(1)/tools/assistant/lib/Makefile.Release'
+    ## $(MAKE) -C '$(1)/tools/assistant' -j '$(JOBS)' install
```

Error log (if you don't do the commenting above):

```
[...]

PR #1009, opened 2015-11-21
Fix MySQL and PostgreSQL posix build

saiarcot895:
Both of these applications assume pthreads isn't available on Windows and redefine those functions, resulting in duplicate definitions.

### DIFF
--- a/src/libmysqlclient-2-posix-fixes.patch
+++ b/src/libmysqlclient-2-posix-fixes.patch
@@ -1,0 +1,86 @@
+This file is part of MXE.
+See index.html for further information.
+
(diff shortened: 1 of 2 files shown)


tonytheodore:
Thanks!

For postgresql, I was [trying to minimise patches](https://github.com/mxe/mxe/issues/565#issuecomment-66451236) but I should have an update to 9.4.5 soon.

tonytheodore:
Unfortunately, this breaks normal builds, reverted for now until further testing https://github.com/mxe/mxe/commit/1a94ae0e6777c841e005366bf0f85b51a3c9f959.

PR #1011, opened 2015-11-26
add cryptopp (https://www.cryptopp.com/)

fiesh:
This patch adds cryptopp, the open source cryptography licensed under Boost Software License 1.0

### DIFF
--- a/index.html
+++ b/index.html
@@ -1197,6 +1197,10 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
         <td class="website"><a href="http://www.freedesktop.org/wiki/Software/cppunit/">CppUnit</a></td>
     </tr>
     <tr>
+        <td class="package">cryptopp</td>
+        <td class="website"><a href="https://www.cryptopp.com/">Crypto++ Library</a></td>
+    </tr>
+    <tr>
         <td class="package">crystalhd</td>
         <td class="website"><a href="http://www.broadcom.com/support/crystal_hd/">Broadcom Crystal HD Headers</a></td>
     </tr>
--- a/src/cryptopp-1-fixes.patch
+++ b/src/cryptopp-1-fixes.patch
@@ -1,0 +1,18 @@
+This file is part of MXE.
+See index.html for further information.
+
+Fix broken Makefile, add relevant object files and don't use -DCRYPTOPP_EXPORTS
+
+diff -ur a/GNUmakefile b/GNUmakefile
+--- a/GNUmakefile	2015-11-22 22:17:59.000000000 +0100
++++ b/GNUmakefile	2015-11-29 13:04:15.587252779 +0100
+@@ -255,8 +255,7 @@
+ LIBOBJS := $(filter-out $(TESTOBJS),$(OBJS))
+ 
+ # List cryptlib.cpp first in an attempt to tame C++ static initialization problems
+-DLLSRCS := cryptlib.cpp algebra.cpp algparam.cpp asn.cpp basecode.cpp cbcmac.cpp channels.cpp des.cpp dessp.cpp dh.cpp dll.cpp dsa.cpp ec2n.cpp eccrypto.cpp ecp.cpp eprecomp.cpp files.cpp filters.cpp fips140.cpp fipstest.cpp gf2n.cpp gfpcrypt.cpp hex.cpp hmac.cpp integer.cpp iterhash.cpp misc.cpp modes.cpp modexppc.cpp mqueue.cpp nbtheory.cpp oaep.cpp osrng.cpp pch.cpp pkcspad.cpp pubkey.cpp queue.cpp randpool.cpp rdtables.cpp rijndael.cpp rng.cpp rsa.cpp sha.cpp simple.cpp skipjack.cpp strciphr.cpp trdlocal.cpp
+-DLLOBJS := $(DLLSRCS:.cpp=.export.o)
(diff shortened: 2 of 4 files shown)


starius:
`$(PREFIX)` and `$(1)` can contain spaces, that is why they should be quoted with `'` in build rules.

index.html
```diff
@@ -1197,6 +1197,10 @@ <h2 id="packages">List of Packages</h2>
         <td class="website"><a href="http://www.freedesktop.org/wiki/Software/cppunit/">CppUnit</a></td>
     </tr>
     <tr>
+        <td class="package">cryptopp</td>
+        <td class="website"><a href="https://www.cryptopp.com/">Coin3D</a></td>
```
starius: s/Coin3D/Crypto++/

src/cryptopp.mk
```diff
@@ -0,0 +1,38 @@
...
+	    AR=$(TARGET)-ar \
+	    LD=$(TARGET)-ld \
+	    CXXFLAGS='-DNDEBUG -O3 -mtune=native -pipe' \
+	    $(if $(BUILD_STATIC),libcryptopp.a,cryptopp.dll)
+
+    mkdir -p $(PREFIX)/$(TARGET)/include/cryptopp
+    chmod 755 $(PREFIX)/$(TARGET)/include/cryptopp
+    cp $(1)*.h $(PREFIX)/$(TARGET)/include/cryptopp
+    chmod 644 $(PREFIX)/$(TARGET)/include/cryptopp/*
+    cp $(1)$(if $(BUILD_STATIC),libcryptopp.a,cryptopp.dll) $(PREFIX)/$(TARGET)/lib
```
TimothyGu: - `dll` goes to `bin/`
- `.dll.a` goes to `lib/`
- if there's no `.dll.a` make sure one is generated. See our policies here: https://github.com/mxe/mxe/wiki/Adding-Shared-Support-to-Packages#pre-commit-checklist
- you could consider using `$(INSTALL) -d` and `$(INSTALL)` rather than `mkdir -p`, `cp` and `chmod`.

starius:
Files generated by `make cryptopp` for all 4 targets:
https://gist.github.com/starius/ea45d62687e3add42121

src/cryptopp-1-fixes.patch
```diff
@@ -0,0 +1,17 @@
...
++DLLSRCS += 3way.o adhoc.o adler32.o algebra.o algparam.o arc4.o asn.o authenc.o base32.o base64.o basecode.o bench2.o bench.o bfinit.o blowfish.o blumshub.o camellia.o cast.o casts.o cbcmac.o ccm.o channels.o cmac.o cpu.o crc.o cryptlib.o datatest.o default.o des.o dessp.o dh2.o dh.o dll.o dlltest.o dsa.o eax.o ec2n.o eccrypto.o ecp.o elgamal.o emsa2.o eprecomp.o esign.o files.o filters.o fips140.o fipsalgt.o fipstest.o gcm.o gf2_32.o gf256.o gf2n.o gfpcrypt.o gost.o gzip.o hex.o hmac.o hrtimer.o ida.o idea.o integer.o iterhash.o luc.o mars.o marss.o md2.o md4.o md5.o misc.o modes.o mqueue.o mqv.o nbtheory.o network.o oaep.o osrng.o panama.o pkcspad.o polynomi.o pssr.o pubkey.o queue.o rabin.o randpool.o rc2.o rc5.o rc6.o rdrand.o rdtables.o regtest.o rijndael.o ripemd.o rng.o rsa.o rw.o safer.o salsa.o seal.o seed.o serpent.o sha3.o shacal2.o sha.o sharkbox.o shark.o skipjack.o socketft.o sosemanuk.o square.o squaretb.o strciphr.o tea.o test.o tftables.o tiger.o tigertab.o trdlocal.o ttmac.o twofish.o validat1.o validat2.o validat3.o vmac.o wait.o wake.o whrlpool.o winpipes.o xtrcrypt.o xtr.o zdeflate.o zinflate.o zlib.o
+ DLLOBJS := $(DLLSRCS:.cpp=.export.o)
+
```
tonytheodore: Replacing `*.cpp` with `*.o` will mean these won't use the `%.export.o` rule which is probably not what you want.

fiesh: Thank your for spotting this!  After fiddling around for quite a while, I realize that it just doesn't work if the %.export.o rule is used.  It seems -DCRYPTOPP_EXPORT, what is then defined, does some weird (MSVC I guess) magic that keeps things from working.  I even tested the resulting DLL file with the test binary under windows, it works perfectly fine without the export rule.  So this is actually what should be set, and I updated the patch to reflect this.

fiesh:
I incorporated hopefully all the very good feedback I got into these patches.  If all is well it could be rebased and added, I didn't want to rebase right away to help keep track of the changes.

@TimothyGu thank you for the link to https://github.com/mxe/mxe/wiki/Adding-Shared-Support-to-Packages.  I had no idea it existed, and only followed the documentation on http://mxe.cc/#creating-packages.  Maybe it could be linked there?

fiesh:
OK, there's still something amiss.  Our bigger project doesn't link against the lib.  I'll try to fix that.

tonytheodore:
I'm guessing it's related to the `%.export.o` rule, if that isn't used, the [name-mangling](https://github.com/weidai11/cryptopp/blob/master/config.h#L570) won't be consistent.

fiesh:
Hmm, I'm not sure.  I first built a static library by hand, and used it to test it on our internal project.  That worked fine, and then I created the MXE makefile.  Since the makefile only ever uses the `%.export.o` rule for building a DLL, it might be something else.

fiesh:
Oh, it was actually because our project uses the system wide MXE installation that has GCC 5.2.0, whereas my test installation has 4.9.3.  So all should be fine once I get the compiler versions right.

tonytheodore:
You can try the new (experimental) plugins if you're happy with later gcc versions:
```
make MXE_PLUGIN_DIRS=plugins/gcc52
```
This will essentially build gcc 5.2.0 as it was prior to #965.

fiesh:
OK, so I made the test a little bit bigger and included all necessary files in the dll.  Now everything works fine here.  So please review again!

starius:
cryptopp files, commit fc4ad4754ea31ca6aa62f6d43a06f7fe20012dea 
https://gist.github.com/starius/694e3e564b0082f5cfde
Looks good.

tonytheodore:
Run `make cleanup-style` to strip out the tabs, squash into a single commit, and it's good to go.

fiesh:
done!

tonytheodore:
Thanks!

P.S. this type of patching is difficult to maintain for those of us who are not familiar with the package, please keep it updated!

Issue #983, opened 2015-11-11
some patches are not git am'able


starius:
Example:

```
$ git am ../../src/libmysqlclient-1-fixes.patch
Patch format detection failed.

# remove header ...

$ git am ../../src/libmysqlclient-1-fixes.patch
Patch does not have a valid e-mail address.

$ sed 's/From: MXE/From: MXE@MXE/' -i ../../src/libmysqlclient-1-fixes.patch

$ git am ../../src/libmysqlclient-1-fixes.patch
OK
```
- [ ] make all patches `git am`able
- [ ] `git am && make-patch` should not change the original patch

TimothyGu:
It can be fixed through a simple regex as you can see. These files are created with the [patch-tool-mxe](https://github.com/mxe/mxe/blob/master/tools/patch-tool-mxe#L104).

starius:
Why not to store patches in a format accepted by `git am`?

I think we can start with an output of `git format-patch -p --stdout --no-numbered`.
Lines like `From f33262c29c253608f57e91af0e96bdb7bfa3cc00 Mon Sep 17 00:00:00 2001` should be removed or commit ID unified. `From:` should keep original author to know who can help with this patch. I think that replacing it with `From: MXE` is a bad idea as author information is lost and `git am` doesn't accept this string. `Signed-off-by:` can be removed. Final strings like `--` and `1.9.1` can be removed or unified. Finally MXE copyright header should be added in such a form that doesn't break `git am`.

starius:
How I patch package `foo`:

  * `make download-foo`
  * create an empty directory `foo`
  * unpack tar file from pkg
  * create new GIT repo in `foo/foo-1.2.3`
  * commit all files of `foo`
  * apply existing patches with `git am`
  * add / remove / modify commits to fix something [*]
  * create new patch file with `git format-patch` and copy it to `src/foo-1-fixes.patch`
  * `make foo`

I think all of them except [*] can be automated:

  * `make edit-patch-foo`
  * `cd patched-foo`
  * [*]
  * `make save-patch-foo`

TimothyGu:
I actually have an ancient branch a long time ago doing something like that. Then I found out about patch-tool-mxe which works pretty well.

starius:
`patch-tool-mxe` doesn't work with some patches.

```
$ ./tools/patch-tool-mxe init luabind
$ ./tools/patch-tool-mxe import luabind 2-elif
[...]

PR #1016, opened 2015-11-29
Patch tool improvements

starius:
See #983


tonytheodore:
Looks good to go if you're happy with it.

starius:
> I have a slight preference for numbered patches, if it's only a matter of size, I'd rather keep them

I have removed numbers because adding new patches changes total number of patches, introducing changes in all `[PATCH X/Y]` lines. It makes a diff harder to read. For example: https://github.com/mxe/mxe/commit/d8af97c4e9fe9e4305d6447a7b7d42d6d42e3f35#diff-e21cb80d9bc071d6f00be6b26da13647L9

tonytheodore:
Okay, thanks!

PR #1026, opened 2015-12-02
ENH: fix shared build for itk 4.8.2

rkanavath:
fix shared lib building on itk


tonytheodore:
Succeeds on all four targets - thanks!

Reworded commit https://github.com/mxe/mxe/commit/b317f60efb5215dae1c9c6419192822354af476f and enable shared build https://github.com/mxe/mxe/commit/6ac06b85784ebf0f3448ac589d76ddb003520afc

PR #1027, opened 2015-12-02
Add openthreads

rkanavath:
add openthreads using openscenegraph archive

### DIFF
--- a/index.html
+++ b/index.html
@@ -2051,6 +2051,10 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
         <td class="website"><a href="https://www.openssl.org/">openssl</a></td>
     </tr>
     <tr>
+      <td class="package">openthreads</td>
+        <td class="website"><a href="http://www.openscenegraph.org/">OpenThreads</a></td>
+    </tr>
+    <tr>
         <td class="package">opus</td>
         <td class="website"><a href="http://opus-codec.org/">opus</a></td>
     </tr>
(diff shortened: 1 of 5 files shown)


tonytheodore:
Good idea, it's currently building a lot of dependencies just for openthreads support. A couple of notes:

* Use variables for the common metadata, see [bfd](https://github.com/mxe/mxe/blob/master/src/bfd.mk) for an example.
* Needs corresponding disabling of openthreads in openscenegraph (unless they are installed in different locations)

rkanavath:
done.

tonytheodore:
Great! Squash the three openthreads commits into a single one and it's good to.

rkanavath:
happy ?

tonytheodore:
Now there are 6 commits?? Do you want me to do the rebasing?

rkanavath:
yes please. I always mess up with rebase.

tonytheodore:
Thanks! Merged https://github.com/mxe/mxe/compare/b24fe61...a854196

rkanavath:
for next time. 
can you tell how you did squashing commits?

tonytheodore:
In this case, `rebase` wouldn't even replay with no changes. I'm no `git` expert so I had to fall back to `git diff master` and `patch` to manually recommit the changes.

I think you did a `squash` in this commit https://github.com/bugbrains/mxe/commit/59047d11c0611270d1faac6537bfd34cb357375e, and then did some other "forcing" to make it merge.

In general though, `squash` doesn't do what I expect - I use `fixup` (even when I say "squash"). Try to merge and rebase frequently and always do a plain rebase before `rebase -i`.

PR #1017, opened 2015-11-29
install native and cross lua executables & add LuaRocks plugin

starius:

PR #1032, opened 2015-12-03
gcc: don't install *.dlls to version specific directory

tonytheodore:
See https://github.com/mxe/mxe/pull/1017#issuecomment-161243848

* any future side-by-side installs will use targets as a higher level
  directory separation, we don't want to mix libs built with different
  versions of the compiler.
* add note about keeping `--enable-version-specific-runtime-libs`
* remove TODO, there's no sane way to configure the install

```
$ find usr -name libgcc_s_*.dll
usr/i686-w64-mingw32.shared/bin/libgcc_s_sjlj-1.dll
usr/i686-w64-mingw32.shared.posix/bin/libgcc_s_sjlj-1.dll
usr/x86_64-w64-mingw32.shared/bin/libgcc_s_seh-1.dll
usr/x86_64-w64-mingw32.shared.posix/bin/libgcc_s_seh-1.dll
```

### DIFF
--- a/src/gcc.mk
+++ b/src/gcc.mk
@@ -52,11 +52,17 @@ define $(PKG)_CONFIGURE
 endef
 
 define $(PKG)_POST_BUILD
-    # TODO: find a way to configure the installation of these correctly
-    # ignore rm failure as parallel build may have cleaned up, but
-    # don't wildcard all libs so future additions will be detected
+    # - no non-trivial way to configure installation of *.dlls
+    #   each sudbir has it's own variations of variables like:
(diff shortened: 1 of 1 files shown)


starius:
Can https://github.com/mxe/mxe/tree/gcc-dll be removed?

tonytheodore:
Gone https://github.com/mxe/mxe/tree/gcc-dll

starius:
#1017 is working now.

tonytheodore:
Thanks, this has no build-time impact and is hopefully the last time we look at it (until 5.2 and `libcc1`).

PR #1035, opened 2015-12-03
add package "tor" to plugin "apps"

starius:
Tor is free software and an open network that helps you defend
against traffic analysis, a form of network surveillance that
threatens personal freedom and privacy, confidential business
activities and relationships, and state security.

https://www.torproject.org/

Previous attempt: #808


tonytheodore:
Thanks!

PR #1029, opened 2015-12-02
OSSIM: 1.8.18 -> 1.8.20-1

rkanavath:
merge this only after merging  https://github.com/mxe/mxe/pull/1027

PR #1036, opened 2015-12-04
ossim: fixes subdir and build

rkanavath:
follow up for https://github.com/mxe/mxe/pull/1029


tonytheodore:
Thanks for following up!

The changes to `ossim.mk` are cosmetic but the patch needs further thought. Take your time to get it into a state for upstream submission.

rkanavath:
the fix has been applied in the upstream already. but i pushed it after release. 

regarding patch, i think the ossimConfig.h and ossimVersion.h are added during the build. They shouldn't be there. I can make a new diff.

tonytheodore:
So, for the remaining patch to `ossim/include/ossim/point_cloud/ossimGenericPointCloudHandler.h`, can you provide a patch header along the lines of (https://github.com/mxe/mxe/blob/master/src/vtk-2-findfreetype.patch):

```
This file is part of MXE.
See index.html for further information.

This patch has been taken from:
http://www.vtk.org/Bug/view.php?id=12688
```

tonytheodore:
As an aside, we generally try to make packages as feature-complete as possible. If `ossim` has support for Point Cloud (package `pcl`), then that should be added as a dependency and have those features enabled.

I know you were probably trying to reduce the dependencies for `ossim` by breaking out `openthreads`, but still, it's better if we provide fully enabled packages and let users decide to minimise them according to their needs.

rkanavath:
AFAICT, ossim does not use pcl or any other for the  ossimGenericPointCloudHandler. usually ossim tries to not add any extra dependencies into their core.

tonytheodore:
Okay, so ignore PointCloud for now, ossim has some awareness of it's existence and does what it needs to do.

Back to the patch, I see you've added a link in the patch header:

https://github.com/rkanavath/ossim/commit/643f0b1f6984b97e7a72af7ff05085b7e716f934

Which has the commit message:
```
fix ossim-gsoc.cpp test compile on windows msvc 2012
```

We've all done such things and I'll reiterate my previous comments. Take your time to propose a solution to the ossim project and provide a link to that discussion.

rkanavath:
you need me to create a bug report, close it and provide a link?

the fix is already in their development branch.

sorry it was not clear.

starius:
I think, it worth providing the following link in the patch: https://github.com/ossimlabs/ossim/pull/3

tonytheodore:
Thanks! Rebased and merged https://github.com/mxe/mxe/compare/f2f1e99...a1b1fef

Issue #976, opened 2015-11-10
Create a package for GeoIP data


starius:
See https://github.com/mxe/mxe/issues/969#issuecomment-155248955

tonytheodore:
See #1050

PR #1040, opened 2015-12-06
add unpacker for .deb files

starius:
It is needed for #976, to unpack Debian package with GeoIP.dat downloaded from https://packages.debian.org/jessie/all/geoip-database/download

`tar xf` instead of `tar xzf` / `xz -dc '$(1)' | tar xf -` because Debian package can contain `data.tar.gz` or `data.tar.xz`. Probably this code can be rewritten better.


tonytheodore:
Looks okay to me, modern `tar` can detect what it's extracting - not worth rewriting unless there are issues. Thanks!

PR #1044, opened 2015-12-08
disable doxygen in MXE_CONFIGURE_OPTS

### LINKED ISSUE #952: extra documentation is installed on Jessie
I have compared lists of file names on Wheezy and on Jessie (with #925 applied) and found that the following files present on Jessie but not on Wheezy:
https://gist.github.com/starius/20c45b87c18e842fb1a5
All of them are documentation. Maybe documentation elimination doesn't work on Jessie.

tonytheodore:
fixes #952 and should provide a general solution.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,13 +68,18 @@ endef
 null  :=
 space := $(null) $(null)
 
+MXE_DISABLE_DOC_OPTS = \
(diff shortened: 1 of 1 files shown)


tonytheodore:
Harmless and effective - the general solution can come later.

PR #1045, opened 2015-12-08
add Tcl/Tk

fiesh:
For Tcl: I resorted to only building Tcl, not its additional packages that are again configure / make based and, I think, should be separate packages.  (Are there thoughts of introducing gentoo-like USE flags maybe?)

For Tk: Alas Tk requires Tcl to be present and compiled in order to compile.  So I built it again.  Not quite sure how to handle this better!

### DIFF
--- a/index.html
+++ b/index.html
@@ -2421,6 +2421,10 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
         <td class="website"><a href="http://developer.kde.org/~wheeler/taglib.html">TagLib</a></td>
     </tr>
     <tr>
+        <td class="package">tcl</td>
+        <td class="website"><a href="https://www.tcl.tk/">Tcl</a></td>
+    </tr>
+    <tr>
         <td class="package">tclap</td>
         <td class="website"><a href="http://tclap.sourceforge.net/">tclap</a></td>
     </tr>
@@ -2447,6 +2451,10 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
     <tr>
         <td class="package">tinyxml2</td>
         <td class="website"><a href="http://grinninglizard.com/tinyxml2/">tinyxml2</a></td>
+    </tr>
+    <tr>
+        <td class="package">tk</td>
+        <td class="website"><a href="https://www.tcl.tk/">Tk</a></td>
     </tr>
     <tr>
         <td class="package">tre</td>
--- a/src/tcl.mk
+++ b/src/tcl.mk
@@ -1,0 +1,28 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := tcl
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 8.6.4
+$(PKG)_CHECKSUM := 9e6ed94c981c1d0c5f5fefb8112d06c6bf4d050a7327e95e71d417c416519c8d
+$(PKG)_SUBDIR   := $(PKG)$($(PKG)_VERSION)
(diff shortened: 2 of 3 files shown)


tonytheodore:
You may have missed my [previous comment](https://github.com/mxe/mxe/pull/948#issuecomment-162736661), these are now merged as a plugin!

I was trying to convey the idea that you shouldn't go for reduced functionality or complex build rules in order to squeeze into the core MXE, instead focus on adding new packages from the Tcl/Tk ecosystem.

fiesh:
Ah shoot, and I completely missed that someone else already submitted a pull request for Tcl/Tk!  So I'll close this one.

fiesh:
Although, looking at it, it provides a rather clean install just by calling the install-binaries and install-libraries routines

fiesh:
Hmm, I guess I'm not quite sure how to understand the plugins system then.  Wouldn't it be correct to have Tcl/Tk in the core MXE and its additional packages lib sqlite in the plugins?

tonytheodore:
Crazy! In my mind, I was replying to @highperformancecoder in relation to #948. A comment like "I haven't found the necessary few hours to do this yet." and a new pull request a few hours later?

fiesh:
Haha, and I was completely oblivious of the parallel pull request.  So I don't know what's with mine, it compiles nicely and, as far as I can see, all files are installed in the correct position.  It omits the crazy packages of Tcl that I think shouldn't be handled in the MXE core just like so, and it does recompile Tcl when compiling Tk since that, alas, is necessary.

tonytheodore:
> Hmm, I guess I'm not quite sure how to understand the plugins system then.

It's just another directory that is optionally included, but those are `makefiles` with full awareness of every aspect of MXE. Plugins may not be the best name - those usually have defined interfaces. These are more like "injection attacks" but `make` is like that.

> Wouldn't it be correct to have Tcl/Tk in the core MXE and its additional packages lib sqlite in the plugins?

Theoretically, yes. Practically, `guile` has never been updated as it requires a newer native version of itself to build and `ocaml` hasn't been updated because it "does" build a native version of itself.

> It omits the crazy packages of Tcl that I think shouldn't be handled in the MXE core just like so, and it does recompile Tcl when compiling Tk since that, alas, is necessary.

With a plugin, you can focus less on trimming down with complex rules and spend your time on expanding with simpler rules.

fiesh:
So `plugins` is basically a less stable test bed?  And Tcl/Tk could go in there?  I'm still not quite following I think.

tonytheodore:
> So plugins is basically a less stable test bed?

Interesting question, plugins are new and were initially conceived as a place for experimental/optional/out-of-scope builds. They're still optional, but [`luarocks`](https://github.com/mxe/mxe/tree/master/plugins/luarocks) is more reproducible and performant than it's parent project - there's nothing about plugins that is more or less stable or supported.

`ocaml` and `guile` are "stable" in the sense that they haven't been updated and don't break. I consider this to be a more fragile position.

`tcl.tk` could follow a same path unless someone spends the time to keep them updated. I mention all these packages together as interpreters/compilers aren't like normal libraries when it comes to maintenance.

> And Tcl/Tk could go in there?

[Already there](https://github.com/mxe/mxe/tree/master/plugins/tcl.tk) in a form that isn't heavily modified or trimmed down - a more stable and complete base to build future work on until we get a better understanding of how it fits in with MXE.

tonytheodore:
Maybe a different way to say it is that plugins are optional so you'll have the freedom to cater to a self-selecting group of users.

Issue #782, opened 2015-07-30
Download failed or wrong checksum of package mingw-w64!


uwehermann:
Error:
```
Download failed or wrong checksum of package mingw-w64!
------------------------------------------------------------
 10200K .......... .......... .......... .......... .......... 97%  728K 1s
 10250K .......... .......... .......... .......... .......... 97%  299K 1s
 10300K .......... .......... .......... .......... .......... 98%  723K 1s
 10350K .......... .......... .......... .......... .......... 98%  293K 0s
 10400K .......... .......... .......... .......... .......... 99%  700K 0s
 10450K .......... .......... .......... .......... .......... 99%  310K 0s
 10500K .......... .......... .......... .......... ......... 100%  739K=30s
2015-07-30 16:46:46 (356 KB/s) - written to stdout [10803001/10803001]
------------------------------------------------------------
```

Log:
```
--2015-07-30 16:46:16--  http://downloads.sourceforge.net/project/mingw-w64/mingw-w64/mingw-w64-release/mingw-w64-v4.0.2.tar.bz2
Resolving downloads.sourceforge.net (downloads.sourceforge.net)... 216.34.181.59
Connecting to downloads.sourceforge.net (downloads.sourceforge.net)|216.34.181.59|:80... connected.
HTTP request sent, awaiting response... 302 Found
Location: http://netcologne.dl.sourceforge.net/project/mingw-w64/mingw-w64/mingw-w64-release/mingw-w64-v4.0.2.tar.bz2 [following]
--2015-07-30 16:46:16--  http://netcologne.dl.sourceforge.net/project/mingw-w64/mingw-w64/mingw-w64-release/mingw-w64-v4.0.2.tar.bz2
Resolving netcologne.dl.sourceforge.net (netcologne.dl.sourceforge.net)... 78.35.24.46, 2001:4dd0:1234:6::5f
Connecting to netcologne.dl.sourceforge.net (netcologne.dl.sourceforge.net)|78.35.24.46|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 10803001 (10M) [application/octet-stream]
Saving to: â€˜STDOUTâ€™

     0K .......... .......... .......... .......... ..........  0% 66.2K 2m39s
    50K .......... .......... .......... .......... ..........  0%  371K 93s
   100K .......... .......... .......... .......... ..........  1%  289K 74s
   150K .......... .......... .......... .......... ..........  1%  269K 65s
[...]

PR #1052, opened 2015-12-13
#782 fix, wrong checksum calculation command for OSX

CAMOBAP795:
Fix for issue which has been described in https://github.com/mxe/mxe/issues/782#issuecomment-164198871


tonytheodore:
Thanks!

PR #1054, opened 2015-12-13
patch-tool-mxe: do not replace object id with 000

starius:
Previously patch-tool-mxe produced the following diffs:

```diff
    diff --git a/CMakeLists.txt b/CMakeLists.txt
    index 0000000..0000000 100644
    --- a/CMakeLists.txt
    +++ b/CMakeLists.txt
```

patch tool refused to apply this patch:

```
    The next patch would create the file CMakeLists.txt,
    which already exists!  Assume -R? [n]
```

This commit replaces 0000000..0000000 with 1111111..2222222.


tonytheodore:
Thanks!

PR #1056, opened 2015-12-13
build-pkg: build empty dependencies of non-empty

starius:
Otherwise empty dependency produces no `.deb` file preventing non-empty package from being installed.

Example: non-empty package `gcc` depends on empty package `mingw-w64`.

### DIFF
--- a/tools/build-pkg.lua
+++ b/tools/build-pkg.lua
@@ -622,21 +622,40 @@ local function buildPackages(items, item2deps)
 end
 
 local function makeDebs(items, item2deps, item2ver, item2files)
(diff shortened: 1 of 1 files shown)


tonytheodore:
Thanks!

P.S. It may have been simpler to remove the empty deps from`make print-deps-for-build-pkg`.

starius:
> It may have been simpler to remove the empty deps from `make print-deps-for-build-pkg`

For chain A->B->C, where A and C are non-empty and B is empty, removing B from `make print-deps-for-build-pkg` would result in A not depending on C, which is an error.

PR #1059, opened 2015-12-14
Muparser shared library fix

rkanavath:
fix linking of muparser shared library

### DIFF
--- a/src/muparser-2-dllfixes.patch
+++ b/src/muparser-2-dllfixes.patch
@@ -1,0 +1,161 @@
+diff -burN muparser_v2_2_3.orig/include/muParserBase.h muparser_v2_2_3/include/muParserBase.h
+--- muparser_v2_2_3.orig/include/muParserBase.h	2015-04-27 14:34:08.489392134 +0200
++++ muparser_v2_2_3/include/muParserBase.h	2015-04-27 15:41:29.402019065 +0200
+@@ -40,7 +40,7 @@
+ #include "muParserBytecode.h"
+ #include "muParserError.h"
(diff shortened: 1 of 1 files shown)


starius:
1. Provide linking error being fixed (e.g., output of linker + linker command).
  2. Please squash 2 commits into 1 commit.
  3. Remove unrelated changes (e.g., [removing empty lines](https://github.com/mxe/mxe/pull/1059/files#diff-23175ff26f687472443d805f3b4f4a02R55)). Unrelated changes increase diff size and make it harder to read the diff.

tonytheodore:
Why not use the [existing `MUPARSERLIB_EXPORTS` define and `API_EXPORT(TYPE)` macro](https://github.com/beltoforion/muparser/blob/master/include/muParserDLL.h#L28-L36)?

rkanavath:
how can i use it with class. API_EXPORT is used for function in the current source. 

For classes I need another export variable or way to use API_EXPORT.

I can rename MUPARSER_DLL_EXPORT to MUPARSER_CLASS_EXPORT
if that is okay

rkanavath:
is this okay for merge?

tonytheodore:
> is this okay for merge?

Not quite yet, see above https://github.com/mxe/mxe/pull/1059#issuecomment-164491892. Can you also provide a small example of the code that causes the error?

tonytheodore:
A link to some upstream discussion about your approach would also be useful.

tonytheodore:
See https://github.com/beltoforion/muparser/issues/41

PR #1061, opened 2015-12-14
GDAL: install pkg-config and gdal data files

rkanavath:
gdal package config and gdal data are essentinal for other libraries using gdal. 

This PR fixes installing gdal data to prefix/target/share/gdal


starius:
It builds successfully. Creates `prefix/target/share/gdal` and `prefix/target/lib/pkgconfig/gdal.pc`

Can you add a simple test for compilation and linking using the gdal.pc file?

rkanavath:
okay for merge ?

starius:
Please, fix the following before merge:

  * `mkdir -p '$(PREFIX)/$(TARGET)/bin/`
  * rebase your commits

rkanavath:
added mkdir
rebased last two.

starius:
```
mkdir -p '$(PREFIX)/$(TARGET)/bin
```

Forgotten `'`.

Can you rebase and squash all your commits into one commit? It's hard to read 6 commits with merges and conflicts.

src/gdal.mk
```diff
@@ -87,4 +87,10 @@ define $(PKG)_BUILD
     $(MAKE) -C '$(1)/ogr'   -j '$(JOBS)' install OGR_ENABLED=
     $(MAKE) -C '$(1)/apps'  -j '$(JOBS)' install
     ln -sf '$(PREFIX)/$(TARGET)/bin/gdal-config' '$(PREFIX)/bin/$(TARGET)-gdal-config'
+
+    mkdir -p '$(PREFIX)/$(TARGET)/bin
```
TimothyGu: I don't think the `mkdir` is necessary in this case. If it was needed the symlink above wouldn't work :)

starius: `ln -s` doesn't fail if a file (its first argument) is absent.

tonytheodore: `mysql_config` and `xml2-config` are already present in `$(PREFIX)/$(TARGET)/bin`.

starius: Probabably `mkdir` is not needed here. I just want to be safe, since [many packages](https://github.com/mxe/mxe/issues/1111#issuecomment-169280181) are broken because of trying to copy `test-foo.exe` to missing `usr/<target>/bin/`.

tonytheodore: I see, `usr/<target>/bin/` is empty and `git` doesn't add it. We should just create the basic layout in `mxe-conf` and put dummy files in there.

rkanavath:
Should I keep mkdir ?

tonytheodore:
The `mkdir` can be removed now, it will be created earlier in the build with #1149.

tonytheodore:
I'm not sure what you're doing in your git repo locally, but these commits don't replay themselves. When people say `squash`, they almost always mean `fixup`. Rebase is a feature you should become comfortable with (I'm no expert by any means), but it saves you (or others) from doing this:

```
git checkout -b bugbrains-fix_gdal_data master
git pull https://github.com/bugbrains/mxe.git fix_gdal_data
git diff master > gdal.patch
git checkout master
patch -p1 < gdal.patch
rm gdal.patch
git checkout -B gdal-new
... remove `mkdir -p` line
git add .
git commit -m'gdal: install pkg-config and data files, add test using pkg-config' [--author='Rashad Kanavath <rashad.kanavath@c-s.fr>' ]
git checkout master
git merge gdal-new
```

`git diff` is the method-of-last-resort, your contributions would be much more quickly reviewed if you can remove anything unrelated to the proposed changes.

Your contributions are welcome, just take the time to learn how to make them easier for everyone.

Merged: https://github.com/mxe/mxe/compare/0f64a5a...31e5701

UniqueFool:
Note that GDAL is another library that is already supported by mxe which would benefit from supporting OpenCL [1]

[1]: https://github.com/mxe/mxe/issues/1161

Issue #682, opened 2015-05-11
OpenCV CMake modules


asarnow:
Right now there seems to be a minor bug with the CMake modules for OpenCV.  My apologies if I have missed something obvious.
- `mxe/usr/<target>/OpenCVConfig.cmake` and `mxe/usr/<target>/OpenCVConfig-version.cmake` seem to be wrong versions, which don't work.
- `mxe/usr/<target>/lib/OpenCV*.cmake` files work, but:
  - CMake module search path includes `<target>`, but not `<target>/lib`
  - OpenCV import prefix is set to parent directory of the `.cmake` file which is used.
  - OpenCV libraries are found in `<target>/lib`

My temporary work around is to copy the `.cmake` files from `<target>/lib` to `<target>`, and then manually modify the `CMakeCache.txt` so `OpenCV_DIR` and `OpenCV_CONFIG_PATH` point to `<target>/lib`.

I think the old module files should be removed, and the find package module should search in `<target>/lib` for `OpenCVConfig.cmake`, `OpenCVModules.cmake` and `OpenCVModules-release.cmake`, or the new files should be moved up one directory and "/lib" added after every `_IMPORT_PREFIX` reference.

tonytheodore:
Looks related to #598. There's a major version update to opencv waiting, lets see what happens after that.

PR #1062, opened 2015-12-14
opencv: fix opencvconfig.cmake

rkanavath:
patches in mxe does not fix the generated opencvconfig.cmake. 

simply sed 'ing installed opencvconfig fixes it

### DIFF
--- a/src/opencv.mk
+++ b/src/opencv.mk
@@ -58,6 +58,8 @@ define $(PKG)_BUILD
     $(SED) -i 's,share/OpenCV/3rdparty/,,g' '$(1).build/unix-install/opencv.pc'
     $(INSTALL) -m755 '$(1).build/unix-install/opencv.pc' '$(PREFIX)/$(TARGET)/lib/pkgconfig'
 
+   $(SED) -i 's,.{OpenCV_ARCH}/.{OpenCV_RUNTIME}/,,g' '$(PREFIX)/$(TARGET)/OpenCVConfig.cmake'
+
     '$(TARGET)-g++' \
         -W -Wall -Werror -ansi -pedantic \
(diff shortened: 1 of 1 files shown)


starius:
Indentation is broken.

starius:
Package `opencv` installs several cmake files:

  * usr/i686-w64-mingw32.static/OpenCVConfig.cmake
  * usr/i686-w64-mingw32.static/lib/OpenCVConfig.cmake
  * usr/x86_64-w64-mingw32.shared/OpenCVConfig.cmake
  * usr/x86_64-w64-mingw32.shared/lib/OpenCVConfig.cmake
  * usr/x86_64-w64-mingw32.static/OpenCVConfig.cmake
  * usr/x86_64-w64-mingw32.static/lib/OpenCVConfig.cmake
  * usr/i686-w64-mingw32.shared/OpenCVConfig.cmake
  * usr/i686-w64-mingw32.shared/lib/OpenCVConfig.cmake

And we also have `usr/<target>/OpenCVConfig-version.cmake`.

What is the difference between these 3 cmake files? Why are they installed not to `usr/<target>/share/cmake/`?

tonytheodore:
Related issue #682, it would be good to get to the bottom of this and avoid the use of `sed`.

rkanavath:
well, the bottom of it is the patch used in opencv by mxe. I see that it fixes some cmake code to generate a proper config.cmake. But this was not happening.

rkanavath:
There should be a single OpenCVConfig.cmake for both i386 and x86_64 archs.

tonytheodore:
These now go in `<target>\share\OpenCV` which is in the [standard search paths](https://cmake.org/cmake/help/v3.10/command/find_package.html#command:find_package).

PR #1064, opened 2015-12-14
also build and install FreeImagePlus

fiesh:
Install the C++ wrapper of freeimage as well

### DIFF
--- a/src/freeimage.mk
+++ b/src/freeimage.mk
@@ -25,12 +25,22 @@ define $(PKG)_BUILD
         FREEIMAGE_LIBRARY_TYPE=STATIC \
         TARGET=freeimage
 
+    $(MAKE) -C '$(1)' -j '$(JOBS)' -f Makefile.fip \
+        CXX='$(TARGET)-g++' \
+        CXXFLAGS='-DFREEIMAGE_LIB' \
+        CC='$(TARGET)-gcc' \
+        AR='$(TARGET)-ar' \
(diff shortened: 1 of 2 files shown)


starius:
Could you replace too similar calls to `$(MAKE)` with `for target in freeimage freeimageplus; do ...`?

fiesh:
Good idea, thanks, done and rebased.

starius:
It builds for me. :+1: 

This package installs `freeimage.pc` file for C.
Could you also install `freeimageplus.pc` for C++, please?

This package has a test of compilation and linking for C version.
Could you add the a test for C++ wrapper, please?

fiesh:
OK, put the test and the pkgconfig in -- and fixed the building, it was incorrect as the test revealed.  also rebased. Thanks for the feedback.

starius:
It builds successfully here.

src/freeimage.mk
```diff
@@ -23,14 +23,24 @@ define $(PKG)_BUILD
         AR='$(TARGET)-ar' \
         RC='$(TARGET)-windres' \
         FREEIMAGE_LIBRARY_TYPE=STATIC \
-        TARGET=freeimage
+        TARGET=freeimage; \
```
tonytheodore: Minor nitpick, we don't want to continue here.

fiesh: Thanks, true, I updated and rebased!

tonytheodore:
Thanks!

PR #1080, opened 2015-12-17
Make Travis CI build build-matrix.html

TimothyGu:
Untested, but probably works. Also we should consider dynamically generate HTML through JSON or something like that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 /settings.mk
-/build-matrix.html
 /usr*/
 /gits/
 /log*/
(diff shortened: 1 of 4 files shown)


starius:
It commits changes with `git commit -a -m 'Update versions.json' || true`.
We should update the commit message. Maybe "Update generated files"?

TimothyGu:
Fixed.

starius:
+1 for this.
Rebase & Merge.

tonytheodore:
> Also we should consider dynamically generate HTML through JSON or something like that.

Agreed, we also should move the package list back to `src/*.mk` and generate that also.

Thanks! [Rebased and merged](https://github.com/mxe/mxe/compare/cd70bd2...abda011).

PR #1081, opened 2015-12-17
Update ffmpeg and its dependencies

TimothyGu:
@starius, I'd love it if you could test these packages on your machine as it seems to be a LOT faster than mine :)

PR #1082, opened 2015-12-17
Remove gcc patch

TimothyGu:
The first one seems to be needed for MinGW.org only. The patch was originally provided to fix _clear87 not defined errors. A simple test program from [_clear87 documentation](https://msdn.microsoft.com/en-us/library/49bs2z07.aspx) seems to be working.

The second one is not needed since [this commit](http://sourceforge.net/p/mingw-w64/code/6303/) (committed on 2013-09-20).


starius:
Is #1082 a part of #1081 ?

TimothyGu:
> Is #1082 a part of #1081 ?

Oops… seems like it is, but it was not intended to be. I'll remove this commit from #1081 since it is of a different subject.

tonytheodore:
Does removing the first patch also mean the [mingw-w64 patch](https://github.com/mxe/mxe/blob/master/src/mingw-w64-1.patch) is no longer necessary? That one seems related to the `-isystem` cflag messing up include order.

tonytheodore:
The first does seem to be a relic of the old mingw.org approach of custom spec files and include orders. 

The second is indeed unnecessary [after investigating intrinsics for qtwebkit](https://github.com/mxe/mxe/commit/9a6c17c0e363a31a8fd483b10970225b5f2fe2cb).

I've added this to #1659.

PR #1083, opened 2015-12-17
Use my fork of libgnurx

TimothyGu:
The original libgnurx is based off glibc 2.5, which has multiple known vulnerabilities, including but not limited to:

- CVE-2013-0242: Buffer overflow in the extend_buffers function in the
  regular expression matcher (posix/regexec.c) in glibc, possibly 2.17
  and earlier, allows context-dependent attackers to cause a denial of
  service (memory corruption and crash) via crafted multibyte
  characters.

### DIFF
--- a/src/libgnurx.mk
+++ b/src/libgnurx.mk
@@ -3,20 +3,18 @@
 
 PKG             := libgnurx
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 2.5.1
-$(PKG)_CHECKSUM := 7147b7f806ec3d007843b38e19f42a5b7c65894a57ffc297a76b0dcd5f675d76
+$(PKG)_VERSION  := 2.6.1
+$(PKG)_CHECKSUM := ee6edc110c6b63d0469f4f05ef187564b310cc8a88b6566310a4aebd48b612c7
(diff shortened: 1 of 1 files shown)


starius:
Can we add part of [your changes](https://github.com/TimothyGu/libgnurx/compare/875b17197a04177bdd7263628825f0c5b1924d1b...master) as a patch? E.g., changes in COPYING.LIB or adding/removing whitespaces can be omitted.

What is the source of https://github.com/TimothyGu/libgnurx/commit/66cc2c6c50c8aded6873d14b3fbf13df5e306c35 ? It changes COPYING.LIB and source files.

TimothyGu:
In case you didn't know, libgnurx is "the regex functionality from glibc 2.22 extracted into a separate library, for Win32."

> Can we add part of your changes as a patch? E.g., changes in COPYING.LIB or adding/removing whitespaces can be omitted.

It's difficult, and that's why I didn't do so in the first place. It's much easier to adapt the new glibc to a stand-alone version for Windows than vice versa.

> What is the source of TimothyGu/libgnurx@66cc2c6 ? It changes COPYING.LIB and source files.

It imports the source of glibc 2.22, as it is written in the commit message. The changes to the files all came from the glibc 2.22 tarball.

PR #1089, opened 2015-12-20
libass: update to 0.13.1

avih:


tonytheodore:
Thanks, we were just looking into this.

See https://github.com/mxe/mxe/pull/1081#issuecomment-165362698 and https://github.com/libass/libass/issues/209, there are issues with the shared build in this release (that also breaks ffmpeg).

PR #1094, opened 2015-12-20
qrencode

ckvsoft:
add QREncode to mxe


starius:
Please add some more information to commit messages instead of "Create ...". It should explain why this patch was added and what problem does it solve.

starius:
In https://github.com/ckvsoft/mxe/commit/a11bdf4e218da55624ad53fc2fc80d74f3e45d09 (Create qrencode-4-tests.patch):

  * why did you add `TESTS` variable? It is not used in other patches.
  * why did you swap `../libqrencode.la` and `$(DECODER_LIBS)`. `libqrencode` seems to depend on `libdecoder`, so it should be listed before `libdecoder`, as in the upstream.

ckvsoft:
QREncode creates QR-Code from Text
It was used in a QT-Program (QRK) wich is in development. QRK is an CashBox Application.
I'm using Linux but this App also should run under Windows. So i need this with mxe and hope other users are happy too.

starius:
In https://github.com/ckvsoft/mxe/commit/0ed2c3a92b14706e04e51e93a75d308d19d76dc0 (Create qrencode-2-bzero.patch ): why `#if defined(__MINGW32__)` is needed? Isn't `#  if !defined(bzero)` enough?

starius:
Please, replace URL `http://fukuchi.org/works/qrencode/qrencode-3.4.4.tar.bz2` with `http://fukuchi.org/works/$(PKG)/$($(PKG)_FILE)`. The version should not be hard coded in other variables.

ckvsoft:
the __MINGW32__ patch is not mine,sorry i have not tested without. Is this a problem?

starius:
If a patch is not needed, it should not be added. We have to maintain all the patches. If a package has many patches, it is hard to update.

ckvsoft:
OK, i'll clean up

ckvsoft:
cleanup useless patches

starius:
Failed to build: https://gist.github.com/222ce27f760a17d5e8b7
It seems to depend on package `png`.

starius:
Please, add `pkgconfig` file generation for package `qrencode` and add a test of compilation and linking. The test should use `pkgconfig` to get compilation and linking flags. You can use other packages as an example (e.g., [lua](https://github.com/mxe/mxe/blob/c26f3225fba8ffe9dd54db759d4680a1314d350f/src/lua.mk#L23)).

ckvsoft:
Sorry

starius:
TODO:
  * tests, see https://github.com/mxe/mxe/pull/1094#issuecomment-166211138
  * squash all commits into one commit using `git rebase`

tonytheodore:
Replaced with #1890

Issue #1085, opened 2015-12-18
Building MXE in directory with special characters


starius:
Following on from #1030, MXE doesn't build in a directory with spaces in name:
```
Makefile:228: *** GNU Make chokes on paths with spaces. Stop.
```

It fails with another error if a directory contains `$`:
```
~/mxe-with-$pecial-ch@rs?-and-кириллица-!%^&()$ make -j 6
make: *** No rule to make target `/home/mxe/mxe-with-ecial-ch@rs?-and-кириллица-!%^&()/usr/i686-w64-mingw32.static/installed/a52dec', needed by `a52dec'.  Stop.
```

Can we fix it? I thought we add `'` around paths (e.g. `'$(PREFIX)/$(TARGET)/lib/pkgconfig/$(PKG).pc'`) to prevent errors like this. If we can't fix it, then error messages should be same for space and for `$`.

tonytheodore:
Interesting. Most quoting we do is for shell commands, but we don't quote the file targets. I think we could either try quoting those targets or switching to pattern rules (the latter is something I've been meaning to try anyway).

starius:
Commit 8aa5ccc123b429c59b453b7be2fce9ef3bea678a. Directory `/home/mxe/mxe-libmysqlclient-++-plus-plus`.

The following packages fail to build:

  * libmysqlclient (fixed in 2e71b21211db4cea44d5518cc2afda2e699361a5)
  * vtk (fixed in #1090)
  * opencv (fixed in #1093)
  * exiv2 (unrelated, fixed in 6e50094684b9f8058f9fd94af8457e4ced306b57)

Other packages were built successfully.

tonytheodore:
Seems to be resolved.

Issue #1015, opened 2015-11-27
Ensure plugins work with settings.mk


tonytheodore:
Along with prefix and log from #967 and possibly pkg in case people want to do fully out-of-source builds.

tonytheodore:
Also, the dollar/space checks (#1097) should cover all these manually set paths.

starius:
Out-of-source builds are of interest for building something on top of Debian packages with MXE binary distribution. Example 1: https://github.com/mxe/mxe/issues/962#issuecomment-160294290 Example 2: install `qt` to /usr/lib/mxe and build some Qt based tool in a home dir. Is it a possible scenario? The problem is that such a build has to find headers and libs in /usr/lib/mxe, but install build results to some other directory.

tonytheodore:
Also, general docs (see https://github.com/mxe/mxe/pull/1138#issuecomment-169303377).

tonytheodore:
@starius both examples above should be possible, though I'm not sure how the second would work if installing a lib in home directory then a tool that requires it.

starius:
I also don't know how the second should work. Maybe it is a bad idea to mix two MXE directories.

PR #1097, opened 2015-12-22
Makefile: error on paths with dollar signs

tonytheodore:
see #1085. 

Dollars signs are okay in some places, but they don't play well with `eval` or variable targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,6 +230,11 @@ ifneq ($(words $(PWD)),1)
     $(error GNU Make chokes on paths with spaces)
 endif
 
+# dollar signs also cause troubles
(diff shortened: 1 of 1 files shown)


starius:
+1

Testing:
```
mxe-$dollar$ make gcc
Makefile:235: *** GNU Make chokes on paths with dollar signs.  Stop.

mxe-no-dollar$ make gcc
[create settings.mk]
[using autodetected 6 job(s)]
[check requirements]
[download]  gmp
...
```

tonytheodore:
Thanks, I just realised both checks should be aware of `MXE_PLUGIN_DIRS` and any manually paths in `settings.mk` (#1015).

PR #1098, opened 2015-12-22
fix freeimage build

fiesh:
Alas, I broke the freeimage build.  I was gone for the weekend and therefore only noticed now.  Sorry for this.

### DIFF
--- a/src/freeimage-2-fixes.patch
+++ b/src/freeimage-2-fixes.patch
@@ -1,0 +1,18 @@
+This file is part of MXE.
+See index.html for further information.
+
+Add the FREEIMAGE_LIB define in order to keep the build process from defining
+dllspec specifications.
+
+diff -ur a/Makefile.fip b/Makefile.fip
+--- a/Makefile.fip	2010-12-09 15:12:12.000000000 +0100
++++ b/Makefile.fip	2015-12-22 13:24:33.071944345 +0100
+@@ -17,7 +17,7 @@
+ MODULES := $(MODULES:.cpp=.o)
+ CFLAGS ?= -O3 -fPIC -fexceptions -fvisibility=hidden -DNO_LCMS
+ CFLAGS += $(INCLUDE)
+-CXXFLAGS ?= -O3 -fPIC -fexceptions -fvisibility=hidden -Wno-ctor-dtor-privacy
++CXXFLAGS ?= -O3 -fexceptions -fvisibility=hidden -Wno-ctor-dtor-privacy -DFREEIMAGE_LIB
+ CXXFLAGS += $(INCLUDE)
+ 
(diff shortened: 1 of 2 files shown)


starius:
1. Please explain your changes. What is the error? Why `-DFREEIMAGE_LIB` was moved from freeimage.mk to Makefile.fip? Why `-fPIC -fexceptions -fvisibility=hidden -Wno-ctor-dtor-privacy` were removed?
2. Please add a patch, not a raw diff. See previous patches from `src/freeimage-1-fixes.patch`. You can use `tools/patch-tool-mxe` to make such patches automatically. Add the explanation (1) as a commit message of the patch.
```
 From ... Mon Sep 17 00:00:00 2001
 From: ...
 Date: ...
 Subject: [PATCH] ...
```

fiesh:
1. The error is that it does not buid.  The reason is that `CXXFLAGS="-DFREEIMAGE_LIB"` blocks all other `CXXFLAGS` in the makefile, resulting in many missing includes.  One could replicate these includes all in the `CXXFLAGS` in `freeimage.mk`, but that wouldn't seem like a good idea.  `-FPIC` is pointless for win targets.  `-fexeptions` is the default.  `-fvisibility=hidden` doesn't make sense with windows targets either, as far as I understand.  `-Wno-ctor-dtor-privacy` is a warning and therefore doesn't affect the resulting binary.

2. I don't understand the difference between a patch and a raw diff, sorry.

starius:
Thanks!

1. Seems to be fixed in https://github.com/mxe/mxe/commit/4e33d730957c26fb6444ebd09bbcd5e51f07631f
2. A patch is a text dump of a commit, i.e. diff + **metadata**. It consists of an author (you), a date and a commit message which explains why this patch is needed. This information would be very useful in the future, if something fails or the library is updated.

Please find in my first comment how a patch looks like (From, Date, Subject). Currently your diff looks as if it is a part of the previous patch (`[PATCH 2/2] mingw-w64 fixes`).

fiesh:
I made the patch a separate file.  I didn't really know what to write besides the explanation for the change, I hope that's sufficient?

fiesh:
I also rebased.

starius:
It's fine, but don't remove `-fPIC`, because it is unrelated to the patch.

fiesh:
-fPIC only causes warnings since it's mandatory for the Windows platform anyway, so there is no point in having it.

starius:
Add this information to the comment, please:
```
+Add the FREEIMAGE_LIB define in order to keep the build process from defining
+dllspec specifications. Remove -fPIC to fix a warning.
```

fiesh:
Thank you, good idea, fixed!

starius:
Thanks!

starius:
See also https://github.com/mxe/mxe/pull/1100
That's what I meant by converting a diff into a patch.

Issue #1104, opened 2015-12-27
small fix after boost update


pavelvat:
add to boost-1-fixes.patch fix for "std::sprintf" (remove the prefix "std::")
https://svn.boost.org/trac/boost/ticket/11872

starius:
This error is caused by `usr/i686-w64-mingw32.shared/include/libintl.h` (installed by `gettext`), which contains the following code:

```
    359 #if !(defined sprintf && defined _GL_STDIO_H) /* don't override gnulib */
    360 #undef sprintf
    361 #define sprintf libintl_sprintf
    362 extern int sprintf (char *, const char *, ...);
    363 #endif
    364 #if !(defined vsprintf && defined _GL_STDIO_H) /* don't override gnulib */
    365 #undef vsprintf
    366 #define vsprintf libintl_vsprintf
    367 extern int vsprintf (char *, const char *, va_list);
    368 #endif
```

So `sprintf` is replaced with `libintl_sprintf` and `std::sprintf` is replaced with `std::libintl_sprintf`. This can result in error in any package using `std::sprintf`. I think, we should patch `gettext`, not `boost`.

pavelvat:
Currently boost-1-fixes.patch contain my patch: https://github.com/boostorg/interprocess/pull/23 
This patch is not necessary, because BOOST_USE_WINDOWS_H not defined:
```
#ifdef BOOST_USE_WINDOWS_H
#include <windows.h>

#  if defined(BOOST_INTERPROCESS_BOOTSTAMP_IS_LASTBOOTUPTIME)
#  include <Wbemidl.h>
#  include <Objbase.h>
#  endif

#include <Shlobj.h>
#endif
```
so case sensitivity for headers is irrelevant. Previously pull https://github.com/boostorg/interprocess/pull/15 is obsolete for boost-1.60.

starius:
@pavelvat, please create a pull request to remove patch "Fixed case sensetive for linux mingw" from src/boost-1-fixes.patch.

pavelvat:
@starius, I do not think it should be done (next version of boost will include this patch). In any case, you can remove the patch without my pull request. This patch was needed to boost-1.57, and may be will be needed in future versions of boost.

starius:
usr/i686-w64-mingw32.shared/include/libintl.h:

```
 84    The redirection happens by default through macros in C, so that &gettext
 85    is independent of the compilation unit, but through inline functions in
[...]

PR #1107, opened 2015-12-27
boost: remove unneeded patch "Fixed case sensetive for linux mingw"

starius:
See https://github.com/mxe/mxe/issues/1104#issuecomment-167437968


tonytheodore:
Thanks!

PR #1116, opened 2015-12-31
add package libechonest

pavelvat:


src/libechonest.mk
```diff
@@ -0,0 +1,28 @@
...
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 2.3.1
+$(PKG)_CHECKSUM := ab961ab952df30c5234b548031594d7e281e7c9f2a9d1ce91fe5421ddde85e7c
+$(PKG)_SUBDIR   := $(PKG)-$($(PKG)_VERSION)
+$(PKG)_FILE     := $(PKG)-$($(PKG)_VERSION).tar.gz
+$(PKG)_URL      := https://github.com/lfranchi/libechonest/archive/$($(PKG)_VERSION).tar.gz
+$(PKG)_DEPS     := gcc qt qjson
+
+define $(PKG)_UPDATE
+    $(WGET) -q -O- 'https://github.com/lfranchi/libechonest/archive/' | \
```
starius: Use MXE_GET_GITHUB_TAGS

src/libechonest.mk
```diff
@@ -0,0 +1,28 @@
...
+    $(SED) -n 's,.*/\([0-9][^"]*\)/"\.tar.*,\1,p' | \
+    sort | uniq | \
+    head -1
+endef
+
+define $(PKG)_BUILD
+    mkdir '$(1)/build'
+    cd '$(1)/build' && cmake .. \
```
starius: Use prefixed cmake wrapper (`'$(TARGET)-cmake'`).

src/libechonest.mk
```diff
@@ -0,0 +1,28 @@
...
+        -DCMAKE_TOOLCHAIN_FILE='$(CMAKE_TOOLCHAIN_FILE)'
+
+    $(MAKE) -C '$(1)/build' -j '$(JOBS)' install
```
starius: Split this command into building (`-j '$(JOBS)'`) and `install` (`-j 1`).

src/libechonest.mk
```diff
@@ -0,0 +1,28 @@
...
+endef
+
+$(PKG)_BUILD_STATIC =
```
starius: What is wrong with static build?

src/libechonest.mk
pavelvat: I don't like '$(TARGET)-cmake' because it is required adding MXE-directory in environment variable $PATH.
```-DCMAKE_TOOLCHAIN_FILE='$(CMAKE_TOOLCHAIN_FILE)'``` - it is working without editing $PATH.

**Comment by starius** [comment]
The build succeeded. New files: https://gist.github.com/d191deb6954157d1d1d9

Please install `.pc` file to `<TARGET>/lib/pkgconfig/` and provide a sample source file and build rule for it (using the `.pc` file) to test compilation and linking.

File: src/libechonest.mk (hunk 3 above)
Comment (pavelvat) [inline src/libechonest.mk]: Why?

File: src/libechonest.mk (hunk 4 above)
Comment (pavelvat) [inline src/libechonest.mk]: Seems it is not supported static build.

File: src/libechonest.mk (hunk 2 above)
Comment (tonytheodore) [inline src/libechonest.mk]: > I don't like '$(TARGET)-cmake' because it is required adding MXE-directory in environment variable $PATH.

The `$PATH` is set by the main `Makefile` so building MXE doesn't require setting any environment variables. Using it does require setting the `$PATH` (at least) so you probably already have it set.

>  `-DCMAKE_TOOLCHAIN_FILE='$(CMAKE_TOOLCHAIN_FILE)'` - it is working without editing $PATH.

Do you mean that usage works in a normal shell session without setting the `$PATH`? I don't see how it could and would be interested to see how you do this.

Comment (pavelvat) [inline src/libechonest.mk]: >Do you mean that usage works in a normal shell session without setting the $PATH?

Yes. I am building EiskaltDC++ without editing $PATH in ~/.bashrc
For example: 
https://gist.github.com/pavelvat/1af9cf37d529bf41d87e
https://gist.github.com/pavelvat/009e52249a41133ad720

Comment (starius) [inline src/libechonest.mk]: > Do you mean that usage works in a normal shell session without setting the $PATH? I don't see how it could and would be interested to see how you do this.

I build NPG-explorer with MXE without changing `PATH`: https://github.com/npge/npge/blob/master/windows/build.sh

cmake seems to invoke build tools with whole paths.

File: src/libechonest.mk (hunk 3 above)
Comment (starius) [inline src/libechonest.mk]: There are packages for which `install` target doesn't depend on `build` target (see https://github.com/mxe/mxe/pull/1017#discussion_r46104558). There are also [packages](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=42980) for which parallel `make install` can fail. That is why separating these steps is a good practice.

File: src/libechonest.mk (hunk 2 above)
Comment (starius) [inline src/libechonest.mk]: If we change cmake wrapper in future (e.g., add some options to `cmake`), then code using plain cmake would need patching.

Comment (tonytheodore) [inline src/libechonest.mk]: Thanks @pavelvat and @starius, I hadn't considered that usage before. Some absolute paths are set by the toolchain file but a notable omission is `*-ar` and I'm not sure what impact `set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)` has on this.

File: src/libechonest.mk (hunk 4 above)
Comment (starius) [inline src/libechonest.mk]: I have tried to build it statically: https://gist.github.com/2a03a7d2db94a94ebf64
Build errors seem to be caused by missing libraries to link to.

File: src/libechonest.mk (hunk 2 above)
Comment (tonytheodore) [inline src/libechonest.mk]: > If we change cmake wrapper in future (e.g., add some options to cmake), then code using plain cmake would need patching.

Indeed, any code calling `cmake` directly will eventually become outdated. One of the main points of the wrapper is to use our own version and projects should start using it now as a means of future-proofing.

Comment (starius) [inline src/libechonest.mk]: @pavelvat, cmake wrapper does use our version of cmake. See file `usr/bin/i686-w64-mingw32.static-cmake`, it invokes file `/path/to/mxe/usr/x86_64-unknown-linux-gnu/bin/cmake`.

File: src/libechonest.mk (hunk 4 above)
Comment (pavelvat) [inline src/libechonest.mk]: https://github.com/lfranchi/libechonest/blob/master/src/CMakeLists.txt#L47

Comment (starius) [inline src/libechonest.mk]: It compiles statically in Linux machine after replacing `SHARED` with `STATIC`. After applying [the patch](https://gist.github.com/f0d4a8e53640f0b04c11) MXE shared target builds as before and static target [fails](https://gist.github.com/20b64c4141921b8f25dc) with dllimport linking errors. Can you fix them, please? Maybe we should fix this in upstream also.

@tonytheodore, can you see [CMake warning](https://gist.github.com/starius/20b64c4141921b8f25dc#file-libechonest_i686-w64-mingw32-static-L46) from the build. CMake modules of MXE seem to be installed to multiple directories and conflict with themselves.

Comment (tonytheodore) [inline src/libechonest.mk]: added note about CMake warnings to #971. I think we still want the old behaviour and warnings until we convert the modules to patches.

File: src/libechonest.mk (hunk 2 above)
Comment (pavelvat) [inline src/libechonest.mk]: '$(1)/build' is correct?
libechonest.mk:
```
    mkdir '$(1)/build'
    cd '$(1)/build' && cmake .. \
```

For example, 
mman-win32.mk:
```
    mkdir '$(1).build'
    cd    '$(1).build' && '$(TARGET)-cmake' '$(1)'\
```
muparserx.mk:
```
    cd '$(1)' && '$(TARGET)-cmake' \
```
ossim.mk:
```
    mkdir '$(1).build'
    cd '$(1).build' && '$(TARGET)-cmake' \
```
boost.mk:
```
    cd '$(1).test-cmake' && '$(TARGET)-cmake' \
```

'$(1)' after '$(TARGET)-cmake' for mman-win32.mk?

Comment (starius) [inline src/libechonest.mk]: `'$(1).build'` is preferred over `'$(1)/build'` not to write build results into sources directory.

**Comment by pavelvat** [comment]
>You can compile libechonest's tests instead.

By default:
```option(ECHONEST_BUILD_TESTS "Build all unit tests" ON)```

**Comment by starius** [comment]
I see. Please install file test_libechonest.exe to target/bin directory.

**Comment by pavelvat** [comment]
This option: option(ECHONEST_BUILD_TESTS "Build all unit tests" ON)
builds this testing *.exe files:
ArtistTest.exe CatalogTest.exe GenreTest.exe PlaylistTest.exe SongTest.exe TrackTest.exe
What are you want? Copying one from these files to "target/bin" and renaming it to test_libechonest.exe?

**Comment by starius** [comment]
I think, one file is enough. Choose any of them. Main idea of this copying is to break a build which for some reason doesn't create that file.

**Comment by TimothyGu** [comment]
Reading libechonest's CMakeLists, it seems that it could depend either on Qt 4 or Qt 5, with priority to Qt 5 (even though support for Qt 5 is "still experimental and not officially supported"). You'll have to find a way to prevent it from linking to Qt 5 if the user has both versions installed.

**Comment by pavelvat** [comment]
libechonest.mk:
>$(PKG)_DEPS     := gcc qt qjson

libechonest-2.3.1/CMakeLists.txt(12):
>option(BUILD_WITH_QT4 "Build libechonest with Qt4 no matter if Qt5 was found" ON)

libechonest-2.3.1/CMakeLists.txt(14):
```
if( NOT BUILD_WITH_QT4 )
    find_package(Qt5Core QUIET)
```

So, in order to enable build with Qt5 is required add to libechonest.mk:
```
    cd '$(1).build' && '$(TARGET)-cmake' '$(1)' \
        -DBUILD_WITH_QT4=OFF
```

**Comment by starius** [comment]
Looks good. The last thing to do is to install .pc file to `<TARGET>/lib/pkgconfig/` to simplify the use.

**Comment by pavelvat** [comment]
@starius 
This:
```
    # create pkg-config file
    $(INSTALL) -d '$(PREFIX)/$(TARGET)/lib/pkgconfig'
    (echo 'Name: $(PKG)'; \
     echo 'Version: $($(PKG)_VERSION)'; \
     echo 'Description: $(PKG)'; \
     echo 'Cflags: -I''$(PREFIX)/$(TARGET)/qt/include/'' -I''$(PREFIX)/$(TARGET)/qt/include/QtCore'' -I''$(PREFIX)/$(TARGET)/qt/include/QtNetwork'' '; \
     echo 'Libs: -lechonest';) \
     > '$(PREFIX)/$(TARGET)/lib/pkgconfig/$(PKG).pc'

    '$(TARGET)-g++' \
        -W -Wall -Werror -ansi -pedantic \
        '$(2).cpp' -o '$(PREFIX)/$(TARGET)/bin/test-$(PKG).exe' \
        `'$(TARGET)-pkg-config' libechonest --cflags --libs`
```
successful build trivial libechonest-test.cpp file:
```
#include <echonest/Song.h>

int main()
{
    Echonest::Song a;
    return 0;
}
```
How to configure Cflags for the general case (not only QtCore and QtNetwork)?

**Comment by starius** [comment]
Why do you need `-I''$(PREFIX)/$(TARGET)/qt/include/QtCore'' -I''$(PREFIX)/$(TARGET)/qt/include/QtNetwork'' `, not only `-I''$(PREFIX)/$(TARGET)/qt/include/''`?

`Qt*.pc` files use the following Cflags:
```
Cflags: -DQT_THREAD_SUPPORT -I/usr/lib/mxe/usr/i686-w64-mingw32.static/qt/include -I${includedir}
```

**Comment by pavelvat** [comment]
Because without paths to headers for QtCore and QtNetwork occur compile error for file libechonest-test.cpp. 
log:
```
# create pkg-config file
install -d '/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/lib/pkgconfig'
(echo 'Name: libechonest'; echo 'Version: 2.3.1'; echo 'Description: libechonest'; echo 'Cflags: -I''/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/qt/include/'' '; echo 'Libs: -lechonest';) > '/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/lib/pkgconfig/libechonest.pc'
'i686-w64-mingw32.shared-g++' -W -Wall -Werror -ansi -pedantic './src/libechonest-test.cpp' -o '/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/bin/test-libechonest.exe' `'i686-w64-mingw32.shared-pkg-config' libechonest --cflags --libs`
In file included from /home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/include/echonest/Track.h:21:0,
                 from /home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/include/echonest/Song.h:22,
                 from ./src/libechonest-test.cpp:6:
/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/include/echonest/AudioSummary.h:23:23: fatal error: QSharedData: No such file or directory
 #include <QSharedData>
                       ^
compilation terminated.

```

QtCore.pc:
```
prefix=/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/qt
exec_prefix=${prefix}
libdir=${prefix}/lib
includedir=${prefix}/include/QtCore
qt_config=lex yacc warn_on uic resources incremental_off windows release ReleaseBuild Release build_pass qt warn_on release link_prl copy_dir_files debug_and_release debug_and_release_target precompile_header cross_compile shared def_files_disabled exceptions rtti no_mocdepend release stl qt_no_framework cross_compile system-sqlite release stl precompile_header mmx 3dnow sse sse2 sse3 ssse3 sse4_1 sse4_2 avx dylib create_prl link_prl depend_includepath QTDIR_build release ReleaseBuild Release build_pass qt_install_headers qt warn_on depend_includepath qmake_cache target_qt debug_and_release create_pc release ReleaseBuild Release build_pass no_autoqmake moc thread dll shared
moc_location=${prefix}/bin/moc
uic_location=${prefix}/bin/uic
rcc_location=${prefix}/bin/rcc
lupdate_location=${prefix}/bin/lupdate
lrelease_location=${prefix}/bin/lrelease

Name: Qtcore
Description: Qtcore Library
Version: 4.8.7
Libs: -L${libdir} -lQtCore4
Libs.private: -lole32 -luuid -lws2_32 -ladvapi32 -lshell32 -luser32 -lkernel32 .obj/release-shared/QtCore_resource_res.o -lz
Cflags: -DQT_SHARED -DQT_THREAD_SUPPORT -I/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/qt/include -I${includedir}
```

libechonest.pc:
```
Name: libechonest
Version: 2.3.1
Description: libechonest
Cflags: -I/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/qt/include/
Libs: -lechonest

```
QtCore.pc contain variable ```includedir=${prefix}/include/QtCore``` so: 
```
Cflags: -DQT_SHARED -DQT_THREAD_SUPPORT -I/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/qt/include -I/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/qt/include/QtCore
```

**Comment by starius** [comment]
> How to configure Cflags for the general case (not only QtCore and QtNetwork)?

You need only QtCore, QtNetwork and QtTest, because libeachonest [depends](https://github.com/lfranchi/libechonest) only on them:

> Qt, only the QtCore, QtNetwork and QtTest modules (www.qtsoftware.com)

You can add `Requires: QtCore QtNetwork QtTest` and pkg-config should add their Cflags and Libs automatically. See https://secure.freedesktop.org/~dbn/pkg-config-guide.html#using

**Comment by starius** [comment]
Builds successfully. Files installed:

```
  i686-w64-mingw32.shared/bin/libechonest.dll           
  i686-w64-mingw32.shared/bin/test-libechonest.exe      
  i686-w64-mingw32.shared/include/echonest/             
  i686-w64-mingw32.shared/installed/libechonest         
  i686-w64-mingw32.shared/lib/libechonest.dll.a         
  i686-w64-mingw32.shared/lib/pkgconfig/libechonest.pc  
  i686-w64-mingw32.static/installed/libechonest         
  x86_64-w64-mingw32.shared/bin/libechonest.dll         
  x86_64-w64-mingw32.shared/bin/test-libechonest.exe    
  x86_64-w64-mingw32.shared/include/echonest/           
  x86_64-w64-mingw32.shared/installed/libechonest       
  x86_64-w64-mingw32.shared/lib/libechonest.dll.a       
  x86_64-w64-mingw32.shared/lib/pkgconfig/libechonest.pc
  x86_64-w64-mingw32.static/installed/libechonest       
```

Looks good for me. I would add trivial libechonest-test.cpp file to test pc config.

starius:
Ping. I think, it is time to rebase and merge this.

tonytheodore:
Thanks! Rebased and merged: https://github.com/mxe/mxe/commit/2794426ecd5f9c11f97d57355b4cac244a49f054.

PR #1117, opened 2016-01-01
build-pkg: add toposort checks and implement it internally, without tsort tool

starius:
Log of full build of commit 815d8fe7e18deb2ae50ade063f98cfdd6d155581 + 3 commits of this pull request: https://gist.github.com/45615365c6fb51beca77

### DIFF
--- a/tools/build-pkg.lua
+++ b/tools/build-pkg.lua
@@ -261,6 +261,30 @@ local function sortForBuild(items, item2deps)
     return build_list
 end
 
+-- return if build_list is ordered topologically
+local function isTopoOrdered(build_list, items, item2deps)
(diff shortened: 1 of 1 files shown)


tonytheodore:
What is the purpose of these two commits?
 * build-pkg: check each item builds one package	
 * build-pkg: check toposort

Are they checks for the `toposort` algorithm or did you find examples of each behaviour?

starius:
They are checks for the `toposort` algorithm and its input (`make print-deps-for-build-pkg`). I remember we had this problem in past, when build-pkg used to build multiple packages at a time.

tonytheodore:
I see, thanks!

Issue #1108, opened 2015-12-29
Problems with Archlinux


suizokukan:
# my configuration :
* an up to date Archlinux (`Linux archsuizokukan 4.2.5-1-ARCH #1 SMP PREEMPT Tue Oct 27 08:13:28 CET 2015 x86_64 GNU/Linux`)
* alure is installed (a `$pacman -Ss alure` returns `community/alure 1.2-4 [installed]`)
* a fresh installation of mxe : `git clone https://github.com/mxe/mxe`, `cd mxe`, then `make MXE_TARGETS='x86_64-w64-mingw32.static'`

# the error :
`make MXE_TARGETS='x86_64-w64-mingw32.static'` compiles a lot of stuff (mxe-conf, gcc... zlib, ...) and exits returning the following message :
> Failed to build package alure for target x86_64-w64-mingw32.static!

# the logs from log/alure_x86_64-w64-mingw32.static :
> [ 92%] Building CXX object CMakeFiles/ALURE32-static.dir/src/codec_mpg123.obj
> /home/suizokukan/mxe20151229/usr/bin/x86_64-w64-mingw32.static-g++   -DDYNLOAD=1 -DHAVE_CONFIG_H -D_FILE_OFFSET_BITS=64 -D_WIN32 -DAL_LIBTYPE_STATIC -DALURE_STATIC_LIBRARY -O3 -DNDEBUG @CMakeFiles/ALURE32-static.dir/includes_CXX.rsp   -Wextra -Wall -funswitch-loops  -DALURE_STATIC_LIBRARY -o CMakeFiles/ALURE32-static.dir/src/codec_mpg123.obj -c /home/suizokukan/mxe20151229/tmp-alure-x86_64-w64-mingw32.static/alure-1.2/src/codec_mpg123.cpp
> /home/suizokukan/mxe20151229/tmp-alure-x86_64-w64-mingw32.static/alure-1.2/src/codec_mpg123.cpp:33:20: fatal error: mpg123.h: No such file or directory
> #include <mpg123.h>

Any help would be appreciated !

starius:
1. Please upload full logs of packages `mpg123` and `alure` to [Gist](https://gist.github.com/) and post the link.
2. Please provide the output of
    * `find usr/* -name mpg123.h`
    * `find usr/* -name '*mpg123*'`

suizokukan:
Hi, thanks for your answer.

* `$find usr/* -name mpg123.h` : no file
[...]

Issue #1023, opened 2015-12-01
cmake: incorrect PKG_CONFIG_EXECUTABLE setting is used


uwehermann:
I was under the impression that an issue we're seeing was the same as #904, but I don't think that is the case anymore. Here's what seems to have been changed and goes wrong.

The setup:

A CMakeLists.txt file for a project is using something like this to find pkg-config-using libraries:
```
include(FindPkgConfig)
...
list(APPEND PKGDEPS libfoo>=x.y.z)
find_package(PkgConfig)
pkg_check_modules(PKGDEPS REQUIRED ${PKGDEPS})
```

Old (working) mxe.conf had this:
```set(PKG_CONFIG_EXECUTABLE /home/uwe/mxe-git/usr/bin/i686-w64-mingw32.static-pkg-config)```

New (not working) mxe.conf has this:
```
file(GLOB mxe_cmake_files
    "/home/uwe/mxe-git/usr/i686-w64-mingw32.static/share/cmake/mxe-conf.d/*.cmake"
)
foreach(mxe_cmake_file ${mxe_cmake_files})
    include(${mxe_cmake_file})
endforeach()
```
And ```/home/uwe/mxe-git/usr/i686-w64-mingw32.static/share/cmake/mxe-conf.d/pkgconf.cmake``` contains:
```set(PKG_CONFIG_EXECUTABLE /home/uwe/mxe-git/usr/bin/i686-w64-mingw32.static-pkg-config CACHE PATH "pkg-config executable")```

The problem/difference is the "CACHE" here, I think. With the above ```PKG_CONFIG_EXECUTABLE``` is ```/usr/bin/pkg-config``` before *and* after the ```set``` line. When dropping the ```CACHE PATH "pkg-config executable"``` part, the ```PKG_CONFIG_EXECUTABLE``` after the ```set``` line is now correct again:

```
/home/uwe/mxe-git/usr/bin/i686-w64-mingw32.static-pkg-config
```

Using the correct ```PKG_CONFIG_EXECUTABLE``` the host cmake now finds the packages correctly again.

Please double-check if performing this change makes sense for MXE and all its requirements, I only checked one very specific use-case here.

starius:
CACHE was added in 5a81e60e04efab090e4f90772666857260b4e2d3, #896

tonytheodore:
From cmake docs:
>Since cache entries are meant to provide user-settable values this does not overwrite existing cache entries by default. Use the FORCE option to overwrite existing entries.

[...]

PR #1118, opened 2016-01-01
alure: disable FluidSynth support

starius:
See https://github.com/mxe/mxe/issues/1108
See https://github.com/mxe/mxe/issues/1023


tonytheodore:
Thanks!

PR #1120, opened 2016-01-01
patch-tool-mxe: always output text diffs

starius:
Otherwise it provides binary diff for jack/waf file, which is a mix of Python and tar.


tonytheodore:
Thanks!

Issue #1122, opened 2016-01-01
MXE_GET_GITHUB_TAGS should use "latest release" if it exist


starius:
Repository https://github.com/kdeforche/wt/ uses releases `version-rcX` for draft releases. `$(call MXE_GET_GITHUB_TAGS, kdeforche/wt)` finds `3.3.5-rc2` instead of `3.3.5`. We can fix this either by filtering out versions matching `-rc` or by checking API URL https://api.github.com/repos/kdeforche/wt/releases/latest before tags URL.

tonytheodore:
The next iteration of #1002 has common and custom tag filtering:

```
# Packages must set the following metadata:
#   Track branch - Tarball API
#     GH_CONF := owner/repo/branch
#     updates will use the last commit from the specified branch as
#     a version string and bypass `sort -V`
#
#   Track tags - Archive API
#     GH_CONF := owner/repo, tag prefix, tag suffix, tag filter, version separator
#     updates will construct a version number based on:
#     <tag prefix><s/<version sep>/./version><tag suffix>

# common tag filtering is applied with `grep -v`:
GITHUB_TAG_FILTER := 'alpha\|beta\|rc'
```

I hit a snag with a new download API in `libass` so I think I'll just tidy it up, submit a new PR and iterate from there.

tonytheodore:
See `mxe.github.mk` for latest usage, `MXE_GET_GITHUB_TAGS` has been removed.

PR #1123, opened 2016-01-01
wt: update from 3.3.4 to 3.3.5

starius:
Remove patches merged in upstream.

See also https://github.com/mxe/mxe/issues/1122


tonytheodore:
Thanks!

PR #1124, opened 2016-01-02
Fix aubio for python 3

starius:
see #1108

### DIFF
--- a/src/aubio-1-fixes.patch
+++ b/src/aubio-1-fixes.patch
@@ -178,3 +178,407 @@ index 1111111..2222222 100644
  
      bld( source = 'aubio.pc.in' )
  
+
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Boris Nagaev <bnagaev@gmail.com>
+Date: Sat, 2 Jan 2016 22:16:34 +0300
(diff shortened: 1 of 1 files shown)


tonytheodore:
Works with python 2.7.1 but strange errors with 3.5.1 https://gist.github.com/tonytheodore/50d78962897f792f64da

starius:
I think, it worth upgrading whole waf. There is a problem with [how aubio upgrades waf](https://github.com/aubio/aubio/blob/master/Makefile) and [waf itself](https://github.com/waf-project/waf/issues/1677).

tonytheodore:
> I think, it worth upgrading whole waf

Can we put it in `$(BUILD)/bin` and use that instead of `./waf`?

starius:
I think, we should add waf as a package.

starius:
See https://github.com/waf-project/waf/issues/1677#issuecomment-168491796
> The waf file does unpack a content that is dependent on the python version in order to support Python 2.5.

We should use binary "waf" file rather than "waflib".

starius:
The problem was reported to aubio project: https://github.com/aubio/aubio/issues/41

starius:
Replaced with https://github.com/mxe/mxe/pull/1128 and package waf (not yet published).

PR #1129, opened 2016-01-03
[WIP] add package waf and use it from jack and aubio

starius:


starius:
Fails to build jack.

starius:
By the way, the following trick can be used not to rebuild all packages:
```
touch usr/*/installed/*
touch src/aubio.mk src/jack.mk
make aubio jack
```
Can we deal with global rebuilds after adding new native packages?

starius:
It builds successfully here (Python 2).

TimothyGu:
I'm not sure about this. Waf is just like autoconf in that every version doesn't absolutely preserve backwards compatibility. I'd recommend doing this on an ad hoc basis. Now I can see that it fixes an issue with aubio, but it might be more work than necessary to do this for jack if it already builds fine as is.

tonytheodore:
> Can we deal with global rebuilds after adding new native packages?

Yes, we'll switch to explicit dependencies. See https://github.com/mxe/mxe/issues/516#issuecomment-161567909

starius:
@TimothyGu, I think patching jack's and aubio's wafs is not ideal, as those wafs are old and don't support Python 3. If we create a [wrapper](https://github.com/mxe/mxe/issues/1108#issuecomment-168581756) for python, we would require Python 2, not Python 3. So MXE would become unusable on systems without Python 2, but with Python 3, e.g. [default installation of Arch Linux](https://wiki.archlinux.org/index.php/Python#Python_2). New waf is known to work with both Python 2 and Python 3.

> Waf is just like autoconf in that every version doesn't absolutely preserve backwards compatibility

I think, it is like CMake in that it has few dependencies and it is easy to ship.

> might be more work than necessary to do this for jack if it already builds fine as is

In this case we can build aubio with our waf and jack with its own jack :-)
However I believe jack can be fixed to use our waf.

TimothyGu:
Ping.

> In this case we can build aubio with our waf and jack with its own [waf] :-)

I still think that this is the best way to go before any issues arise with jack.

starius:
Pong.

The problem is here: https://github.com/mxe/mxe/issues/1108#issuecomment-168549984

I'm investigating it.

starius:
Now jack uses jack's waf (as before) and aubio uses MXE's waf.

TimothyGu:
Let's merge this then. It allows one to test #1139 more easily and if it broke something we could just revert it.

Issue #134, opened 2013-02-06
patch workflow


mabrand:
Continued from https://github.com/mxe/mxe/issues/128

> I'm certainly not against using git for this, but even just the few steps you mentioned could clearly be automated to facilitate faster patching, while still using git.

Maybe, but there are lots of variables. Many of the packages in MXE have patches that not managed with git, maybe because the contributor uses other methods. Some of the git-generated patches in MXE are based on branches split off from official repo branches, while others are based on tarballs. When you are starting with a git-generated patch file in MXE, you might use "git am" to recreate the branch in a new repo, or just continue work on the branch that was originally used to generated it. In other cases, you might want to import non-git-generated patches from MXE into a git branch, which ususally takes more work.

PR #1134, opened 2016-01-04
reimplement patch-tool-mxe in Makefile

### LINKED ISSUE #1063: patch-tool-mxe doesn't work with plugins
```
$ MXE_PLUGIN_DIRS=plugins/apps/ ./tools/patch-tool-mxe init tor             
grep: /home/mxe/mxe/src/tor.mk: No such file or directory
grep: /home/mxe/mxe/src/tor.mk: No such file or directory
grep: /home/mxe/mxe/src/tor.mk: No such file or directory
Error: /home/mxe/mxe/gits/ already exists. Cancelling init.
```

starius:
fix #1063


tonytheodore:
Looks good at first glance!

One thing I'd like to do is move new functions into (say) `Makefile.patch` and `include` them from the main `Makefile`. Then we can gradually switch to a more modern [naming convention](https://www.gnu.org/software/make/manual/html_node/Using-Variables.html#Using-Variables):

>It is traditional to use upper case letters in variable names, but we recommend using lower case letters for variable names that serve internal purposes in the makefile, and reserving upper case for parameters that control implicit rules or for parameters that the user should override with command options (see Overriding Variables).

starius:
Maybe we can instead include main Makefile from Makefile.patch, since functions from Makefile.patch are needed only in Makefile.patch?

Another thing I want to do is to add `import-all-patches` feature, because it is tedious to copy and paste names of all patches of e.g. qbittorrent.

tonytheodore:
> Maybe we can instead include main Makefile from Makefile.patch, since functions from Makefile.patch are needed only in Makefile.patch?

That would then require calling `make -f Makefile.patch` which would become tedious and is harder to integrate with the tab completion in `tools/mxe-activate`. The latter becomes especially convenient in the patch generation case as it can use an alias that can be invoked from any directory.

Also, down the track, it would be good if all package builds use the same patching logic - at least to automatically prepare the repo for failing builds, see:

http://lists.nongnu.org/archive/html/mingw-cross-env-list/2010-10/msg00018.html
https://github.com/mxe/mxe/issues/134

In that case, the main Makefile would also use those functions.

starius:
Patching functions were moved to patch.mk.

starius:
`-M9` added to `git format-patch`
rebased

tonytheodore:
All related targets should also be in `patch.mk`

starius:
Targets were moved to `patch.mk`.
Rebased.

tonytheodore:
Thanks!

PR #1136, opened 2016-01-05
Update qbittorrent from 3.2.3 to 3.3.1

starius:
Build result: http://redjohn.tk/mxe/qbittorrent.exe
One patch was sent to upstream: https://github.com/qbittorrent/qBittorrent/pull/4505


tonytheodore:
Thanks!

PR #1138, opened 2016-01-06
add selected plugins to backup download

### LINKED ISSUE #1135: add packages of plugins to backup server
```
$ MXE_PLUGIN_DIRS=plugins/apps/ make qbittorrent
[download]  qbittorrent
MXE Warning! Downloading qbittorrent from backup.

Download failed or wrong checksum of package qbittorrent!
--2016-01-05 20:33:52--  http://d1yihgixbnrglp.cloudfront.net/qbittorrent%2D3%2E2%2E3%2Etar%2Exz
Resolving d1yihgixbnrglp.cloudfront.net (d1yihgixbnrglp.cloudfront.net)... 54.192.206.22, 54.192.206.221, 54.192.206.86, ...
Connecting to d1yihgixbnrglp.cloudfront.net (d1yihgixbnrglp.cloudfront.net)|54.192.206.22|:80... connected.
HTTP request sent, awaiting response... 404 Not Found
2016-01-05 20:33:53 ERROR 404: Not Found.
Download failed!
```

SourceForge sometimes returns 404 for its files.

tonytheodore:
Plugins like gcc52 cause the current file to be deleted so just use a whitelist of directories.

closes #1135

### DIFF
--- a/tools/s3-fetch-and-sync
+++ b/tools/s3-fetch-and-sync
@@ -21,6 +21,10 @@
 # Useful to prevent the same issue from being filed multiple times
 file_issue=true
 
+# List of plugin dirs to include in downloads
+# Can't include all subdirs since some affect versions
+plugin_dirs=`echo plugins/{apps,luarocks,native,tcl.tk}`
+
 cd ~/mxe && git pull
 
 # Test downloading without falling back to S3 download server.
@@ -28,6 +32,7 @@ cd ~/mxe && git pull
 if ! ( \
     cd ~/mxe && \
         make download -k MXE_NO_BACKUP_DL=true MXE_VERBOSE=true \
+            MXE_PLUGIN_DIRS="$plugin_dirs" \
             2>&1 >tmp-download-log \
 ) && $file_issue; then
     # If one or more download process fails, upload log to sprunge.us (a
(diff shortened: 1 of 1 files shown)


starius:
> Plugins like gcc52 cause the current file to be deleted

Does it mean that only one version of tarball per package is stored? I think, all versions should be kept to guarantee that MXE commits from past are also buildable.

tonytheodore:
No, all versions are kept on S3, this only affects how they are first uploaded. The gcc52 plugin overrides the current version so would never see the next update (if it was included).

We could go the other way and blacklist, but I don't think were going to end up with a lot of plugin directories.

starius:
Whitelist is Ok. Please add a reminder about it to plugins/README.md.

tonytheodore:
Cron job was broken by https://github.com/mxe/mxe/commit/32b76a932d4b785c011b584102adceb49211371a as the local file had been modified:

```
mxe@mxe:~/mxe$ git diff
diff --git a/tools/s3-fetch-and-sync b/tools/s3-fetch-and-sync
index 5b1a943..e96c08e 100755
--- a/tools/s3-fetch-and-sync
+++ b/tools/s3-fetch-and-sync
@@ -16,7 +16,7 @@

 # Manual switch to enable/disable issue filing on GitHub.
 # Useful to prevent the same issue from being filed multiple times
-file_issue=true
+file_issue=false

 cd ~/mxe && git pull
```

starius:
`file_issue` doesn't file an issue on Github. It uploads log to sprunge. Maybe gist is better?

tonytheodore:
I've been experimenting with gist https://github.com/tonytheodore/mxe/issues/6, seems to work fine.

PR #1144, opened 2016-01-06
A proper patch for compiling qbittorrent

ZachBacon:
Just worked with another guy in the #qbitorrent channel on freenode and this patch is slightly more better for compiling qbittorrent, it changes configure.ac and qbittorrent.m4 to properly detect qmake in mxe and also for my personal favourite build system msys2 when I'm on windows, figured you'd like to use it, the other changes I'm going to push also into qbittorrent git as well (as I saw one of you guys already pushed a PR for the caps issue)

### DIFF
--- a/plugins/apps/qbittorrent-1-fixes.patch
+++ b/plugins/apps/qbittorrent-1-fixes.patch
@@ -1,70 +1,45 @@
-This file is part of MXE.
-See index.html for further information.
-
-Contains ad hoc patches for cross building.
-
-From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
-From: Boris Nagaev <bnagaev@gmail.com>
-Date: Sun, 30 Aug 2015 00:28:50 +0200
-Subject: [PATCH] do not check qmake existance
-
-Fix ./configure error:
-
-checking for mxe/usr/i686-w64-mingw32.static/qt/bin/qmake...
-configure: error: cannot check for file existence when
-cross compiling
-
-diff --git a/configure b/configure
-index 1111111..2222222 100755
---- a/configure
-+++ b/configure
-@@ -4500,7 +4500,7 @@ if eval \${$as_ac_File+:} false; then :
-   $as_echo_n "(cached) " >&6
- else
-   test "$cross_compiling" = yes &&
--  as_fn_error $? "cannot check for file existence when cross compiling" "$LINENO" 5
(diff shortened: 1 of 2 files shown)


starius:
Thank you! I prefer patches over diffs. I converted your diff back to patch: https://github.com/DoctorWho11/mxe/pull/1

starius:
Fails to build: https://gist.github.com/3644e5edaa6175cef784

Fix typo: s/boostrap/bootstrap/

DoctorWho11:
yeah I noticed that...

starius:
After fixting a typo in "bootstrap.sh", fails with linking errors: https://gist.github.com/31c744964977e50c5155

DoctorWho11:
it links for me at least 64bit, another thing though is that I primarily built with release basically make release. The patch basically follows the original just moved the configure changes into configure.ac I'll do some more testing via some vm's

starius:
Same linking error on x86_64-w64-mingw32.static: https://gist.github.com/27646a32e517cc66a7dd

tonytheodore:
`bootstrap.sh` only calls `aclocal` and `autoconf`, it's usually better to invoke `autoreconf -fi`

tonytheodore:
> just moved the configure changes into configure.ac

Something happened in converting back to patch: https://github.com/DoctorWho11/mxe/blob/422700bb68bf78027d5bb438fc25341d4e471bf6/plugins/apps/qbittorrent-1-fixes.patch#L115-L135

starius:
This work is not finished.

> Something happened in converting back to patch: https://github.com/DoctorWho11/mxe/blob/422700bb68bf78027d5bb438fc25341d4e471bf6/plugins/apps/qbittorrent-1-fixes.patch#L115-L135

Something is wrong with GitHub highlighting. Here these lines are not highlighted:
https://github.com/DoctorWho11/mxe/blob/master/plugins/apps/qbittorrent-1-fixes.patch

tonytheodore:
That was me highlighting the patch. The relevant lines are https://github.com/DoctorWho11/mxe/blob/master/plugins/apps/qbittorrent-1-fixes.patch#L123-L126:

```
diff --git a/configure b/configure
index 1111111..2222222 100755
--- a/configure
+++ b/configure
```

`configure` is being patched and regenerated. These changes should be in `configure.ac` (mentioned above).

tonytheodore:
Ping following #1307.

tonytheodore:
Lots of updates to qbittorrent since, better to work on updating it.

Issue #1111, opened 2015-12-30
Finding undeclared dependencies


starius:
Following on from #1110, #1039, 068e20243d2b531c032f04fcff07bd613594d9ce, 20342ed8294bdddb6de27d415b802640f340fc4d, 469ccb819c48bb24e67bf74fd4f7e45334fa326a, sometimes a package doesn't have all dependencies in its dependency list. A build may still succeed because the missing dependency happens to be built. Such bugs are hard to detect because they happen rarely.

There are two types of undeclared dependencies:
1. undeclared requiremenets. If a requirement is absent, a following package or its descendant fails to build.
2. Undeclared optional dependencies. Presence of optional dependency affects a following package or its descendant: sets of files differ or (harder to test) sets of features differ.

I found two techniques to find undeclared dependencies.

Consider two packages, a and b, such that each of them can be built without building another one. We call them *independent packages*.

## Git branches (whitebox approach)

When a package is built, the only built packages are its own dependencies. This can be implemented as a part of `build-pkg`: a git branch is created for each package; its parents are git branches of package direct dependencies. So build list is replaced with build graph. This approach guarantees that no independent packages of package `a` present when package `a` is built. So it solves problem 1 (undeclared requiremenets).

It can be enhanced to solve problem 2 as well (Undeclared optional dependencies). Having built all packages, rebuild each of them one more time, but use all packages except its followers as parents. Now all independent packages of package `a` present when package `a` is built. So each package has two builds: before independent packages and after independent packages. Compare files produced by these two builds and find discrepancies.

## Many build lists (blackbox approach, not working atm)

(This approach seems to be of thoretical value only.)

To detect effects between each pair of packages, one build is not sufficient. To make sure that `a` is not `b`'s undeclared dependency and vice versa, two builds are required. I studied if two builds are sufficient to cover all pairs of unrelated packages in any build graph. Unfortunately some build graphs require more than two builds to cover all unrelated pairs. For example, the following graph ("heart"):

```
      a -> b
      |    |
      v    |
 c -> d    |
 |         |
 v         v
 e ------> f

```

Here `a -> b` means "b depends on a". No pair of builds can cover all pairs of independent packages. However it can be covered in 3 builds:

  * abcdef
  * ceabfd
  * acdebf

[...]

PR #1149, opened 2016-01-07
mxe-conf: create basic non-empty directory hierarchy

tonytheodore:
https://github.com/mxe/mxe/pull/1061/files#r49049496
https://github.com/mxe/mxe/issues/1111#issuecomment-169280181

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -329,10 +329,13 @@ define CHECK_REQUIREMENT_VERSION
     fi
 
 endef
-$(shell [ -d '$(PREFIX)/installed' ] || mkdir -p '$(PREFIX)/installed')
+
+%/.gitkeep:
+	+@mkdir -p '$(dir $@)'
+	@touch '$@'
 
 check-requirements: $(PREFIX)/installed/check-requirements
-$(PREFIX)/installed/check-requirements: $(MAKEFILE)
+$(PREFIX)/installed/check-requirements: $(MAKEFILE) | $(PREFIX)/installed/.gitkeep
 	@echo '[check requirements]'
 	$(foreach REQUIREMENT,$(REQUIREMENTS),$(call CHECK_REQUIREMENT,$(REQUIREMENT)))
 	$(call CHECK_REQUIREMENT_VERSION,autoconf,2\.6[8-9]\|2\.[7-9][0-9])
@@ -349,7 +352,7 @@ $(PREFIX)/installed/check-requirements: $(MAKEFILE)
 
 .PHONY: print-git-oneline
 print-git-oneline: $(PREFIX)/installed/print-git-oneline-$(GIT_HEAD)
-$(PREFIX)/installed/print-git-oneline-$(GIT_HEAD):
+$(PREFIX)/installed/print-git-oneline-$(GIT_HEAD): | $(PREFIX)/installed/.gitkeep
 	@git log --pretty=tformat:'[git-log]   %h %s' -1 | cat
 	@rm -f '$(PREFIX)/installed/print-git-oneline-'*
 	@touch '$@'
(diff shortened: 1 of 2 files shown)


starius:
IMHO, `s/mxe-hier/.gitkeep/`

starius:
Two methods of creating a directory are used in the same file:

  * `mkdir -p` (CMAKE_TOOLCHAIN_DIR and BUILD)
  * `install -d`

Can you use `mkdir -p` instead of `install -d`?

starius:
Can you also include my commit 4fb55fe5a16c32a29a8b0353175691bfbd643b9a into this pull request?

tonytheodore:
> IMHO, s/mxe-hier/.gitkeep/

Agreed, that's much clearer.

> Can you use mkdir -p instead of install -d?

Done, I didn't know `mkdir` had the `-m` option.

Added https://github.com/mxe/mxe/commit/4fb55fe5a16c32a29a8b0353175691bfbd643b9a and rebased.

starius:
```
$ make mxe-conf MXE_TARGETS=x86_64-unknown-linux-gnu
$ find usr/ -empty -type d
usr/readonly

$ make mxe-conf
# in other terminal:
$ find usr/ -empty -type d
usr/readonly
usr/i686-w64-mingw32.static/installed
```

By the way, `find usr/ -empty -type d` is a good check for `build-pkg`.

starius:
List of empty directories in built MXE:
https://gist.github.com/starius/2b1baa915a1b51e712e7

starius:
`usr/x86_64-unknown-linux-gnu/bin/` has no `.gitkeep`

tonytheodore:
I've added `usr/readonly`, but it requires a `chmod 0755` before `clean`. Not sure if there's a better way to handle it.

> usr/x86_64-unknown-linux-gnu/bin/ has no .gitkeep

It shouldn't need one as `wine` should always be present.

> usr/i686-w64-mingw32.static/installed

This shouldn't be empty after the first package is built so `git` will track it.

starius:
> I've added usr/readonly, but it requires a chmod 0755 before clean. Not sure if there's a better way to handle it.

Remove a directory, not a file in it.

>>usr/i686-w64-mingw32.static/installed

>This shouldn't be empty after the first package is built so git will track it. 

It is empty while `x86_64-unknown-linux-gnu` packages are building. They are separate items in `build-pkg`.

tonytheodore:
> Remove a directory, not a file in it.

I was a little surprised by this too, a directory needs both write and execute permissions to be deleted:
```
$ mkdir -p tmp/readonly
$ touch tmp/readonly/.gitkeep
$ chmod 0444 tmp/readonly
$ find tmp/
tmp/
tmp//readonly
find: tmp//readonly/.gitkeep: Permission denied
find: tmp//readonly: Permission denied

$ chmod 0555 tmp/readonly
$ find tmp/
tmp/
tmp//readonly
tmp//readonly/.gitkeep

$ rm -rf tmp/
rm: tmp//readonly/.gitkeep: Permission denied
rm: tmp//readonly: Directory not empty
rm: tmp/: Directory not empty

$ chmod 0755 tmp/readonly/
$ rm -rf tmp/
```

The errors with `0444` permissions are almost comical as they reveal the contents anyway.

> It is empty while x86_64-unknown-linux-gnu packages are building. They are separate items in build-pkg

Good point, I don't think this can be fixed in mxe-conf (at least not in a way that allows `make -t` to work).

starius:
> at least not in a way that allows make -t to work

I think it is not critical for `build-pkg`, as this directory is created on each item, so `touch '$(PREFIX)/$(3)/installed/$(1)'` may not fail. But `build-pkg` will produce a message about empty directory.

BTW, `build-pkg` [uses](https://github.com/mxe/mxe/issues/1111#issuecomment-170243406) `touch` instead of `make -t`.

tonytheodore:
Okay, the `*/installed` directories are now only created as order-only dependencies and won't be empty. The `.gitkeep` files in those are just to keep `make -t` happy having a file to work with instead of a directory - it's useful for debugging sometimes to see what `make` thinks it needs to do.

starius:
Looks good!

```
$ make mxe-conf MXE_TARGETS=x86_64-unknown-linux-gnu
$ find usr/ -empty -type d
<nothing>
```

Issue #1112, opened 2015-12-30
Failed to build package gtk2 for target i686-w64-mingw32.static


pavelvat:
https://gist.github.com/pavelvat/545d72d7dfc16a64c70d

starius:
The error message:
```
(here=`pwd` ; cd . && /usr/bin/perl $here/geninclude.pl appwindow.c assistant.c builder.c button_box.c changedisplay.c clipboard.c colorsel.c combobox.c dialog.c drawingarea.c editable_cells.c entry_buffer.c entry_completion.c expander.c hypertext.c iconview.c iconview_edit.c images.c infobar.c links.c list_store.c menus.c offscreen_window.c offscreen_window2.c panes.c pickers.c pixbufs.c printing.c rotated_text.c search_entry.c sizegroup.c spinner.c stock_browser.c textview.c textscroll.c toolpalette.c tree_store.c ui_manager.c) > demos.h
Can't use 'defined(@array)' (Maybe you should just omit the defined()?) at /home/pavel/builds/mxe/tmp-gtk2-i686-w64-mingw32.static/gtk+-2.24.22/demos/gtk-demo/geninclude.pl line 43.
Makefile:1024: recipe for target 'demos.h' failed
```

This error was fixed in version 2.24.28:
https://git.gnome.org/browse/gtk+/commit/?h=gtk-2-24&id=3da748930fdcb19aaae32ae2fc6cb2d802857752

starius:
I have updated `gtk2` in #1113.
Can you test it please?

pavelvat:
> Can you test it please?

I do not use gtk2. I just build gst-plugins-good and how I see #1113 you already tested it.

starius:
I built gst-plugins-good and other packages successfully, but I don't use it, so I am not sure it works correctly.

PR #1151, opened 2016-01-07
print Perl version to log

starius:
See https://github.com/mxe/mxe/issues/1108#issuecomment-169556116
See https://github.com/mxe/mxe/issues/1112

From log:
```
perl --version 2>&1 | head -3

This is perl 5, version 18, subversion 2 (v5.18.2) built for x86_64-linux-gnu-thread-multi
(with 41 registered patches, see perl -V for more detail)
```


tonytheodore:
Thanks!

PR #1152, opened 2016-01-08
pcre: enable C++

starius:
Application EiskaltDC++ requires pcre with C++ support.

See https://github.com/mxe/mxe/pull/1127#issuecomment-169418284

All packages were built successfully on all 4 targets (cb8a7deb8fe5259b89ad2e5a986b477fa5808162 + this commit). Then rebased.

/cc @pavelvat


tonytheodore:
Thanks!

PR #1153, opened 2016-01-08
update taglib from 1.7.2 to 1.10

starius:
Remove [the patch](https://bugs.kde.org/show_bug.cgi?id=268441) which should be in upstream. The project was moved to GitHub.

Packages `taglib` and its dependency `gst-plugins-good` were built successfully on static targets. Shared targets are disabled.

See http://lists.nongnu.org/archive/html/mingw-cross-env-list/2016-01/msg00001.html


tonytheodore:
> Shared targets are disabled.

These build and seem to install dlls/libs correctly. https://github.com/LuaAndC/mxe/pull/2. (`gst-plugins-good` shared is still disabled).

tonytheodore:
Thanks!

PR #1156, opened 2016-01-09
mxe-activate: update completions and change alias to mxe-make

tonytheodore:
* remove `build-only*` - too low level for general use
* add new patch related commands and catch-all completions
* `mxe-make` alias is more descriptive


tonytheodore:
I'm probably the only person that uses this and I like the new patch targets - especially combined with ` wget -O- ... | git am`

PR #1157, opened 2016-01-09
gstreamer updates and shared builds

tonytheodore:
Tested on all four targets.

* `liboil` is in maintenance only phase
* `libshout` shared build still disabled
* `gst-plugins-good` doesn't depend on `gtk2`


tonytheodore:
Changed to WIP to see if the whole chain can be updated.

tonytheodore:
Removed [WIP] and rebased, should be good to go.

starius:
openthreads is broken: https://gist.github.com/233e125933113e0c08b8

tonytheodore:
I see, a case of the third pass you mentioned in https://github.com/mxe/mxe/issues/1111#issuecomment-170243406. All these optional packages should be disabled.

tonytheodore:
openthreads fixed in https://github.com/tonytheodore/mxe/commit/8d4c213de874f44c89f9ba9979d0686e370efe34

tonytheodore:
Successfully retested (i686-w64-mingw32.static) with building over existing full build to weed out ordering conflicts.

starius:
Successfully built on all 4 targets, all packages.

tonytheodore:
Thanks!

Issue #1163, opened 2016-01-09
nsis: remove ELF binaries from usr/<target>/bin


starius:
See https://github.com/mxe/mxe/pull/1127#issuecomment-170286752

Whole list of files installed by nsis (i686-w64-mingw32.static):
https://gist.github.com/8df9210a44ad9ead741a

I think, the following files should be removed:
- usr/i686-w64-mingw32.static/bin/GenPat
- usr/i686-w64-mingw32.static/bin/LibraryLocal
- usr/i686-w64-mingw32.static/bin/makensis

Files GenPat and LibraryLocal can be moved to `/usr/bin/<target>-tool`. GenPat seems to be useless in MXE, because it is needed in development, not in build. I am not sure about LibraryLocal.

GenPat - A program to create patch files in VPatch format.
LibraryLocal - A program to get version numbers from DLL files for use with NSIS.

tonytheodore:
Strange things going on with [nsis releases](http://sourceforge.net/p/nsis/code/HEAD/tree/NSIS/tags/), after six years, they do five releases in the past month? Might be worth updating to the latest first.

tonytheodore:
I wonder if the build can be split into native and cross components?

starius:
Debian has packages `nsis` and `nsis-common`. E.g., executables belong to `nsis` and DLLs belong to `nsis-common`. Targets are as follows:

  * INSTALL_nsis = install-utils install-compiler install-conf
  * INSTALL_nsis-common = install-stubs install-plugins install-data install-utils

(from debian/rules)

`install-compiler` installs `makensis`.

PR #1164, opened 2016-01-10
nsis: update from 2.46 to 2.50

starius:
see https://github.com/mxe/mxe/issues/1163

Build is successful. `usr/bin/i686-w64-mingw32.static-makensis` was tested with packing NPG-exporer.


tonytheodore:
Thanks!

Issue #1143, opened 2016-01-06
commit b4decd3: incorrect linking with libaspell-15.dll


pavelvat:
after b4decd3 on launch application EiskaltDC++ has error: https://github.com/mxe/mxe/pull/1127#issuecomment-169441880 file is missing libdl.dll.a
Build of EiskalDC++ with -DUSE_ASPELL=OFF is working fine.

TimothyGu:
Huh, libdl.dll.a doesn't contain anything important; it only contains the entry points to the real library, which is libdl.dll.

pavelvat:
>Are you building static or shared target?

shared

starius:
I thought, *.dll.a are needed at build time and *.dll at run time. How can the application complain about *.dll.a at run time?

pavelvat:
>I thought, *.dll.a are needed at build time and *.dll at run time.

It is right.

>How can the application complain about *.dll.a at run time?

Yes, it is very strange. I am copying libdl.dll.a to execution directory and Windows say that libdl.dll.a is not correct dll file.

starius:
> Windows say that libdl.dll.a is not correct dll file.

Indeed, libdl.dll.a is not correct dll file.

Have you tried to copy libdl.dll instead of libdl.dll.a?

pavelvat:
>Have you tried to copy libdl.dll instead of libdl.dll.a?

Of course. But libdl.dll is not required.

pavelvat:
I can't reproduce this bug on Debian Sid and clean build MXE.
Now for target i686-w64-mingw32.shared occur SIGSEGV on exit from program:
http://pastebin.com/cwTKYrXk
For build with -DUSE_ASPELL=OFF don't occur SIGSEGV on exit from program.

pavelvat:
I reproduce this bug for target i686-w64-mingw32.shared on clean installation of Ubuntu 14.04 and clean build of MXE.

starius:
> Now for target i686-w64-mingw32.shared occur SIGSEGV on exit from program:

It may be a bug in the software. Have you run it under Valgrind (maybe in Linux, if there are problems in Windows)?

> I reproduce this bug for target i686-w64-mingw32.shared on clean installation of Ubuntu 14.04 and clean build of MXE.

Are you speaking about dynamic linking error or SIGSEGV?

pavelvat:
>Are you speaking about dynamic linking error or SIGSEGV?

about dynamic linking. 

copydlldeps.py output:
[...]

Issue #589, opened 2014-12-17
copydlldeps.py doesn't work for x86_64


alekseysidorov:
Tested on Debian unstable

```bash
['/home/aleksey/develop/crossdev/mxe/tools/copydlldeps.py', '/home/aleksey/develop/cleverCNC/build/crossdev//install_root/mingw32_sniper_x86_64/CleverCNC/bin', '-C', '/home/aleksey/develop/cleverCNC/build/crossdev//install_root/mingw32_sniper_x86_64/CleverCNC/bin', '/home/aleksey/develop/cleverCNC/build/crossdev//install_root/mingw32_sniper_x86_64/CleverCNC/bin/imageformats', '-L', '/home/aleksey/develop/crossdev/mxe/usr/x86_64-w64-mingw32.shared/bin', '/home/aleksey/develop/crossdev/mxe/usr/x86_64-w64-mingw32.shared/lib', '/home/aleksey/develop/crossdev/mxe/usr/x86_64-w64-mingw32.shared/qt5/bin']
Traceback (most recent call last):
  File "/home/aleksey/develop/crossdev/mxe/tools/copydlldeps.py", line 155, in <module>
    for dependency_dll in get_imports(pe_filename_full):
  File "/home/aleksey/develop/crossdev/mxe/tools/copydlldeps.py", line 73, in get_imports
    seek_to_rva(rva_of_import_table)
  File "/home/aleksey/develop/crossdev/mxe/tools/copydlldeps.py", line 69, in seek_to_rva
    raise ValueError('Could not find section for RVA.')
ValueError: Could not find section for RVA.

```

tonytheodore:
Should be fixed in https://github.com/mxe/mxe/compare/6f830ee...4d76a91

PR #1165, opened 2016-01-10
add eiskaltdcpp.mk to mxe/plugins/apps

pavelvat:
Problems:
1)```$(PKG)_BUILD_STATIC =``` because linked error for static builds: http://pastebin.com/A9C4T5Vx
It is can be fixed by adding names of depend libraries to https://github.com/eiskaltdcpp/eiskaltdcpp/blob/14bd7229c12f06a4251c8a57795e4121e1b53d01/eiskaltdcpp-qt/CMakeLists.txt#L346
How to get full list names of depend libraries for static Qt5?
2)```$(PKG)_BUILD_x86_64-w64-mingw32 =``` because mxe/tools/copydlldeps.py don't work for x86_64, see https://github.com/mxe/mxe/issues/589
3)https://github.com/mxe/mxe/pull/1127#issuecomment-170281302 so 'nsis' from:
```
$(PKG)_DEPS     := gcc zlib bzip2 libiconv gettext miniupnpc jsoncpp aspell libidn openssl pcre lua boost qt5 nsis
```
can not be build for x86_64-w64-mingw32.static x86_64-w64-mingw32.shared targets. It can be fixed in common case by using Linux package nsis instead nsis.mk.
4)Currently eiskaltdcpp-qt.exe don't launch due to https://github.com/mxe/mxe/issues/1143. It can be fixed by editing eiskaltdcpp.mk:
```
        -DUSE_ASPELL=OFF \
```
and disabled this lines
```
        $(TARGET)-makensis '$(1).build/EiskaltDC++.nsi'; \
        $(INSTALL) '$(1).build/EiskaltDC++-2.4.0-x86.exe' '$(PREFIX)/$(TARGET)/apps/$(PKG)/nsis'
```
because NSIS script required file libaspell-15.dll.
5)Installer enabled only for shared builds because NSIS script required refactoring for support static builds. I will make it in future.
6)line:
```
        $(INSTALL) '$(1).build/EiskaltDC++-2.4.0-x86.exe' '$(PREFIX)/$(TARGET)/apps/$(PKG)/nsis'
```
work only for i686-w64-mingw32. How to check "if(x86_64)"?


starius:
> How to get full list names of depend libraries for static Qt5?

Like this:

```
./usr/bin/i686-w64-mingw32.static-pkg-config -libs Qt5Core
```

Add other Qt components if you need them. Full list of Qt5's pc files can be found in `usr/i686-w64-mingw32.static/qt5/lib/pkgconfig/`.


> How to check "if(x86_64)"?

`$(if $(findstring x86_64,$(TARGET)),your code,)`

See `src/openblas.mk`

pavelvat:
1)Fixed.
2)Enable x86_64 build, but don't copy dll-depends due to https://github.com/mxe/mxe/issues/589
3)I think that the best option would be to add Linux package ```nsis``` to http://mxe.cc/#requirements
4)https://github.com/mxe/mxe/issues/1143#issuecomment-177262302
5)Installer disabled for x86_64_shared due to https://github.com/mxe/mxe/issues/589, Installer disabled for static builds because I don't know how to detect from NSIS at compile time is ```eiskaltdcpp-qt.exe``` 32-bit or 64-bit file. In order to detect 32-bit or 64-bit shared builds uses this: https://github.com/eiskaltdcpp/eiskaltdcpp/blob/master/windows/EiskaltDC%2B%2B.nsi#L3-L19, so I can detect static type if these two files don't exist simultaneously: ```libgcc_s_sjlj-1.dll```, ```libgcc_s_seh-1.dll```. But for static type I can't detect 32-bit or 64-bit case must be used to compile EiskaltDC++ installer.
6)Fixed.

starius:
3) Are you sure that system package `nsis` (e.g., from Debian) can build a 64-bit installer?

pavelvat:
64-bit installer does not need. 32-bit installer completely provides all that is needed.

starius:
@pavelvat, I think it is generally a bad idea to add a package to requirements, if it is built by MXE and can be used. The main problem here is that targets of `eiskaltdcpp` and `nsis` are different. This could be fixed, should MXE allow a cross-target dependency.

Can we add the following syntax to `_DEPS`:
```
$(PKG)_DEPS     :=  gcc zlib bzip2  ... i686-w64-mingw32.static~nsis
```
?

tonytheodore:
The mechanism in #516 (Add mechanism for depending on build system packages) for cross-target deps would cause entire toolchain to be built again. This may be better handled in #961 (Build gcc once) and I've added a note there.

pavelvat:
From https://github.com/Alexpux/MINGW-packages/pull/993:
>The bug in 64-bit installers has finally been fixed upstream.

tonytheodore:
Bump cross-target dependency

pavelvat:
Updated for Qt 5.6 and rebased.
Required multiCall from #1278:
``` 
        '$(TOP_DIR)/tools/copydlldeps.sh' -c \
                                          -d '$(PREFIX)/$(TARGET)/apps/$(PKG)/$(PKG)' \
                                          -F '$(PREFIX)/$(TARGET)/apps/$(PKG)/$(PKG)' \
                                          -F '$(PREFIX)/$(TARGET)/apps/$(PKG)/$(PKG)/sqldrivers' \
                                          -R '$(PREFIX)/$(TARGET)';
```
Current status:
1)Fixed
2)Fixed
3)required nsis for 64-bit mxe-targets (currently MXE contain nsis version 2.50, but version 2.51 is available https://sourceforge.net/projects/nsis/ and seems contain some fixes for 64-build)
4)Fixed
5)Fixed
6)Fixed
7)Not Fixed (#1104)

plugins/apps/eiskaltdcpp.mk
```diff
@@ -0,0 +1,80 @@
...
+# See index.html for further information.
+
+PKG             := eiskaltdcpp
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 4f5ebdea7a53f5d2492f391add9c1848ba377919
+$(PKG)_CHECKSUM := ebada8be7c280e3433d69f132e6a46d47da0fe31a1b201fe397b8b18b6304e73
+$(PKG)_SUBDIR   := $(PKG)-$($(PKG)_VERSION)
+$(PKG)_FILE     := $(PKG)-$($(PKG)_VERSION).tar.gz
+$(PKG)_URL      := https://github.com/$(PKG)/$(PKG)/archive/$($(PKG)_VERSION).tar.gz
+$(PKG)_DEPS     := gcc aspell boost jsoncpp libidn lua miniupnpc nsis qt5
```
tehnick: I think full `qt5` target requires too much build time and disk space.
Could you specify only really used `qt` modules?

tehnick:
3) Probably fixed: I have just successfully built `nsis-2.50` from master branch of MXE for `x86_64-w64-mingw32.static` target with `gcc-4.9.3` and with `gcc-6.1.0` (see [gcc6 plugin](https://github.com/mxe/mxe/tree/master/plugins/gcc6) for more details). Recheck this please.

tehnick:
3) https://github.com/mxe/mxe/pull/1340

pavelvat:
>Could you specify only really used qt modules?

Done.

Issue #1170, opened 2016-01-11
libieee1284 fails to build in Wheezy


starius:
@s09bQ5, libieee1284 fails to build in Debian Wheezy.
Build log: https://gist.github.com/a9bf3d7ab554614d1dcc
Commit 77a96085cf90110b8589cae50ab0f9ad93b6ba54

s09bQ5:
The build fails because patch < 2.7 (Wheezy has 2.6.1) can't handle the rename done in the patch.
Shortening the patch by using the rename feature of git was requested by @TimothyGu in #1131.

TimothyGu:
I believe that it's time we drop Wheezy support completely. Historically, Debian oldstable has always been a pain to support and therefore unsupported:

> Only the latest Debian stable series is supported.
> http://mxe.cc/#requirements-debian

Wheezy was the stable for a long time, and we generally have pretty good support for it. But now, the times have changed; more specifically, we have to sacrifice 577 lines' worth of readability for a 3-year-old operating system, that people who actually care about bleeding-edge software offered in MXE *probably* do not use.

starius:
I think we can fully drop Wheezy support when Jessie becomes oldstable.

tonytheodore:
Is there a backport for `patch` on Wheezy? Better still, could we name the patch in a way that triggers the patch workflow and apply it with `git am`?

starius:
> Is there a backport for patch on Wheezy?

No

> Better still, could we name the patch in a way that triggers the patch workflow and apply it with git am?

I thought about it. On the one hand, you are right, one patch format is easier for us to maintain and develop new features. On the other hand, strict requirements for patches would increase MXE entry threshold decreasing the number of contributions we get. Probably some contributors are not familiar with git patches, but they still can create patches manually using `diff` tool or other tool. If we switch to git patches, we should at least provide a converter from existing format to git patch format.

tonytheodore:
> I believe that it's time we drop Wheezy support completely.

@TimothyGu, Wheezy support is still required to build packages, especially for Travis CI which uses a version of `libc` prior to Jessie. See https://github.com/mxe/mxe/issues/1177#issuecomment-173752040.

The latest version of patch (2.7.5) is now being built for Wheezy (#1194) and I think it is safe to revert #1171.

starius:
[...]

PR #1171, opened 2016-01-11
fix libieee1284 on Wheezy and prevent patch-tool-mxe from using rename

starius:
see https://github.com/mxe/mxe/issues/1170


tonytheodore:
Thanks!

Issue #1166, opened 2016-01-10
fix packages for gcc 5.3


starius:
Following on from https://github.com/mxe/mxe/issues/964#issuecomment-170358288 , the following packages failed to compile with gcc 5.3.0:

libical: https://gist.github.com/c1dbf9714263ff4dcc8c
~~libieee1284: https://gist.github.com/b2f3d510014903930c01~~

Commit 51219d14b85178ce8b23241291b8769737973e7b
Version of gcc was changed to 5.3.0 in `plugins/gcc52/gcc52-overlay.mk`.
Build command: `make MXE_PLUGIN_DIRS=plugins/gcc52 -j 3`

tonytheodore:
libieee1284 doesn't seem related to `gcc` but `automake` version. It this the first time it's tested on wheezy?

starius:
@tonytheodore, you're right. See https://github.com/mxe/mxe/issues/1170

tonytheodore:
Could be an ordering issue, lots of warnings but no errors https://gist.github.com/tonytheodore/3bbeae762dccb1a3e659

tonytheodore:
I installed `dlfcn-win32` and the build still succeeds.

tonytheodore:
Did you update `isl` also?

starius:
I hadn't updated `isl` when reporting this issue. But I've updated it now, because I cloned you branch `gcc5`.

tonytheodore:
I built again with previous `isl` and all packages installed and can't replicate the issue.

starius:
I built it successfully with old isl too. Probably it is an ordering issue.

tonytheodore:
gcc is now at 5.4 #1541 without build failures.

PR #1172, opened 2016-01-12
gcc plugin: rename gcc52 to gcc5 and update 5.2.0 --> 5.3.0

tonytheodore:
See https://github.com/mxe/mxe/issues/964#issuecomment-170358288

A single 5 series package will be easier to maintain. If anyone wants to use a specific version, they can use this as a template and maintain it locally.

### DIFF
--- a/plugins/gcc5/gcc5-overlay.mk
+++ b/plugins/gcc5/gcc5-overlay.mk
@@ -11,15 +11,15 @@ PKG             := cloog
 $(PKG)_TARGETS  := $(MXE_TARGETS)
 
 PKG             := isl
-$(PKG)_VERSION  := 0.14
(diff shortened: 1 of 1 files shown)


starius:
See also https://github.com/mxe/mxe/issues/1166
I'll build `libical` from scratch to make sure that it is an ordering issue.

starius:
`libical` was built successfully for `i686-w64-mingw32.static`.

tonytheodore:
So is it `isl`, wheezy, or ordering?

starius:
Let me built it again, without updating `isl`.

starius:
`libical` was built successfully with old `isl` (0.14). It should be an ordering issue.

tonytheodore:
Rebased and merged: https://github.com/mxe/mxe/commit/eae41ccbfd983312b99cbf1461d6217b701a57b0

PR #1175, opened 2016-01-12
Allow qtserialport_qt4 to be compiled statically

dgburr:
The QtSerialPort library expects QT_STATIC to be defined when it is being compiled as a static library.  If is not defined then the Q_SERIALPORT_EXPORT macro will be set to one of Q_DECL_IMPORT/Q_DECL_EXPORT, resulting in linking errors when the test cases are built under Windows.

The following changeset introduced support for the QT_STATIC macro into QT:

https://gitlab.com/pteam/pteam-qtbase/commit/96166fa56abb52157387c4911efbd4e5e6beee93

This change is not included in QT 4.8.7 (appears to be 5.x only) and therefore QT_STATIC is not defined when QtSerialPort is compiled against 4.x QT versions.

The workaround used here is for the QSerialPort build system to defined QT_STATIC when it is being compiled statically.

### DIFF
--- a/src/qtserialport_qt4.mk
+++ b/src/qtserialport_qt4.mk
@@ -15,6 +15,12 @@ $(PKG)_DEPS     := gcc qt
 
 $(PKG)_UPDATE   := $(call MXE_GET_GITHUB_SHA, $($(PKG)_GH_USER)/$($(PKG)_GH_REPO), $($(PKG)_GH_TREE))
 
+define $(PKG)_BUILD
+    cd '$(1)' && '$(PREFIX)/$(TARGET)/qt/bin/qmake'
+    $(MAKE) -C '$(1)' -j '$(JOBS)'
+    $(MAKE) -C '$(1)' -j 1 install
+endef
+
 define $(PKG)_BUILD_SHARED
     cd '$(1)' && '$(PREFIX)/$(TARGET)/qt/bin/qmake'
     $(MAKE) -C '$(1)' -j '$(JOBS)'
--- a/src/qtserialport_qt4.mk
+++ b/src/qtserialport_qt4.mk
@@ -19,4 +19,6 @@ define $(PKG)_BUILD
     cd '$(1)' && '$(PREFIX)/$(TARGET)/qt/bin/qmake'
     $(MAKE) -C '$(1)' -j '$(JOBS)'
     $(MAKE) -C '$(1)' -j 1 install
+    # install one of the test programs
+    cp -f '$(1)/examples/serialport/cenumerator/debug/cenumerator.exe' '$(PREFIX)/$(TARGET)/bin/test-qtserialport_qt4.exe'
 endef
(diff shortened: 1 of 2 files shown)


src/qtserialport_qt4.mk
```diff
@@ -15,6 +15,12 @@ $(PKG)_DEPS     := gcc qt
 
 $(PKG)_UPDATE   := $(call MXE_GET_GITHUB_SHA, $($(PKG)_GH_USER)/$($(PKG)_GH_REPO), $($(PKG)_GH_TREE))
 
+define $(PKG)_BUILD
+    cd '$(1)' && '$(PREFIX)/$(TARGET)/qt/bin/qmake'
+    $(MAKE) -C '$(1)' -j '$(JOBS)'
+    $(MAKE) -C '$(1)' -j 1 install
+endef
+
 define $(PKG)_BUILD_SHARED
```
tonytheodore: Just remove `_SHARED` from this line - the rules are identical.

tonytheodore:
Do you have a small program to test that the linking works?

dgburr:
The qtserialport library includes numerous test programs which are built when the package is compiled.  The test program which was previously failing to link is called "cenumerator".

dgburr:
I have created a pull request to merge the fix upstream, see qtproject/qtserialport#2

If this change gets merged then $(PKG)_VERSION can be bumped in qtserialport_qt4.mk and qtserialport_qt4-1-static.patch can be removed.

starius:
> The qtserialport library includes numerous test programs which are built when the package is compiled. The test program which was previously failing to link is called "cenumerator".

Can you copy this test program to `$(PREFIX)/$(TARGET)/bin/test-qtserialport_qt4.exe`, please?

src/qtserialport_qt4.mk
```diff
@@ -15,8 +15,10 @@ $(PKG)_DEPS     := gcc qt
 
 $(PKG)_UPDATE   := $(call MXE_GET_GITHUB_SHA, $($(PKG)_GH_USER)/$($(PKG)_GH_REPO), $($(PKG)_GH_TREE))
 
-define $(PKG)_BUILD_SHARED
+define $(PKG)_BUILD
     cd '$(1)' && '$(PREFIX)/$(TARGET)/qt/bin/qmake'
     $(MAKE) -C '$(1)' -j '$(JOBS)'
     $(MAKE) -C '$(1)' -j 1 install
+    # install one of the test programs
+    cp -f '$(1)/examples/serialport/cenumerator/debug/cenumerator.exe' '$(PREFIX)/$(TARGET)/bin/test-qtserialport_qt4.exe'
```
tonytheodore: `s/debug/release/`

dgburr:
Good catch! I have been working in a branch which builds both release and debug versions....

tonytheodore:
Thanks!

PR #1176, opened 2016-01-13
add new package quazip

zgyarmati:
Tested with 'i686-w64-mingw32.static' and 'x86_64-w64-mingw32.shared' triplets


starius:
Tested with all 4 targets on Debian Jessie. The build succeeded.

Problems:

  * it installs DLLs to `usr/$(TARGET)/lib`. Example: `usr/i686-w64-mingw32.shared/lib/quazip.dll`. It should install DLLs to `usr/$(TARGET)/bin`.
  * Please install .pc file to `usr/<TARGET>/lib/pkgconfig/`.
  * Provide a sample source file and build rule for it (using the .pc file) to test compilation and linking. Install the executable to `usr/$(TARGET)/bin/test-quazip.exe`.

zgyarmati:
Dear starius, 

thanks for checking this.

* I patched the quazip/quazip.pro file to install the .dll-s into bin if we do a shared build

* It turned out that it's not that straightforward to get qmake generate the proper pkgconfig file, and i also hit the Qt bug 30898 [1], so i had to add an additional qtbase patch to fix qmake. This or similar patch is also added other cross-build environments [2], so i think it might be useful here also for other packages. If there is a better way to do the .pc file generation for a Qt lib in MXE, please let me now, so for this was the best i was able to come up.

* i also added a simple test source file and compile it using the pkgconfig flags

If anything else is needed to get this merged, please let me know. 

[1] https://bugreports.qt.io/browse/QTBUG-30898
[2] https://build.opensuse.org/package/view_file/windows:mingw:win32/mingw32-libqt5-qtbase/qt5-prevent-debug-library-names-in-pkgconfig-files.patch

starius:
@zgyarmati, thank you for your updates. Why have you added recent commits from MXE to this pull request? E.g., 4b53d9c131613cc14bc49ef3435f592e25d00329. They are unrelated to `quazip`. Can you rebase your pull request to keep only needed commits?

zgyarmati:
by mistake... Fixed now and i've also squashed the last 2 commits, as the second one was only adding the license headers.
Sorry for the noise

versions.json
```diff
@@ -318,6 +318,7 @@
     "qtwinextras": "5.5.1",
     "qtxlsxwriter": "ad90b6a2c21b300138ceb9fe9030a5917230f92d",
     "qtxmlpatterns": "5.5.1",
+    "quazip": "0.7.1",
```
starius: You should not update file ` versions.json`. It is done automatically by @MXEBot.

src/quazip.mk
```diff
@@ -0,0 +1,31 @@
...
+
+
+define $(PKG)_BUILD
+    cd '$(1)' && '$(PREFIX)/$(TARGET)/qt5/bin/qmake' PREFIX=$(PREFIX)/$(TARGET) $(if $(BUILD_STATIC), CONFIG\+=staticlib)
+    $(MAKE) -C '$(1)' -j '$(JOBS)'
+    $(MAKE) -C '$(1)' -j 1 install
+
+    '$(TARGET)-g++' \
+        -W -Wall -Werror -std=c++0x -pedantic \
+        $(if $(BUILD_STATIC), -D QUAZIP_STATIC) \
```
starius: Can this define be moved to the quazip header?

src/quazip.mk
```diff
@@ -0,0 +1,31 @@
...
+endef
+
+
+define $(PKG)_BUILD
+    cd '$(1)' && '$(PREFIX)/$(TARGET)/qt5/bin/qmake' PREFIX=$(PREFIX)/$(TARGET) $(if $(BUILD_STATIC), CONFIG\+=staticlib)
+    $(MAKE) -C '$(1)' -j '$(JOBS)'
+    $(MAKE) -C '$(1)' -j 1 install
+
+    '$(TARGET)-g++' \
+        -W -Wall -Werror -std=c++0x -pedantic \
```
starius: Do you need `-std=c++0x`?
If you need C++11, use `-std=c++11` instead.

src/quazip.mk
```diff
@@ -0,0 +1,31 @@
...
+
+define $(PKG)_UPDATE
+    $(WGET) -q -O- 'http://sourceforge.net/projects/quazip/files/quazip/' | \
+    $(SED) -n 's,.*/\([0-9][^"]*\)/".*,\1,p' | \
+    head -1
+endef
+
+
+define $(PKG)_BUILD
+    cd '$(1)' && '$(PREFIX)/$(TARGET)/qt5/bin/qmake' PREFIX=$(PREFIX)/$(TARGET) $(if $(BUILD_STATIC), CONFIG\+=staticlib)
```
starius: IMHO,
```
cd '$(1)' && '$(PREFIX)/$(TARGET)/qt5/bin/qmake' \
    $(if $(BUILD_STATIC), CONFIG\+=staticlib) \
    PREFIX=$(PREFIX)/$(TARGET)
```

src/quazip.mk
zgyarmati: Well, the quazip README [1] proposes to define this macro on the application side in case of static building (at least in my interpretation), and if it's not strictly necessary, i would avoid further patching of the quazip source tree. 

 What if we add this flag to the quazip.pc file for static targets, so at the application-side no need to explictly care about this?

[1]https://sourceforge.net/p/quazip/code/HEAD/tree/trunk/quazip/README.txt  line 41

src/quazip.mk
zgyarmati: The line was straightly copied from the qtbase.mk, without setting the c++ standard (so falling back to c++ 98, the code won't compile, so i'm going to change to -std=c++11

src/quazip.mk
zgyarmati: Good point, thx

src/quazip.mk
tonytheodore: > What if we add this flag to the quazip.pc file

That is the right place for it - `Cflags.private`

starius: > What if we add this flag to the quazip.pc file

+1 for this

starius:
> i also hit the Qt bug 30898 [1], so i had to add an additional qtbase patch to fix qmake

Do we build Debug version of Qt?

/cc @mabrand (author of QTBUG-30898)

starius:
1. Double empty line after `$(PKG)_UPDATE`
2. I've built commit fd5677bc8b85b66dd546a70b82fad6a3fecc905a for all 4 targets. Files:

  * i686-w64-mingw32.shared/bin/libquazip.a
  * i686-w64-mingw32.shared/bin/quazip.dll
  * i686-w64-mingw32.shared/bin/test-quazip-pkgconfig.exe
  * i686-w64-mingw32.shared/include/quazip/
  * i686-w64-mingw32.shared/installed/quazip
  * i686-w64-mingw32.shared/lib/pkgconfig/quazip.pc
  * i686-w64-mingw32.static/bin/test-quazip-pkgconfig.exe
  * i686-w64-mingw32.static/include/quazip/
  * i686-w64-mingw32.static/installed/quazip
  * i686-w64-mingw32.static/lib/libquazip.a
  * i686-w64-mingw32.static/lib/pkgconfig/quazip.pc
  * x86_64-w64-mingw32.shared/bin/libquazip.a
  * x86_64-w64-mingw32.shared/bin/quazip.dll
  * x86_64-w64-mingw32.shared/bin/test-quazip-pkgconfig.exe
  * x86_64-w64-mingw32.shared/include/quazip/
  * x86_64-w64-mingw32.shared/installed/quazip
  * x86_64-w64-mingw32.shared/lib/pkgconfig/quazip.pc
  * x86_64-w64-mingw32.static/bin/test-quazip-pkgconfig.exe
  * x86_64-w64-mingw32.static/include/quazip/
  * x86_64-w64-mingw32.static/installed/quazip
  * x86_64-w64-mingw32.static/lib/libquazip.a
  * x86_64-w64-mingw32.static/lib/pkgconfig/quazip.pc

Dump of .pc files: https://gist.github.com/starius/b4cfe68f6f67770d8c68

Error: it installs `libquazip.a` to `bin/` for shared targets. Please do not install static library for shared targets.
Issue: pc file doesn't declare Qt5 requirements. If you add Qt5 to `Requires`, you need not pass Qt libraries to `Libs.private`, as `pkgconfig` adds flags of requirements automatically. See https://secure.freedesktop.org/%7Edbn/pkg-config-guide.html#using

zgyarmati:
Thanks for guiding me on this. 
I added the Qt5Core as Required, but oddly enough, i couldn't find a way to get qmake to build only the .dll, even though i set CONFIG += dll, it still builds the static lib as well. Can you suggest any solution with qmake, or it's enough if i just simply delete the libquazip.a file from the quazip.mk file, after calling `make install`?

zgyarmati:
It seems this PR messed up enough, i'm closing this, clean up my patches and file a new one.

PR #1179, opened 2016-01-13
add gnu patch to native plugins

tonytheodore:
See #1170

### DIFF
--- a/plugins/native/patch.mk
+++ b/plugins/native/patch.mk
@@ -1,0 +1,30 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := patch
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 2.7.5
+$(PKG)_CHECKSUM := 7436f5a19f93c3ca83153ce9c5cbe4847e97c5d956e57a220121e741f6e7968f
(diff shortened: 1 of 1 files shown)


plugins/native/patch.mk
```diff
@@ -0,0 +1,30 @@
...
+    $(WGET) -q -O- 'http://ftp.gnu.org/gnu/patch/?C=M;O=D' | \
+    $(SED) -n 's,.*<a href="patch-\([0-9][^"]*\)\.tar.*,\1,p' | \
+    $(SORT) -V | \
+    tail -1
+endef
+
+define $(PKG)_BUILD_$(BUILD)
+    mkdir '$(1).build'
+    cd    '$(1).build' && '$(1)/configure' \
+        --prefix='$(PREFIX)/$(TARGET)'
```
starius: Why `$(PREFIX)/$(TARGET)`, not `$(PREFIX)/$(BUILD)`?

tonytheodore: No real reason, less change in case we ever want to build for a different target? Could use `$(MXE_CONFIGURE_OPTS)` here and it would have the same effect.

starius: I see. Let it be `$(PREFIX)/$(TARGET)` then.

PR #1182, opened 2016-01-13
Update luarocks to 2.3.0, unify patch and add a check to the wrapper

starius:
Tested with commands in `plugins/luarocks/README.md`.


tonytheodore:
Thanks!

Issue #1185, opened 2016-01-16
Build fail with CMake + Qt (QUiLoader)


stloeffler:
When trying to compile the following minimal example
CMakeLists.txt:
```CMake
cmake_minimum_required(VERSION 3.0)
find_package(Qt5 REQUIRED COMPONENTS Core Widgets Gui UiTools)

include_directories(${Qt5UiTools_INCLUDE_DIRS})
add_executable(mxe-cmake-qtuitools mxe-cmake-qtuitools.cpp)
# NB: Qt5Widgets, Qt5Gui, Qt5Core must be given separately; otherwise they are
# appended automatically at the end, making it impossible to satisfy all the
# other implicit dependencies (opengl, harfbuzz, etc.)
target_link_libraries(mxe-cmake-qtuitools ${Qt5UiTools_LIBRARIES} ${Qt5Widgets_LIBRARIES} ${Qt5Gui_LIBRARIES} ${Qt5Core_LIBRARIES} opengl32 harfbuzz jpeg png glib-2.0 winmm ws2_32 intl iconv pcre16 z)
```

mxe-cmake-qtuitools.cpp:
```C++
#include <QUiLoader>

int main() {
	QUiLoader l;
	return 0;
}
```

(which uses QUiLoader, which is part of Qt5UiTools) I get the following error message:
```CMakeFiles/mxe-cmake-qtuitools.dir/build.make:89: *** target pattern contains no '%'.  Stop.```
build.make:89 reads
```mxe-cmake-qtuitools.exe: Qt5::UiPlugin-NOTFOUND```
although Qt5::UiPlugin is built and CMake doesn't complain while configuring.

This seems to be similar to Alexpux/MINGW-packages#762 which apparently has been resolved by patching Qt with https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-qt5-static/0038-qt-5.5.0-cmake-Rearrange-STATIC-vs-INTERFACE-targets.patch

starius:
It is reproducible here. I'll try to rebuild Qt5 with this patch saved to file `src/qtbase-3-make-Rearrange-STATIC-vs-INTERFACE-targets.patch`

starius:
It works with this patch.

starius:
I'll open a pull request adding the patch and the test.

PR #1190, opened 2016-01-17
Qt5 cmake rearrange targets

starius:
See https://github.com/mxe/mxe/issues/1185

### DIFF
--- a/src/qtbase-1-fixes.patch
+++ b/src/qtbase-1-fixes.patch
@@ -167,3 +167,36 @@ index 1111111..2222222 100755
  
  #setup the build parts
  if [ -z "$CFG_BUILD_PARTS" ]; then
+
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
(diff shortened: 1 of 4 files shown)


starius:
Currently the test doesn't work. I'm changing it to standard cmake && make approach.

starius:
`i686-w64-mingw32.static` target tested.

starius:
Tested on all 4 targets.

UniqueFool:
thanks for working on this

tonytheodore:
Thanks @starius!

@mabrand, the new patch targets (https://github.com/mxe/mxe/blob/master/patch.mk) normalise the format of patches to improve `git am` round-tripping (#983) and readability of updates. I don't think this will impact your existing workflow, just when it comes to exporting the patches (assuming a local repo):

```
make export-patch-qtbase GITS_DIR=/path/above/repo qtbase_SUBDIR=qtbase-5...
```

there will be variables for local repos in the future, but the logic is currently similar to the original `patch-tool-mxe`.

PR #1195, opened 2016-01-22
new package: liblsmash

darealshinji:
L-SMASH is a library used for handling MP4 files. Example usage: if x264 was build with L-SMASH support, it's able to save its output as mp4 videos.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1762,6 +1762,10 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
         <td class="website"><a href="http://liblqr.wikidot.com/">liblqr-1</a></td>
     </tr>
     <tr>
+        <td class="package">liblsmash</td>
+        <td class="website"><a href="https://l-smash.github.io/l-smash/">L-SMASH</a></td>
+    </tr>
+    <tr>
         <td class="package">libltdl</td>
         <td class="website"><a href="http://www.gnu.org/software/libtool/manual/html_node/Using-libltdl.html#Using-libltdl">GNU Libtool Library (libltdl)</a></td>
     </tr>
--- a/src/liblsmash-1.patch
+++ b/src/liblsmash-1.patch
@@ -1,0 +1,13 @@
+This file is part of MXE.
+See index.html for further information.
+
+--- a/configure
++++ b/configure
+@@ -198,6 +198,7 @@
+ case "$TARGET_OS" in
+     *mingw*)
+         EXT=".exe"
++        SHARED_NAME="liblsmash-$MAJVER"
+         SHARED_EXT=".dll"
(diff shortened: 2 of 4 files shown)


starius:
1.
We have package [x264](https://github.com/mxe/mxe/blob/master/src/x264.mk). Can it take advantage of L-SMASH? If yes, please add L-SMASH to dependencies of x264 and adjust build rules of x264, if it is needed.

2.
> `--prefix='$(PREFIX)/$(TARGET)' \`
> `--cross-prefix=$(TARGET)- \`
> `$(if $(BUILD_SHARED), --enable-shared --disable-static)`

Please replace these lines with macro `$(MXE_CONFIGURE_OPTS)`.

starius:
Provide updater for the package. Use macro `MXE_GET_GITHUB_TAGS`.

darealshinji:
> Please replace these lines with macro $(MXE_CONFIGURE_OPTS).

Unfortunately the configure script doesn't know what `--host=` means and will exit with an error: https://github.com/l-smash/l-smash/blob/master/configure#L167
Should I patch that part of the configure script?

>We have package x264. Can it take advantage of L-SMASH? If yes, please add L-SMASH to dependencies of x264 and adjust build rules of x264, if it is needed.

Like I said it enables mp4 output. x264's configure script will automatically enable support for L-SMASH if it's found. So yes, I could add it to the build dependencies.

>Provide updater for the package. Use macro MXE_GET_GITHUB_TAGS.

`$(call MXE_GET_GITHUB_TAGS, l-smash/l-smash)` <- Is that the correct usage?

starius:
> Should I patch that part of the configure script?

I think, no. Instead add a comment that `configure` fails with `--host`.

> So yes, I could add it to the build dependencies.

You are welcome.

> `$(call MXE_GET_GITHUB_TAGS, l-smash/l-smash)`

The following code works for me:

```
define $(PKG)_UPDATE
    $(call MXE_GET_GITHUB_TAGS, l-smash/l-smash, v)
endef
```

Note second argument, "v". You can test the updater by changing version to some version from past (e.g., 2.9.0) and invoking `make update-package-liblsmash`. It should change the version to current value.

starius:
Commit d1e5a42c81cd033ad9b585dbdca317e9e3dd81c9 was built successfully for all 4 targets, package liblsmash.

darealshinji:
Apparently liblsmash is used by the CLI but not by the library itself. So I don't know if it makes sense to add it as a dependency to the Makefile.

TimothyGu:
We build the CLI too so yes, please add the dependency.

tonytheodore:
Thanks! Rebased and merged: https://github.com/mxe/mxe/compare/9f9c17c...1f7a3cd

PR #1196, opened 2016-01-23
Added package gst-plugins bad

Roflincopter:
I'm not actively using the bad plugins in my projects, but am using the ugly and good plugins, felt bad leaving the bad plugins out of this. Adding gst-plugins-ugly in a separate pull request.


tonytheodore:
Thanks!

Rebased and merged: https://github.com/mxe/mxe/compare/1da63ba...f7e7fe0

PR #1197, opened 2016-01-23
added package gst-plugins-ugly

Roflincopter:
From the gst-plugins-ugly I actually only use the lameenc plugin, but I also added all the optional deps that mxe had and that the plugins support so that as many of the plugins are being built.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1434,6 +1434,10 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
         <td class="website"><a href="http://gstreamer.freedesktop.org/">gst-plugins-good</a></td>
     </tr>
     <tr>
+        <td class="package">gst-plugins-ugly</td>
+        <td class="website"><a href="http://gstreamer.freedesktop.org/">gst-plugins-ugly</a></td>
+    </tr>
+    <tr>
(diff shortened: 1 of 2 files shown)


tonytheodore:
Thanks, both this and #1196 look good. I have a vague recollection of removing these in the past based on licensing issues (I'll have to search the archives). Do you know the current licence state?

Roflincopter:
Well, what is mxe's standing on this issue? Gstreamer aims to be LGPL itself. Even when linking to GPL code [licensing](http://gstreamer.freedesktop.org/documentation/licensing.html). So I guess if all the libraries (that these plugins depend on) are in MXE and we're not having licensing issues with those. That gstreamer / plugins that are themself LGPL, or maybe the older ones might be MIT if i read correctly, should not poison MXE in any way.

tonytheodore:
I see, plugins may possibly contain code for libraries with license/patent issues but we don't have any of those libs in MXE so there's no problem.

tonytheodore:
Build succeeds on all four targets - thanks!

Rebased and merged: https://github.com/mxe/mxe/compare/1da63ba...f7e7fe0

PR #1203, opened 2016-01-25
build-pkg: set MXE_DIR automatically

starius:
Produce a warning if MXE_DIR != /usr/lib/mxe

When making a debug build, it is better to get a warning than set MXE_DIR manually each time.


tonytheodore:
Thanks!

PR #1208, opened 2016-01-31
index.html: add "How to choose MXE target"

starius:
See http://lists.nongnu.org/archive/html/mingw-cross-env-list/2016-01/msg00022.html


starius:
Rebased.

tonytheodore:
Thanks!

PR #1209, opened 2016-02-01
binutils: enable deterministic archives

tonytheodore:
Since #1206 is going to cause a full rebuild anyway.

see: https://wiki.debian.org/ReproducibleBuilds/TimestampsInStaticLibraries


starius:
Great! :+1:

PR #1213, opened 2016-02-03
add OpenCascade 6.9.0 as a plugin

fiesh:
Alas the test doesn't seem to work with plugins?

There's a big diff file that consists of hundreds of fixes I had to make :(

There already is OCE in MXE, however it's not the current version of OCE, and the current version of OCE isn't up to date with OpenCascade either.


tonytheodore:
I'd suggest creating a [mirror](https://github.com/mirror/mirror) of the [public git repo](http://git.dev.opencascade.org/gitweb/?p=occt.git) and basing downloads/updates on that. There's now a version 7 and we want to start with the most recent - especially if heavy patching makes it harder to update down the track.

Otherwise, a little background on OCE/OpenCascade may be helpful for those of us who aren't familair with the projects:
 - can we update OCE to the latest version
 - can the two exist side-by-side
 - why a plugin for OpenCascade
 - do we need both, which has the most momentum

fiesh:
I actually suggest getting rid of it now, I don't intend to put work into it any more.  OpenCascade has many problems I think, and we switched to something else, so best to drop it.

PR #1214, opened 2016-02-06
add package sparsehash

pavelvat:


starius:
Files installed: https://gist.github.com/1733d1204aaaa1210e83
libsparsehash.pc: https://gist.github.com/3a4fea70c66e03660e29

It seems to be a header-only library.

Could you provide a sample application using this library and install it to '$(PREFIX)/$(TARGET)/bin/test-sparsehash.exe', please?

src/sparsehash.mk
```diff
@@ -0,0 +1,23 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := sparsehash
+$(PKG)_IGNORE   :=
```
starius: Please remove a space after `:=`

starius:
Please rebase the commit. [Its parent](https://github.com/pavelvat/mxe/commit/f9d83fe44b52eefe9f2453bcdcca5feaf3d651f6) is more than one month old.
```
git pull --rebase https://github.com/mxe/mxe master
```

pavelvat:
>It seems to be a header-only library.

Yes it is.

>Could you provide a sample application using this library and install it to '$(PREFIX)/$(TARGET)/bin/test-sparsehash.exe', please?

Is it really needed? By default ```sparsehash``` build various tests (around five minutes on every TARGET), so it's guaranteed correct build (else command ```make sparsehash``` will lead to fail).

starius:
> Is it really needed? By default sparsehash build various tests (around five minutes on every TARGET), so it's guaranteed correct build (else command make sparsehash will lead to fail).

Just copy one of these tests (e.g. `hashtable_test.exe`) to `'$(PREFIX)/$(TARGET)/bin/test-sparsehash.exe'`. If sparsehash decides not to build tests for some reason, the package will fail to build (and we will fix it). Otherwise it may remain broken for a long time.

starius:
It built successfully on all 4 targets and created the following files:
https://gist.github.com/d1e91a09c79766500447

starius:
Thanks!

PR #1217, opened 2016-02-13
qjson: install pkg-config file

pavelvat:


starius:
Well done!

Can you provide a simple test for this package, build it using flags produced by `pkg-config` and install it to `$(PREFIX)/$(TARGET)/bin/test-qjson.exe`, please?

pavelvat:
Any more problems with this PR? This patch used for official builds of https://github.com/clementine-player/Clementine for detecting qjson: https://github.com/clementine-player/Clementine/blob/master/CMakeLists.txt#L76

By the way, a similar problem with cryptopp. MXE has no pkg-config file for cryptopp.

PR #1225, opened 2016-02-21
copydlldeps.md: Test & example section (keepassx)

starius:
@dl5rcw, I have added a working example of using copydlldeps.sh inside MXE. Can you test it, please?


tools/copydlldeps.md
```diff
@@ -45,6 +45,24 @@ if [ ! $( echo $compiler | grep -q "shared" ) ]; then
...
+Let us compile password manager [KeePassX](https://www.keepassx.org) in
+shared mode (i686-w64-mingw32.shared) and gather all its DLL dependencies.
+
+```bash
+$ make keepassx MXE_PLUGIN_DIRS=plugins/apps \
+    MXE_TARGETS='i686-w64-mingw32.shared'
+$ ./tools/copydlldeps.sh --copy --destdir keepassx-bundle \
+    --infile usr/i686-w64-mingw32.shared/bin/KeePassX.exe \
+    --recursivesrcdir usr/i686-w64-mingw32.shared/
+$ cp -r usr/i686-w64-mingw32.shared/bin/{KeePassX.exe,share} keepassx-bundle
```
TimothyGu: Did you mean

```sh
cp -r usr/i686-w64-mingw32.shared/{bin/KeePassX.exe,share} keepassx-bundle
```

starius: No. `keepassx` puts its files to `bin/share`.

starius:
Package `keepassx` installs the following files and directories for target `i686-w64-mingw32.shared`:

  * i686-w64-mingw32.shared/bin/KeePassX.exe
  * i686-w64-mingw32.shared/bin/QtCore4.dll
  * i686-w64-mingw32.shared/bin/QtGui4.dll
  * i686-w64-mingw32.shared/bin/QtXml4.dll
  * i686-w64-mingw32.shared/bin/share/        
  * i686-w64-mingw32.shared/installed/keepassx

pavelvat:
@starius In my opinion ```$(PREFIX)/$(TARGET)/apps/$(PKG)``` is the best variant.
https://github.com/mxe/mxe/pull/1165/files#diff-17d9a471cc45b22771072c4634dc97d9R20

starius:
@pavelvat, agreed with `$(PREFIX)/$(TARGET)/apps/$(PKG)`. Given files of different applications scattered over `bin`, `bin/share` and `share`, it is not easy to collect all files of a single application.

> https://github.com/mxe/mxe/pull/1165/files#diff-17d9a471cc45b22771072c4634dc97d9R20

Why do you have second `$(PKG)` in `apps/$(PKG)/$(PKG)/`?

pavelvat:
>Why do you have second $(PKG) in apps/$(PKG)/$(PKG)/?

Due to '$(PREFIX)/$(TARGET)/apps/$(PKG)/nsis':
https://github.com/mxe/mxe/pull/1165/files#diff-17d9a471cc45b22771072c4634dc97d9R69

tools/copydlldeps.md
dl5rcw: @starius - could not get keepassx started from git - Unknown CMake command "get_git_head_revision". Get me started and I will test.
Review: every qt application seems to need 
```bash
--enforce usr/i686-w64-mingw32.shared/qt5/plugins/platforms/
```  
as an argument for the application to start. In my new version provided, it would then be
```bash
--enforcedir usr/i686-w64-mingw32.shared/qt5/plugins/platforms/
```

TimothyGu:
I prefer `$target/share` over `apps/`. The former is a well-established idiom, while to enforce the latter one would have to change pretty much every single package.

TimothyGu:
Also…

> Given files of different applications scattered over bin, bin/share and share, it is not easy to collect all files of a single application.

I have not **ever** seen any projects that put their architecture-independent files in `bin/share`. And by introducing a new standard `apps/$pkg`, we are falling into the [xkcd standards fallacy](https://xkcd.com/927/).

tools/copydlldeps.md
starius: I mean follow commands in tools/copydlldeps.md which are being added by this PR.

E.g., to build keepassx, `make keepassx MXE_PLUGIN_DIRS=plugins/apps MXE_TARGETS='i686-w64-mingw32.shared'`. Building keepassx manually in MXE is hard and unneeded work.

> Review: every qt application seems to need
> --enforce usr/i686-w64-mingw32.shared/qt5/plugins/platforms/

I hope keepassx doesn't need it, as it uses Qt4, not Qt5.

starius:
@TimothyGu, `bin/share` seems to be an error of current keepassx installation. It should be either `apps/keepassx/share` or `share`. I don't know how to tell keepassx to look for images in `share` instead of `<path/to/keepassx.exe>/share`.

The idea of putting files to common directories (like `share`) is based on reusing files in other targets. E..g, reusing libraries is easy when all library files are located in a single directory, where configure scripts look for them.

However applications are a special case, as they are terminal targets of build graph. Applications are not used by other targets (as libraries do). Instead, they are copied from MXE tree to Windows machines and used there. It is easier to do, when an application has its own directory. Scripts like copydlldeps can be used to find all DLLs used by an executable, but images and other files in `share` can be found only manually.

> while to enforce the latter one would have to change pretty much every single package.

It should be done only for applications. We currently have 3 applications in "apps" plugin and I have done it for all of them in https://github.com/mxe/mxe/issues/1229

TimothyGu:
> It should be done only for applications.

Then it sounds alright to me. Thanks for the clarification. But …

> The idea of putting files to common directories (like `share`) is based on reusing files in other targets.

True, `share` is designed for file sharing, but in MXE, only `usr/share` is reused, so the argument is not applicable to `usr/$target/share`

BTW, it is important to note that when copying files, you most likely want to copy `usr/$target/share` as well for e.g. fontconfig a/o openssl (IIRC). So it is an existing issue that installing to `apps/` cannot fully resolve.

starius:
> BTW, it is important to note that when copying files, you most likely want to copy usr/$target/share as well for e.g. fontconfig a/o openssl (IIRC). So it is an existing issue that installing to apps/ cannot fully resolve.

I see. Maybe we should copy (or symlink) all dependencies to `apps/<pkg>`? keepassx already copies some DLLs.

tools/copydlldeps.md
dl5rcw: @starius - I have had an idea of how to easily add a neat script to test and teach people of how to use copydlldeps.sh
https://github.com/dl5rcw/mxe/blob/master/tools/copydlldepsexample.sh

Basically, it makes use of Qt5 itself. Qt has examples already delivered to the project. I found them stored in pkg directory of my working mxe environment.

It grabs the qtbase, copies to a tmp file that it created before, runs everything needed to build the package and delivers all dependencies to a OUTDIR directory. All you may want to modify is MXEDIR and OUTDIR in the beginning.
Have a look in the script and you may decided to either add it to the copydlldeps.md file, or put it to the tools directory. I think it can teach people how to use mxe, qt5 and deliver a package with all dependencies. I copied cube from my OUTDIR of the server to my desktop and it started right away. 

You may decide if you want to edit the lines to copydlldeps.md in a Test and Example section, or if you want to provide the bash script pure in tools.

starius: I think, `copydlldepsexample.sh` is too long to serve as an example. Example should fit in my small 24x80 terminal :). Moreover it is not only an example of using `copydlldeps.sh`, but also of building and deploying. 

Nevertheless it would be useful as an introduction for newcomers, but not in a form of a shell script doing everything (and including "Do NOT edit below this line"), but in form of article with description of each step supplemented with commands. Try to avoid using variables in examples to make it easier to read by humans. Please also rename COMPILER to TARGET, because it is called TARGET in the rest of MXE code. I think, it can be added to `index.html`.

dl5rcw: It tossed some lines and exchanged variables to text. I wrote it as a test routine and modified to contribute. I personally like variables as I don't have to go through the complete code when adapting. :)

You may add it to index.html, modify as you like. I think it fits nice there, too.

pavelvat:
https://github.com/mxe/mxe/pull/1225#issuecomment-186830569
>agreed with $(PREFIX)/$(TARGET)/apps/$(PKG). 

Maybe it would be better to make a separate top-level directory, for example, mxe/apps/:
```
mxe/apps/i686-w64-mingw32.static
mxe/apps/i686-w64-mingw32.shared
mxe/apps/x86_64-w64-mingw32.static
mxe/apps/x86_64-w64-mingw32.shared
```
because option for recursive search ```-R $(PREFIX)/$(TARGET)/``` for ./copydlldeps.sh lead to copying dll files from directories ```$(PREFIX)/$(TARGET)/apps/$(PKG)/```, which may contain older versions of dll files.

https://github.com/mxe/mxe/pull/1225#issuecomment-186980328
>Maybe we should copy (or symlink) all dependencies to apps/<pkg>?

symlinks obviously unacceptable.
The main idea: to provide a fully ready to use on OS Windows builds of applications. For example, simply copy the directory $(PREFIX)/$(TARGET)/apps/clementine/ to a flash drive or a shared folder and use the software immediately.

starius:
> option for recursive search -R $(PREFIX)/$(TARGET)/ for ./copydlldeps.sh lead to copying dll files from directories $(PREFIX)/$(TARGET)/apps/$(PKG)/, which may contain older versions of dll files.

I prefer to have one directory `$(TARGET)`, not two (`$(PREFIX)/$(TARGET)` and `mxe/apps/$(TARGET)`). Implementation details of tools should not affect file layout. Maybe it is the time to reimplement `copydlldeps.sh` in other language.

tonytheodore:
> Maybe it is the time to reimplement copydlldeps.sh in other language.

Or break it into smaller functions and use `find` and `cp` in a less "opinionated" way. I stumbled across this [ldd for PE](https://github.com/gsauthof/pe-util) utility and usage is along the lines of:

```
peldd ... | xargs cp ...
```

pavelvat:
https://github.com/mxe/mxe/pull/1225#issuecomment-202808064
```
because option for recursive search -R $(PREFIX)/$(TARGET)/ for ./copydlldeps.sh lead to copying dll 

files from directories $(PREFIX)/$(TARGET)/apps/$(PKG)/, which may contain older versions of dll files.
```

It was fixed by using option -X for copydlldeps.sh:
https://github.com/mxe/mxe/blob/master/plugins/apps/clementine_qt4.mk#L84

PR #1229, opened 2016-02-21
"apps" plugin: Install files to PREFIX/TARGET/apps

starius:
See https://github.com/mxe/mxe/pull/1225#issuecomment-186829585
See https://github.com/mxe/mxe/pull/1165/files#diff-17d9a471cc45b22771072c4634dc97d9R20


pavelvat:
Seems it ready to merge.

PR #1231, opened 2016-02-22
build-matrix: replace unsupported char for "X"

starius:
Character used for "X" is not supported by Tor Browser.
The page looks like this:

build matrix page in Tor Browser

Tor Browser version: 5.5.2 (based on Mozilla Firefox 38.6.1).

Tor Browser doesn't support `&#x2717;` (&#x2717;).
Fortunately `&#x2717;` is similar to `&#215;` (&#215;) which is supported:

fixed build matrix page in Tor Browser

It was also reported to Tor Browser team:
https://trac.torproject.org/projects/tor/ticket/18364


TimothyGu:
Ugh…

Patch looks okay to me, but I'd like to wait for Tor Browser Team's reply before merging, in case it was us who did something wrong.

PR #1234, opened 2016-02-23
fix dependencies of build-matrix.html

starius:
Dependencies of build-matrix.html were written without regarding plugins. It resulted in the following error on Debian Wheezy (which uses plugin "plugins/native/wheezy/"):

```
$ make build-matrix.html
make: *** No rule to make target `src/autoconf.mk',
needed by `build-matrix.html'.  Stop.
```

The following targets also need updating:
  * cleanup-style
  * cleanup-deps-style
  * versions.json


tonytheodore:
Thanks!

PR #1236, opened 2016-02-24
qtbase: update patch for fixing `ar' error:

tehnick:
`u' modifier ignored since `D' is the default (see `U')
This error is shown during builds of end-user programs.

### DIFF
--- a/src/qtbase-1-fixes.patch
+++ b/src/qtbase-1-fixes.patch
@@ -200,3 +200,22 @@ index 1111111..2222222 100644
      add_library(Qt5::$${CMAKE_MODULE_NAME} SHARED IMPORTED)
  !!ENDIF
  !!ENDIF
+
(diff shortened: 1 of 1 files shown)


starius:
It is a warning, not an error, doesn't it?

tehnick:
Yes, with default qmake flags projects are buildable. But in console output this message is called "error" and QtCreator shows it as error.

starius:
This PR proposes a well-known work-around: https://bugzilla.redhat.com/show_bug.cgi?id=1155273#c1
I think, we should use it.

Thanks!

tehnick:
Yep. Other build systems are affected as well.

I'll try to send PR to Qt project or will open bug report. Hope that in future this patch will not be necessary.

tehnick:
https://codereview.qt-project.org/#/c/150304/

tehnick:
http://code.qt.io/cgit/qt/qtbase.git/commit/?id=a17a7d3

PR #1240, opened 2016-02-28
libdca: fix broken symlink

### LINKED ISSUE #1239: libdca: broken symlink lib/libdts.a on shared targets

starius:
fix #1239


TimothyGu:
I wouldn't object to removing libdca. It's old, unmaintained, and unused by any projects in MXE. Much better alternatives like FFmpeg and libdcadec exist.

PR #1243, opened 2016-03-02
build-pkg: build a package second time in tree with all packages

starius:
This pull request implements second and third passes of #1111. In original proposal, the second pass builds a package in a tree with all packages except itself and its followers and the third pass rebuilds a package in a tree with all packages built. The third pass seems to be more strict than second, so it was implemented here.

Build log: https://gist.github.com/starius/a165cbb474f4d9c1dd7e
(It is based on commit 62467d9d4dbb69aa632ae684b4a5a3f4c3b4a1ff.)

Package `fontconfig` breaks on the seconds pass.

### DIFF
--- a/tools/build-pkg.lua
+++ b/tools/build-pkg.lua
@@ -430,7 +430,7 @@ local function gitCheckout(new_branch, deps, item2index, pass_of_deps)
 end
 
 local function gitAdd()
-    os.execute(GIT .. 'add .')
+    os.execute(GIT .. 'add --all .')
 end
 
 -- return two lists of filepaths under ./usr/
@@ -442,6 +442,7 @@ local function gitStatus()
     local git_st = io.popen(GIT .. 'status --porcelain', 'r')
     for line in git_st:lines() do
         local status, file = line:match('(..) (.*)')
+        assert(status:sub(2, 2) == ' ')
         status = trim(status)
         if file:sub(1, 1) == '"' then
             -- filename with a space is quoted by git
--- a/tools/build-pkg.lua
+++ b/tools/build-pkg.lua
@@ -602,13 +602,15 @@ local function prepareTree(pass, item, item2deps, prev_files, item2index)
             item2index,
             'first'
         )
-        -- Remove files of item from previous build.
-        for _, file in ipairs(prev_files) do
-            os.remove(file)
-        end
         removeEmptyDirs()
-        gitAdd()
-        gitCommit(("Remove %s to rebuild it"):format(item, pass))
+        if prev_files then
+            -- Remove files of item from previous build.
+            for _, file in ipairs(prev_files) do
+                os.remove(file)
+            end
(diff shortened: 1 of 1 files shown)


tonytheodore:
Thanks, this is looking good! I've done some testing on OS X and it's surfacing some new ordering errors [log](https://gist.github.com/tonytheodore/253a2735ef2033fb212335c28f02dab0). It fails with a `lua` error during the second pass:

```
[134/419] The build is expected to complete in 10.8 hours, on Sun Apr 17 17:03:41 2016
lua: ./tools/build-pkg.lua:603: bad argument #1 to 'ipairs' (table expected, got nil)
stack traceback:
	[C]: in function 'ipairs'
	./tools/build-pkg.lua:603: in upvalue 'prepareTree'
	./tools/build-pkg.lua:651: in upvalue 'buildItem'
	./tools/build-pkg.lua:895: in upvalue 'buildPackages'
	./tools/build-pkg.lua:1066: in local 'main'
	./tools/build-pkg.lua:1080: in main chunk
	[C]: in ?
```

I'm not sure how to investigate further (only building one target at the moment):

```
$ lua -v
Lua 5.3.1  Copyright (C) 1994-2015 Lua.org, PUC-Rio
```

starius:
I fixed `bad argument #1 to 'ipairs'` error. It used to occur only if some package is broken on the first pass.

Rebased.

@tonytheodore, can you rerun it on OSX, please?

tonytheodore:
Thanks, rerun on OSX succeeds:
https://gist.github.com/tonytheodore/ab636c3b1078bc8f4b4179af4a8b5ac8

> The third pass seems to be more strict than second, so it was implemented here.

I can't see any reference to a third pass, did you mean "wasn't implemented here"?

starius:
>> The third pass seems to be more strict than second, so it was implemented here.

> I can't see any reference to a third pass, did you mean "wasn't implemented here"?

Sorry for this confusion.

I implemented third pass (in #1111 notation), but it is called "second" in this pull request, because #1111's "second" pass was not implemented.

According to #1111,

  * first pass is initial build of a package in tree with its dependencies,
  * second pass is build of a package in tree with all packages except packages depending on the package,
  * third pass is re-build of a package in tree with all packages except itself.

Third pass includes more packages in the tree, than second pass (namely packages depending on the package being built). That is why I said that the third pass seemed to be more strict than second.

starius:
> `$ make clean && MXE_NO_DEBS=1 MXE_DIR=`pwd` ./tools/build-pkg.lua`

By the way, `MXE_DIR` is not needed after 8ec0d4dafb6ddd244a88e787ec3a6abac470b331

tonytheodore:
Thanks for clarifying - this is a great addition to the overall QA already offered by `build-pkg`, nice work!



P.S. existing users of `build-pkg` (yourself ;) will be surprised by the default second pass.

PR #1245, opened 2016-03-04
new package 'CODA'

OkoSanto:
remark:

CODA builds a tool go generate Fortran includes from the C source. This tool must be built on the host system.  I'm not sure my way to do that is the "proper" way:

  `cd '$(1)' && $(CC) -I . -o generate-finc fortran/generate-finc.c`

Also, this package could be improved by including HDF4 and HDF5 support (configure options --with-hdf4 and --with-hdf5), but I couldn't get it to build like that.


starius:
Where is `coda-test.c` used?

OkoSanto:
It's supposed to be the test program, but I now see I forgot to add the build instructions for it in `coda.mk`. Do you prefer I do that, or should I just leave out `coda-test.c`?

starius:
Please add build instructions, install exe file to `$(PREFIX)/$(TARGET)/test-coda.exe`.

OkoSanto:
I've added the build instructions (I assume `$(PREFIX)/$(TARGET)/bin/test-coda.exe` is the correct target?).

starius:
Yes, `$(PREFIX)/$(TARGET)/bin/test-coda.exe` is the correct target. There was a typo in my [previous comment](https://github.com/mxe/mxe/pull/1245#issuecomment-192449295).

starius:
I have compiled https://github.com/OkoSanto/mxe/commit/431097d1cce3e89a56912ed4bc4789b8ec5a7e70 on 4 targets. It builds.

Installed files: https://gist.github.com/1ca60e0a84fd9028a116
Can you prevent coda from installing `*/share/coda/doc/html/*`, please?

OkoSanto:
I've added a patch to provide a `--disable-install-docs` option in the  configure script.

starius:
1. Thanks! Documentation is not installed now.
 Now it installs the following files: https://gist.github.com/e102cb0219e1628a533d

2. The following files are installed to `$(PREFIX)/$(TARGET)/bin`:

 * codacheck.exe
 * codacmp.exe
 * codadd.exe
 * codadump.exe
 * codaeval.exe
 * codafind.exe

 Do you use these files in MXE? Can we drop them as well?

OkoSanto:
I do not use these files myself, but included them for other users 
that might need them.

On the other hand, these tools are aimed at users rather than at 
developers, and there is already an official binary distribution for 
windows, so maybe I can leave them out.  What do you prefer?

Some brief descriptions taken from the CODA home page:

*codadump: With the codadump command line tool you can view data from
  any product file that is supported by CODA. The tool allows you to
  inspect the product structure (including array sizes), view and
  export data from the product in ASCII format, and export data into
  HDF4 format.

*codaeval: CODA comes with an expression language. The codaeval tool
  allows you to run such an expression on one or more product files, 
or
  you can use it as a quick way to just syntactically and 
grammatically
  check an expression.

*codafind: The codafind command line tool allows you to search files
  and directories for product files that are supported by CODA. The
  tool comes with a very powerful filtering mechanism that makes it
  possible to search for only those product files that match certain
  filter criteria.

starius:
As developers don't use these tools, I would prefer not to install them as a part of this package. If somebody needs them, he/she'll create another package in [apps](https://github.com/mxe/mxe/tree/master/plugins/apps) plugin for that.

Can you also rebase your commit, please?

```
$ git pull --rebase https://github.com/mxe/mxe/ master
```

OkoSanto:
OK, done!

starius:
Thanks!

PR #1246, opened 2016-03-04
Add HDF-EOS libraries

OkoSanto:
(no matter what the names suggest, HDF-EOS2 and HDF-EOS5 are 2 completely different libraries, one building on HDF4 and the other on HDF5)

### DIFF
--- a/src/hdf-eos2.mk
+++ b/src/hdf-eos2.mk
@@ -1,0 +1,36 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := hdf-eos2
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 19v1.00
+$(PKG)_CHECKSUM := 3fffa081466e85d2b9436d984bc44fe97bbb33ad9d8b7055a322095dc4672e31
+$(PKG)_SUBDIR   := hdfeos
+$(PKG)_FILE     := HDF-EOS2.$($(PKG)_VERSION).tar.Z
+$(PKG)_URL      := ftp://edhs1.gsfc.nasa.gov/edhs/hdfeos/latest_release/$($(PKG)_FILE)
+$(PKG)_DEPS     := gcc hdf4
+
+define $(PKG)_UPDATE
+    echo 'TODO: write update script for hdf-eos2.' >&2;
+    echo $(hdf-eos2_VERSION)
+endef
+
+define $(PKG)_BUILD
+    cd '$(1)' && chmod -R ugo+w .
+    cd '$(1)' && autoconf
+    cd '$(1)' && \
+        ac_cv_func_malloc_0_nonnull=yes \
+        ac_cv_func_realloc_0_nonnull=yes \
+	./configure $(MXE_CONFIGURE_OPTS) \
+        --enable-install-include
+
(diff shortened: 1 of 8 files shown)


OkoSanto:
Anything I can do to help merge this PR?

src/hdf-eos2.mk
```diff
@@ -0,0 +1,34 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := hdf-eos2
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 19v1.00
+$(PKG)_CHECKSUM := 3fffa081466e85d2b9436d984bc44fe97bbb33ad9d8b7055a322095dc4672e31
+$(PKG)_SUBDIR   := hdfeos
+$(PKG)_FILE     := HDF-EOS2.$($(PKG)_VERSION).tar.Z
+$(PKG)_URL      := ftp://edhs1.gsfc.nasa.gov/edhs/hdfeos/latest_release/$(PKG)_FILE
```
tonytheodore: `$(PKG)_FILE` should be `$($(PKG)_FILE)`

tonytheodore:
Thanks, the build for both succeeds on static targets, but the shared builds don't actually builds *.dlls (and fail the test program for hdf-eos5).

I notice there's a cmake build, it's usually better with shared libs and might reduce the need for patch and autoconf. Do you know if cmake is fully supported?

tonytheodore:
> Do you know if cmake is fully supported?

I did a quick check and it needs the appropriate `FindHDF*` cmake files which don't seem to be installed by the respective packages.

I'd disable the shared builds for now and add the usual header to the test programs:

```
/*
 * This file is part of MXE.
 * See index.html for further information.
 */
```

OkoSanto:
> I'd disable the shared builds for now and add the usual header to the test programs:

Done (?) by adding `$(PKG)_BUILD_SHARED = ` to the .mk files.

Also fixed the other issues (sorry about the sloppyness)

tonytheodore:
Thanks, just a couple more tweaks:

 - `$(PKG)` is globally scoped in the `UPDATE` section, use the literal pkg name and `$(hdf-eos5_VERSION)` etc.
 - `$(PKG)` can be used in the `BUILD` section i.e. `test-$(PKG).exe` (optional)
 - hdf-eos2 patch looks like it was created with `make export-patch-*`. This is the preferred format and should be named `hdf-eos2-1-fixes.patch`. Can you convert the hdf-eos5 patches to this format?

OkoSanto:
Done (if I correctly understood the first point).

tonytheodore:
Great, final two comments:
 - add link to bug report for `hdf-eos5-2-fix-compilation-without-szip.patch`
 - consolidate hdf-eos5 patches with:
```
make init-git-hdf-eos5
make import-all-patches-hdf-eos5
make export-patch-hdf-eos5
rm src/hdf-eos5-2-fix-compilation-without-szip.patch
```
The consolidated patches help track the history of a package as seeing the changes to a single file is simpler than finding deleted patches.

OkoSanto:
There is no bugtracker for HDF-EOS5, so I can't link to the bug report (I sent an email). Otherwise: done!

tonytheodore:
Thanks!

PR #1267, opened 2016-03-19
explain how local-pkg-list works

### LINKED ISSUE #1264: remove documentation of LOCAL_PKG_LIST or recover its effect
Variable `LOCAL_PKG_LIST` is not used in code since fde591f2c410fe028c71120addb80cc147a5ec86 but it is still mentioned in `index.html` and in `settings.mk`.

See http://lists.nongnu.org/archive/html/mingw-cross-env-list/2016-03/msg00008.html

starius:
See http://lists.nongnu.org/archive/html/mingw-cross-env-list/2016-03/msg00018.html

fix #1264

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -242,6 +242,7 @@ ifeq ($(IGNORE_SETTINGS),yes)
     $(info [ignore settings.mk])
 else ifeq ($(wildcard $(PWD)/settings.mk),$(PWD)/settings.mk)
     include $(PWD)/settings.mk
+    # `local-pkg-list` may become a default target of `make` here.
 else
     $(info [create settings.mk])
     $(shell { \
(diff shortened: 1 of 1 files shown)


tonytheodore:
Assigned to myself.

TimothyGu:
I prefer not to have this piece of information to be in the main Makefile. FWIW the target may be called `pixie-dust` and depends on the contents of the variable `PIXIES`.

This is the problem I tried (unsuccessfully it seems) to address in https://github.com/mxe/mxe/commit/b89c4d4703d82faa432ccf4117b7e70e712c73d7#diff-b67911656ef5d18c4ae36cb6741b7965R139.

tonytheodore:
I'd say this shouldn't be in the main Makefile (too hard to find) or `settings.mk` (it's only written once and will not pick up any updates).

I'll do an outline document to describe the various extension/customisation points and a more detailed update to https://github.com/mxe/mxe/wiki/How-MXE-builds-its-package.

tonytheodore:
`.DEFAULT_GOAL` from #1740 is more explicit and removes the vagaries of "first target defined".

Issue #1269, opened 2016-03-20
version of protobuf is too old


pavelvat:
```protobuf.mk``` currently have version 2.5.0 this leads to incompatibility when build of ```Clementine```: http://pastebin.com/Zr17VsL9
I tried to upgrade ```protobuf.mk``` to version 2.6.1 and encountered with a very strange problem http://pastebin.com/cb19jtvt
./autogen.sh trying to download http://googletest.googlecode.com/files/gtest-1.5.0.tar.bz2 and fail. ```gtest-1.5.0.tar.bz2``` is successfully downloaded through internet browser or command line (curl, wget), but it fail when downloading inside MXE.

Workaround for this problem:
```
wget http://googletest.googlecode.com/files/gtest-1.5.0.tar.bz2
tar fjx gtest-1.5.0.tar.bz2
wget https://github.com/google/protobuf/archive/2.6.1.tar.gz
tar fzx 2.6.1.tar.gz
cp -r protobuf-2.6.1/ protobuf-2.6.1.orig
cp -r gtest-1.5.0 protobuf-2.6.1/
diff -Naur protobuf-2.6.1.orig/ protobuf-2.6.1 > protobuf-1-fixes.patch
cp protobuf-1-fixes.patch /path/to/mxe/src/
```
clementine.exe was built and launched successfully.

The same problem with ```protobuf.mk``` version 3.0.0-beta-2, but with another link https://googlemock.googlecode.com/files/gmock-1.7.0.zip.

starius:
MXE has [a protection](https://github.com/mxe/mxe/blob/c26f3225fba8ffe9dd54db759d4680a1314d350f/Makefile#L439) against (potentially unverified) downloads during building.

You should create package `gtest` and unpack its source while building `protobuf`. It can be done with macro `PREPARE_PKG_SOURCE`. For example, package `upx` [unpacks](https://github.com/mxe/mxe/blob/c26f3225fba8ffe9dd54db759d4680a1314d350f/src/upx.mk#L20) source of package `uxl`. Add `gtest` to dependencies of `protobuf`.

PR #1270, opened 2016-03-20
nonetwork: print message from replaced functions

starius:
See https://github.com/mxe/mxe/issues/1269

### DIFF
--- a/tools/nonetwork.c
+++ b/tools/nonetwork.c
@@ -4,33 +4,44 @@
 // License: MIT
 
 #include <errno.h>
+#include <stdio.h>
+
+static void print_message() {
+    fprintf(stderr, "Don't use network from MXE build rules!\n");
+}
 
 int connect(int sock, const void *addr, unsigned int len) {
+    print_message();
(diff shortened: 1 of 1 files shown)


tonytheodore:
Nice!

pavelvat:
http://pastebin.com/E4ekNsUR
```
  % Total    % ReceDon't use network from MXE build rules!
```

Maybe one more "\n"?:
```
static void print_message() {
    fprintf(stderr, "\nDon't use network from MXE build rules!\n");
}
```

tonytheodore:
Indeed. Some asterisks would also help with parsing log output - I only checked this in an interactive session.

starius:
I think, calls of `fflush` should be added as well:
```c
static void print_message() {
    fflush(stderr);
    fprintf(stderr, "\nDon't use network from MXE build rules!\n");
    fflush(stderr);
}

pavelvat:
>I think, calls of fflush should be added as well

Ok: http://pastebin.com/Yn0ejyfv

PR #1276, opened 2016-03-26
nonetwork: new line before the message and flush

starius:
`fflush(stderr)` [may be needed](http://mailman.linuxchix.org/pipermail/courses/2002-August/000691.html).

see https://github.com/mxe/mxe/pull/1270


tonytheodore:
Thanks!

PR #1275, opened 2016-03-25
multiCall + case insensitivity

dl5rcw:
now there is just a single commit including multiCall function for the requested option and ifind for more case insensitivity.

PR #1278, opened 2016-03-28
multiCall + case insensitivity

dl5rcw:
End of a long journey (see #1275). @pavelvat - thanks for checking. This should now be okay and included in mxe. Thanks for your assistance!

### DIFF
--- a/tools/copydlldeps.sh
+++ b/tools/copydlldeps.sh
@@ -10,14 +10,14 @@ Welcome to $( basename $0)!
 Authors:  Lars Holger Engelhard - DL5RCW (2016)
           Tiancheng "Timothy" Gu (2014)
 
-Version: 1.1
+Version: 1.3
 
 # This file is part of the MXE Project, sponsored by the named authors
 # it supports the shared build approach by providing an easy way to
 # check for library dependencies in a recursive manner
 
-# Copyright (c) 2014 Tiancheng "Timothy" Gu
-#           (c) 2016 Lars Holger Engelhard - DL5RCW
+# Copyright (c) 2016 Lars Holger Engelhard - DL5RCW
+#           (c) 2014 Tiancheng "Timothy" Gu
 
 #
 # Permission is hereby granted, free of charge, to any person obtaining
@@ -47,7 +47,7 @@ EOF
 OBJDUMP=objdump
 
 # create a temp directory
-tmp=`mktemp -d`
+tmp=$( mktemp -d )
 
 # print an help menu
 help() {
@@ -76,6 +76,7 @@ Operating options:
   -s, --srcdir            [ multiCall ] The directory with DLLs that can be copied. 
   -S, --srcdirs           [ multiCall ] List of directories with DLLs that can be copied. Put "" around them, e.g. "/dir1 /root/dir2 /root/dir3" 
   -R, --recursivesrcdir   [ multiCall ] Target directory for recursive search of folders containing *dll files  
+  -X, --excludepattern    [ multiCall ] Exclude any path that contains such pattern, e.g. /(PREFIX)/(TARGET)/apps/
 
 Optional binary settings:
   -o, --objdump           Specify the path or name of your objdump application
@@ -111,16 +112,16 @@ die() {
 # and findAllSrcDirectories will hunt for dlls in each one recursively
 # it will return a sorted list and duplicates are removed
 findAllSrcDirectories(){
-ar_recursiveDirList=${!1}
-string=""
-for curPath in "${ar_recursiveDirList[@]}"; do
-	for element in $(find $curPath -iname "*.dll"); do
-		#ar_list+="$(dirname $element) "
-		string+="$(dirname $element) "
+	ar_recursiveDirList=${!1}
+	string=""
+	for curPath in "${ar_recursiveDirList[@]}"; do
+		for element in $(find $curPath $excludePattern -iname "*.dll"); do
+			#ar_list+="$(dirname $element) "
+			string+="$(dirname $element) "
+		done
 	done
-done
-string=$(echo "$string" | tr -s ' ' | tr ' ' '\n' | nl | sort -u -k2 | sort -n | cut -f2-)
-echo $string #returns the string 
+	string=$(echo "$string" | tr -s ' ' | tr ' ' '\n' | nl | sort -u -k2 | sort -n | cut -f2-)
+	echo $string #returns the string 
 }
 
 while [ $# -gt 0 ]; do
@@ -150,6 +151,10 @@ while [ $# -gt 0 ]; do
 		;;
 	-R|--recursivesrcdir)
 		recursivesrcdir+=" $1"
+		shift
+		;;
+	-X|--excludepattern)
+		excludepattern+=" $1"
 		shift
 		;;
 	-o|--objdump)
@@ -183,6 +188,8 @@ while [ $# -gt 0 ]; do
 		;;
 	esac
 done
+
+# setting default values if no arguments were given
 if ! [ "$loglevel" ]; then
 	loglevel=0
 fi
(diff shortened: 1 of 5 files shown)


tools/copydlldeps.sh
```diff
@@ -324,16 +334,17 @@ process_enforced_deps(){
...
+	for curFile in $( echo "${enforcedir}" | tr -s ' ' | tr ' ' '\n'); do
+		echo "startig for file $curFile in enforce section"
 		append_deps "$curFile" rel
 		process_enforced_deps "$curFile" 
 	done
 fi
 
 # then we start with our indir or infile list
-for file in $filelist; do
+#for file in $filelist; do
```
starius: Commented code.

tools/copydlldeps.sh
```diff
@@ -343,15 +354,15 @@ echo "I will now search for \$alldeps"
...
 	echo "waiting 10 seconds until I proceed - so you can read my debugOut"
 	sleep 10
 
 	tmpStr=${str_srcDirList}
 	echo "\$alldeps has ${#alldeps[@]} elements"
 	echo "and \$str_srcDirList has ${#str_srcDirList} elements"
-	echo "waiting another 10 seconds"
-	#sleep 10
+	#echo "waiting another 2 seconds"
+	#sleep 2
```
starius: Commented code.

pavelvat:
@dl5rcw See https://github.com/mxe/mxe/pull/1225#issuecomment-202808064. 
Can you prevent searching in directory ```$(PREFIX)/$(TARGET)/apps/```?

dl5rcw:
@pavelvat - this would be another feature. I call it: exclusion pattern. I wrote a solution the past days, but will not push before testing. Maybe end of next weekend, when I return from my business trip.

pavelvat:
@dl5rcw you don't fix second remark from starius above:
```
-	echo "waiting another 10 seconds"
-	#sleep 10
+	#echo "waiting another 2 seconds"
+	#sleep 2
```

dl5rcw:
Please visit my branch. As stated above, I introduced the feature 'exclude pattern'. Of course the suggestion and remark of starius is fixed by version 1.2 already. With all requests that came from various spots, there is already version 1.3 available. Besides fixing the remarks, there is 'exclude patterns' and a nice 'white list feature', as well as a summary. Feel free to give it a try!

pavelvat:
> Of course the suggestion and remark of starius is fixed by version 1.2 already.

not fixed:
https://github.com/mxe/mxe/pull/1278/files#diff-002f74d8843b2b26e7172eb28ddadf1bR393

dl5rcw:
please see: https://github.com/dl5rcw/mxe/blob/master/tools/copydlldeps.sh
```
if [ "$loglevel" -eq 1 ]; then
	echo "waiting 10 seconds until I proceed - so you can read my debugOut"
	sleep 10

	tmpStr=${str_srcDirList}
	echo "\$alldeps has ${#alldeps[@]} elements"
	echo "and \$str_srcDirList has ${#str_srcDirList} elements"
	#echo "waiting another 2 seconds"
	#sleep 2
fi
```
There is the 2 seconds already - but I think I will completely kick out those two lines in the next step.

pavelvat:
@dl5rcw Do you have any plans to add some commits to this pull request?

dl5rcw:
As I moved on, I had to change the machine I was working on and the original repo is gone. So any push seems to not have any affect to this pull request any more. I could possibly close this one.

pavelvat:
>So any push seems to not have any affect to this pull request any more.

Git is a distributed revision control system, so you not need your original repo on old machine.
In order to continue your work on the new machine, run:
```
git clone https://github.com/dl5rcw/mxe.git
```

dl5rcw:
Yeah, thanks. I have done that to improve to version 1.3. But when I did commit + push, it obviously does not affect this pull request. I guess that's why you thought I would not have included the comments.... and when I see the repo in the web account here, I certainly see the new version.

pavelvat:
> But when I did commit + push, it obviously does not affect this pull request.

Why? commit+push send your new commit from your local repo to your remote repo https://github.com/dl5rcw/mxe, this pull request is linked with your remote repo https://github.com/dl5rcw/mxe:
```
dl5rcw wants to merge 2 commits into mxe:master from dl5rcw:master
```

So, due to remove this two commented lines:
```
    #echo "waiting another 2 seconds"
    #sleep 2
```
you must making new commit "remove two commented lines". 
Then you must squash commit:
```
 	introducing excludepattern and whitelist of DLLs   28d54ae
```
with commit "remove two commented lines"
Documentation:
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Changing-Multiple-Commit-Messages
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits

Something like this:
1)git commit -a -m "remove two commented lines"
2)git rebase -i HEAD~2
(2 - is the number of last commits for squashing.)
3)replace ```pick``` on ```squash```: 
```
pick 28d54ae introducing excludepattern and whitelist of DLLs
pick your_some_sha1 remove two commented lines
```
on
```
pick 28d54ae introducing excludepattern and whitelist of DLLs
squash your_some_sha1 remove two commented lines
```
save
edit new message for commit (can remove the text "remove two commented lines")
save
4)make sure that everything is correct
```
git log --pretty=format:"%h - %an, %ar : %s"
```
must contain something like this:
```
28d54ae - Lars, 9 days ago : introducing excludepattern and whitelist of DLLs
9f7c17f - Lars, 8 weeks ago : multiCall + case insensitivity
5366608 - MXEBot as Travis CI, 9 weeks ago : Update versions.json & build-matrix.html
f117a71 - Mark Brand, 9 weeks ago : vmime: update
...
```
5)replace your remote repo (https://github.com/dl5rcw/mxe) state by your local repo state (option -f is necessary):
```
git push origin master -f
```

dl5rcw:
done - according to your howto. thanks.

pavelvat:
@starius This PR seems ready to merge.

starius:
Why was `exit 0` added?

dl5rcw:
I just returned from family vacation. Surprised that this is still open. @starius - there was an exit 1 for the failure exits and exit 0 when help or version was displayed. To make it more convenient in build scripts, a exit 0 was added to the very bottom of the script. That makes it easy to check if everything worked smooth, when not tracing the output with human eyes....

starius:
> exit 0 when help or version was displayed

By default, a shell script already exits with 0 (more precisely, the exit code of last command, `rm`).

> That makes it easy to check if everything worked smooth, when not tracing the output with human eyes.... 

You can add `set -e` to make script exit if a subcommand exits with non-zero.

dl5rcw:
You are all right, the script will always return the last echo $? as the final exit and yes, you may use the set -e option. Yet, any programmer should be aware that a subcommand may return something else than exit 0 which is expected and okay according to the way it is used and according to the man page.
 
Hence, instead of using set -e, it is wise to code a clean exit 0 at the end of your own shell/bash script. 
At least that is what was taught many - well yes - very many years ago... anyway I am not here to teach, but to offer a contribution to your great project. A pleasure.

dl5rcw:
@starius removed exit 0 at the end for your very convenience

tonytheodore:
I can't really comment on the nuances of shell scripts - I prefer the `make` behaviour of failing on subcommands unless explicitly told not to. It's also more obvious to deal with non-zero "pass" codes on the individual command that returns them, but that's a matter of preference.

I'm a little concerned that this tries to do too much and re-implements much of the functionality of standard tools like `cp` and `xargs`. However, if it's working and useful, I'm happy to merge this and let the slimmed down implementation materialise when someone thinks it's necessary.

starius:
This version looks good to me.

@dl5rcw,
could you rebase it one more time to get rid of merge commits and root it in the current master, please?

@tonytheodore,
to get Makefile-like behaviour, a shell script should start with `set -xue` and include only a sequence of commands without branching, loops, functions and argument parsing. If I need aforesaid features, I switch from shell to Lua or Python. Nevertheless this script is written in shell and it is not a problem, if it works correctly.

tonytheodore:
@dl5rcw, thanks! Rebased and merged (with some prefixes for commit messages): https://github.com/mxe/mxe/compare/6f830ee...4d76a91

@starius OT, I just realised I can do this:
```
$ cat shmake
#!/usr/bin/env make -r -R -f

$(info hello $(foo))

all: ; @echo -n

$ ./shmake foo=world!
hello world!
```

PR #1193, opened 2016-01-21
[WIP] CMake: update and improve configuration

tonytheodore:

PR #1279, opened 2016-03-31
Add package vtk7

muellni:
Similar to vtk6, but this one builds against Qt5. It also uses mxe's CMake package. 
I tested it with `i686-w64-mingw32.static` and `i686-w64-mingw32.shared`.

### DIFF
--- a/src/cmake.mk
+++ b/src/cmake.mk
@@ -3,8 +3,8 @@
 
 PKG             := cmake
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 3.0.2
-$(PKG)_CHECKSUM := 6b4ea61eadbbd9bec0ccb383c29d1f4496eacc121ef7acf37c7a24777805693e
+$(PKG)_VERSION  := 3.5.0
+$(PKG)_CHECKSUM := 92c83ad8a4fd6224cf6319a60b399854f55b38ebe9d297c942408b792b1a9efa
 $(PKG)_SUBDIR   := cmake-$($(PKG)_VERSION)
 $(PKG)_FILE     := cmake-$($(PKG)_VERSION).tar.gz
 $(PKG)_URL      := http://www.cmake.org/files/v$(call SHORT_PKG_VERSION,$(PKG))/$($(PKG)_FILE)
--- a/src/cmake.mk
+++ b/src/cmake.mk
@@ -8,7 +8,8 @@ $(PKG)_CHECKSUM := 92c83ad8a4fd6224cf6319a60b399854f55b38ebe9d297c942408b792b1a9
 $(PKG)_SUBDIR   := cmake-$($(PKG)_VERSION)
 $(PKG)_FILE     := cmake-$($(PKG)_VERSION).tar.gz
 $(PKG)_URL      := http://www.cmake.org/files/v$(call SHORT_PKG_VERSION,$(PKG))/$($(PKG)_FILE)
-$(PKG)_TARGETS  := $(BUILD) $(MXE_TARGETS)
+$(PKG)_TARGETS  := $(BUILD)
+$(PKG)_DEPS     :=
 
 define $(PKG)_UPDATE
     $(WGET) -q -O- 'http://www.cmake.org/cmake/resources/software.html' | \
--- a/src/cmake.mk
+++ b/src/cmake.mk
@@ -8,8 +8,7 @@ $(PKG)_CHECKSUM := 92c83ad8a4fd6224cf6319a60b399854f55b38ebe9d297c942408b792b1a9
 $(PKG)_SUBDIR   := cmake-$($(PKG)_VERSION)
 $(PKG)_FILE     := cmake-$($(PKG)_VERSION).tar.gz
 $(PKG)_URL      := http://www.cmake.org/files/v$(call SHORT_PKG_VERSION,$(PKG))/$($(PKG)_FILE)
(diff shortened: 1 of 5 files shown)


src/cmake.mk
```diff
@@ -8,8 +8,7 @@ $(PKG)_CHECKSUM := 92c83ad8a4fd6224cf6319a60b399854f55b38ebe9d297c942408b792b1a9
 $(PKG)_SUBDIR   := cmake-$($(PKG)_VERSION)
 $(PKG)_FILE     := cmake-$($(PKG)_VERSION).tar.gz
 $(PKG)_URL      := http://www.cmake.org/files/v$(call SHORT_PKG_VERSION,$(PKG))/$($(PKG)_FILE)
-$(PKG)_TARGETS  := $(BUILD)
-$(PKG)_DEPS     :=
+$(PKG)_TARGETS  := $(BUILD) $(MXE_TARGETS)
```
starius: Why do you need this change?

starius:
Please remove trailing whitespaces added by you to several files.

starius:
We have packages vtk and vtk6. New one is vtk7. Maybe we should rename vtk to vtk5?

src/cmake.mk
muellni: > Why do you need this change?

When I first added cmake as a dependency to vtk7 it didn't work. But now I try to reproduce the error and it's gone, so I guess this change isn't needed at all.

muellni:
> Please remove trailing whitespaces added by you to several files.

I only found two places, which I now removed in 7c3142a. Did I miss more?

muellni:
> We have packages vtk and vtk6. New one is vtk7. Maybe we should rename vtk to vtk5?

I agree that three VTK packages are too many. 
We could simplify it into a vtk-qt4 and vtk-qt5 package (both version 7). 
Maybe this depends on what people actually use. Are there package usage statistics?

src/cmake.mk
muellni: > Why do you need this change?

So right before the vtk7 package the error triggered: 
```make: *** No rule to make target '/home/sws/projSik/2016/01_build_vtk7/mxe/usr/i686-w64-mingw32.static/installed/cmake', needed by '/home/sws/projSik/2016/01_build_vtk7/mxe/usr/i686-w64-mingw32.static/installed/vtk7'.  Stop.```

This change made it build, but I don't know what it does.

Comment (starius) [inline src/cmake.mk]: Remove `cmake` from `$(PKG)_DEPS     := gcc hdf5 qtbase qttools libpng expat libxml2 jsoncpp cmake tiff`
All cross-compiled packages depend on all native packages implicitly.

Comment (muellni) [inline src/cmake.mk]: > Remove cmake from $(PKG)_DEPS

done at 32c3e47

> All cross-compiled packages depend on all native packages implicitly.

good to know

**Comment by starius** [comment]
The package `vmime` is broken: https://gist.github.com/28829db4c1246a844fb1ee4ec3e28af2

**Comment by tonytheodore** [comment]
> The package vmime is broken

From memory, the `.pc` file is recreated during build and the replacements need to happen after (untested):

```diff
diff --git a/src/vmime.mk b/src/vmime.mk
index 411b140..a14fde5 100644
--- a/src/vmime.mk
+++ b/src/vmime.mk
@@ -36,9 +36,9 @@ define $(PKG)_BUILD
         .
 
     $(MAKE) -C '$(1)' -j '$(JOBS)'
+    $(MAKE) -C '$(1)' install
     $(SED) -i 's,^\(Libs.private:.* \)$(PREFIX)/$(TARGET)/lib/libiconv\.a,\1-liconv,g' $(1)/vmime.pc
     $(if $(BUILD_STATIC),$(SED) -i 's/^\(Cflags:.* \)/\1 -DVMIME_STATIC /g' $(1)/vmime.pc)
-    $(MAKE) -C '$(1)' install
     $(if $(BUILD_SHARED),$(INSTALL) -m644 '$(1)/build/bin/libvmime.dll' '$(PREFIX)/$(TARGET)/bin/')
 
     $(SED) -i 's/posix/windows/g;' '$(1)/examples/example6.cpp'
```

> Are there package usage statistics?

There aren't any usage statistics, but VTK5 is heavily patched and unable to be easily updated. I'd prefer to drop this package (and switch it's downstream `pcl` to something newer).

starius:
The patch doesn't fix the `vmime` failure: https://gist.github.com/a6cb3073310339abf53248345bb8064f

The problem is related to missing `vmime/export-shared.hpp` file.

See also https://github.com/mxe/mxe/pull/1193 (previous attempt to update cmake). It has patches for vmime and other packages.

starius:
Can you move update of cmake to a separate pull request, please?

tonytheodore:
> The patch doesn't fix the vmime failure

Of course! It's modifying the `*.pc` files in the build tree, not the installed version.

Can you try with `$(PREFIX)/$(TARGET)/lib/pkgconfig/vmime.pc`? Setting `-DVMIME_STATIC` skips the inclusion of `vmime/export-shared.hpp`. (I don't have reasonable build machine at the moment).

> Can you move update of cmake to a separate pull request, please?

+1

starius:
I moved vmime related changes to https://github.com/mxe/mxe/pull/1287
It works with both cmake versions.

muellni:
I'm going to create a new pull request which replaces packages ```vtk``` and ```vtk6```. You can close this pull request if you agree.

PR #1282, opened 2016-04-06
Update the sha256 checksum for QtWebkit

firatakandere:
The outdated checksum is updated for the file: [http://ftp.fau.de/qtproject/community_releases/5.6/5.6.0/qtwebkit-opensource-src-5.6.0.tar.xz](http://ftp.fau.de/qtproject/community_releases/5.6/5.6.0/qtwebkit-opensource-src-5.6.0.tar.xz)

### DIFF
--- a/src/qtwebkit.mk
+++ b/src/qtwebkit.mk
@@ -4,7 +4,7 @@
 PKG             := qtwebkit
 $(PKG)_IGNORE   :=
 $(PKG)_VERSION   = $(qtbase_VERSION)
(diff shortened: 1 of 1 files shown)


starius:
I have compared old and new versions. The new version has additional directory `include` in the archive, otherwise they are the same. I think, it should not introduce a problem. I am merging it after rebuilding the package.

starius:
Works for me after applying https://github.com/mxe/mxe/pull/1284 (fix unrelated bug in qtbase).

PR #1284, opened 2016-04-06
fix qtbase on Wheezy (don't use c++1y)

starius:
The error: `g++: error: unrecognized command line option ‘-std=c++1y’`
Full log: https://gist.github.com/147a8e697b4be73dad0218598dab2608

The problem was solved by setting C++ version to C++11 explicitly.


tonytheodore:
This may need to be revisited when we switch to a later version of `gcc`, maybe to wrap it in some sort of feature detection - I'm sure someone will let us know.

Thanks.

PR #1286, opened 2016-04-08
add package djvulibre

darealshinji:
~~I wasn't able to build v3.5.27 yet. I was able to build shared libraries but they were broken.~~
Right now only shared libraries can be build and used. Static libraries, if build, are broken.


src/djvulibre.mk
```diff
@@ -0,0 +1,29 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := djvulibre
+$(PKG)_IGNORE   := 3.5.27
+$(PKG)_VERSION_ := 3.5.25
```
starius: I think, `$(PKG)_VERSION_` should be renamed to `$(PKG)_SHORTVERSION`.

src/djvulibre.mk
```diff
@@ -0,0 +1,29 @@
...
+$(PKG)_VERSION  := $($(PKG)_VERSION_).3
+$(PKG)_CHECKSUM := 898d7ed6dd2fa311a521baa95407a91b20a872d80c45e8245442d64f142cb1e0
+$(PKG)_SUBDIR   := $(PKG)-$($(PKG)_VERSION_)
+$(PKG)_FILE     := $(PKG)-$($(PKG)_VERSION).tar.gz
+$(PKG)_URL      := https://sourceforge.net/projects/djvu/files/DjVuLibre/$($(PKG)_VERSION_)/$($(PKG)_FILE)/download
+$(PKG)_DEPS     := gcc jpeg tiff zlib
+
+define $(PKG)_UPDATE
+    $(WGET) -q -O- 'http://sourceforge.net/projects/djvu/files/DjVuLibre/' | \
```
starius: (Here and in `URL`)
Please use macro `$(SOURCEFORGE_MIRROR)`.

starius:
Can you combine all patches to single file in git like format, please?
To create such patch, `./tools/patch-tool-mxe` can be used:

```
$ ./tools/patch-tool-mxe init djvulibre
$ cd gits/djvulibre-3.5.25
$ git apply path/to/first.patch
$ git commit -m 'description (reason, error fixed, etc)'
$ git apply path/to/second.patch
$ git commit -m 'description (reason, error fixed, etc)'
and so on
$ cd -
$ ./tools/patch-tool-mxe export djvulibre
```

Single file with all patches in unified format makes browsing history easier.

src/djvulibre.mk
darealshinji: How exactly do I have to add the macro? Replacing `https://sourceforge.net` or `sourceforge.net` makes the downloads fail.

darealshinji:
Like this? https://gist.github.com/darealshinji/e0da804a3eff975954408232a3e4cf1b

starius:
> Like this? https://gist.github.com/darealshinji/e0da804a3eff975954408232a3e4cf1b

No, it lacks commit messages, author and date.

I contributed all the changes in the pull request to your branch: https://github.com/darealshinji/mxe/pull/1

darealshinji:
Ah, thanks. Merged.

starius:
Why are shared targets disabled?

darealshinji:
I was able to build them but they didn't seem to work. Maybe you can give it a try?

starius:
I have built static targets. They produce the following files:
https://gist.github.com/74156e2a0a1662f9a2b9f7723a4b0948

Please prevent installation of `$(PREFIX)/$(TARGET)/bin/*.exe` and `$(PREFIX)/$(TARGET)/bin/any2djvu`. The latter is a shell script.

Please prevent installation of `$(PREFIX)/$(TARGET)/share/man`.

Are other files in `$(PREFIX)/$(TARGET)/share` needed to use this library? They seem to be a documentation. If they are not needed, please prevent their installation as well.

Please provide a simple test: file `src/test-djvulibre.c` and a command in the build rules, which produces executable `$(PREFIX)/$(TARGET)/bin/test-djvulibre.exe` from it. It should use pkg-config to generate compiler and linker flags. The executable should exit with 0 exit status, of course.

Having this work done, it will be easier to deal with shared targets, I think.

darealshinji:
Should the executables still be built or should I only build the library?

starius:
Executables should be built to catch build errors.

darealshinji:
Alright. To only install the build stuff you can use `make -C libdjvu install-lib install-include install-pkgconfig`.
This installs the following files for me:
```
include/libdjvu/ddjvuapi.h
include/libdjvu/miniexp.h
lib/libdjvulibre.a
lib/pkgconfig/ddjvuapi.pc
lib/libdjvulibre.la
```

Will this serve as a test program?
``` c
#include <windows.h>
#include <libdjvu/ddjvuapi.h>

int main(int argc, char *argv[])
{
	ddjvu_context_t *djvu_test;
	djvu_test = ddjvu_context_create(argv[0]);
	ddjvu_context_release(djvu_test);
	return 0;
}
```

starius:
> Will this serve as a test program?

Can it work without include of windows.h to keep it cross-platform?
Otherwise looks good to me.

darealshinji:
windows.h wasn't necessary. And the created test program works as expected on Windows and returns 0.

starius:
I'm investigating shared build issues. i686-w64-mingw32.shared works with [the patch](https://gist.github.com/da80fc2a43d49a592294bb882c460e67).

x86_64-w64-mingw32.shared is still broken because of linking problems with jpeg.

I am going to make another pull request against your branch.

starius:
I have fixed shared targets in a pull request https://github.com/darealshinji/mxe/pull/2

src/djvulibre.mk
```diff
@@ -0,0 +1,35 @@
...
+    head -1
+endef
+
+define $(PKG)_BUILD
+    cd '$(1)' && CPPFLAGS='-DDLL_EXPORT' ./configure \
+        $(MXE_CONFIGURE_OPTS) \
+        --disable-desktopfiles \
+        $(if $(BUILD_SHARED),\
+            lt_cv_deplibs_check_method='file_magic file format (pe-i386|pe-x86-64)' \
+            lt_cv_file_magic_cmd='$$OBJDUMP -f')
```
TimothyGu: This is an indication that the libtool included in the tarball is ancient. Try using `autoreconf -fi` instead of manually specifying the variables.

src/djvulibre.mk
```diff
@@ -0,0 +1,35 @@
...
+$(PKG)_DEPS     := gcc jpeg tiff zlib
+
+define $(PKG)_UPDATE
+    $(WGET) -q -O- 'http://sourceforge.net/projects/djvu/files/DjVuLibre/' | \
+    $(SED) -n 's,.*/\([0-9][^A-Za-z"]*\)/".*,\1,p' | \
+    head -1
+endef
+
+define $(PKG)_BUILD
+    cd '$(1)' && CPPFLAGS='-DDLL_EXPORT' ./configure \
```
TimothyGu: Does `-DDLL_EXPORT` have any negative effects on static builds? Just to make sure.

src/djvulibre.mk
starius: Fixed in https://github.com/darealshinji/mxe/pull/3

darealshinji:
Maybe for the static targets we should add `#define MINILISPAPI __declspec(dllexport)` and `#define DDJVUAPI __declspec(dllexport)` to the headers, otherwise the users will always have to manually add `-DDLL_EXPORT` to their build flags to link against the static libraries.

TimothyGu:
> Maybe for the static targets we should add `#define MINILISPAPI __declspec(dllexport)` and `#define DDJVUAPI __declspec(dllexport)` to the headers

That sounds wrong. Actually, we DON'T want to do this for static builds during the compilation process, which is harmless, OR when users are compiling their own code, which is harmful.

darealshinji:
When I built a pdf/djvu viewer and wanted to link it against the static library I got errors about unresolved import library symbols. But apparently the library can't be build at all without the `DLL_EXPORT` macro, so right now only a dll can be build.
``` make
# This file is part of MXE.
# See index.html for further information.

PKG             := djvulibre
$(PKG)_IGNORE   := 3.5.27
$(PKG)_SHORTVER := 3.5.25
$(PKG)_VERSION  := $($(PKG)_SHORTVER).3
$(PKG)_CHECKSUM := 898d7ed6dd2fa311a521baa95407a91b20a872d80c45e8245442d64f142cb1e0
$(PKG)_SUBDIR   := $(PKG)-$($(PKG)_SHORTVER)
$(PKG)_URL      := http://$(SOURCEFORGE_MIRROR)/project/djvu/DjVuLibre/$($(PKG)_SHORTVER)/$($(PKG)_FILE)
$(PKG)_DEPS     := gcc jpeg tiff zlib

define $(PKG)_UPDATE
    $(WGET) -q -O- 'http://sourceforge.net/projects/djvu/files/DjVuLibre/' | \
    $(SED) -n 's,.*/\([0-9][^A-Za-z"]*\)/".*,\1,p' | \
    head -1
endef

define $(PKG)_BUILD_SHARED
    cd '$(1)' && autoreconf -fi
    cd '$(1)' && ./configure $(MXE_CONFIGURE_OPTS) --disable-desktopfiles
    $(MAKE) -C '$(1)/libdjvu' -j 1 install-lib install-include install-pkgconfig

    '$(TARGET)-g++' \
        -W -Wall -Werror -pedantic \
        '$(2).c' -o '$(PREFIX)/$(TARGET)/bin/test-$(PKG).exe' \
        `'$(TARGET)-pkg-config' --cflags --libs ddjvuapi`
endef
```
I still need to check if I can properly build/link the djvu viewer (qpdfview, if anyone wants to know) under the `.shared` mxe targets.

darealshinji:
The shared library builds and runs without problems on the i686-w64-mingw32.shared target.

darealshinji:
Yesterday I gave it another try and I really wasn't able to build a proper static library.
I've updated the Makefile with those new macros.

starius:
It fails here in `automake` with the following error:

```
configure.ac:44: error: version mismatch.  This is Automake 1.15,
configure.ac:44: but the definition used by this AM_INIT_AUTOMAKE
configure.ac:44: comes from Automake 1.14.1.  You should recreate
configure.ac:44: aclocal.m4 with aclocal and run automake again.
```

Why did you [replaced](https://github.com/mxe/mxe/commit/4e4f031a698507c06483036b138d7c98815daa5c#diff-3901ed1b048a274ffcbe83edbc19de50R21) `autoreconf -fi` with `automake`?

darealshinji:
> Why did you replaced autoreconf -fi with automake?

I didn't realize that. It's now fixed.

starius:
Next problem: s/$(TEST_FILE).c/$(TEST_FILE)/

starius:
After applying `s/$(TEST_FILE).c/$(TEST_FILE)/` the build succeeded.
The list of files installed: https://gist.github.com/d866baba1a23870df9308615ea218e14

I applied the following patch:
```diff
diff --git a/src/djvulibre.mk b/src/djvulibre.mk
index 3e4f394..4caf6e1 100644
--- a/src/djvulibre.mk
+++ b/src/djvulibre.mk
@@ -22,10 +22,11 @@ define $(PKG)_BUILD_SHARED
     cd '$(BUILD_DIR)' && $(SOURCE_DIR)/configure \
         $(MXE_CONFIGURE_OPTS) \
         --disable-desktopfiles
-    $(MAKE) -C '$(BUILD_DIR)' -j '$(JOBS)' install-strip
+    $(MAKE) -C '$(BUILD_DIR)' -j '$(JOBS)'
+    $(MAKE) -C '$(BUILD_DIR)' -j 1 install-strip $(MXE_DISABLE_CRUFT)
 
     '$(TARGET)-g++' \
         -W -Wall -Werror -pedantic \
-        '$(TEST_FILE).c' -o '$(PREFIX)/$(TARGET)/bin/test-$(PKG).exe' \
+        '$(TEST_FILE)' -o '$(PREFIX)/$(TARGET)/bin/test-$(PKG).exe' \
         `'$(TARGET)-pkg-config' ddjvuapi --cflags --libs`
 endef
```

to prevent it from installing docs and tools.

(And to prevent installation with -j > 1.)

Now the list of files is as following:
https://gist.github.com/f728df76fdd000b40d10cba04ce392e0

Looks better, but it still installs shell scripts: any2djvu, djvudigital.

Could you apply my patches and prevent installation of these shell scripts?

darealshinji:
> Could you apply my patches and prevent installation of these shell scripts?

Done.

starius:
Exclusion of shell scripts can be achieved without patching source:

```diff
diff --git a/src/djvulibre.mk b/src/djvulibre.mk
index 4caf6e1..5e3fc5b 100644
--- a/src/djvulibre.mk
+++ b/src/djvulibre.mk
@@ -23,7 +23,8 @@ define $(PKG)_BUILD_SHARED
         $(MXE_CONFIGURE_OPTS) \
         --disable-desktopfiles
     $(MAKE) -C '$(BUILD_DIR)' -j '$(JOBS)'
-    $(MAKE) -C '$(BUILD_DIR)' -j 1 install-strip $(MXE_DISABLE_CRUFT)
+    $(MAKE) -C '$(BUILD_DIR)' -j 1 install-strip \
+        $(MXE_DISABLE_CRUFT) dist_bin_SCRIPTS=
 
     '$(TARGET)-g++' \
         -W -Wall -Werror -pedantic \
```

darealshinji:
Very convenient, I didn't know that.

starius:
List of files installed: https://gist.github.com/ea0a95023b6bfa8403f0e6611bf242ff

Looks good to me. Very well done!

PR #1291, opened 2016-04-11
[WIP] add libuv

starius:
Test binary works on Windows.

TODO: patch `libuv-1-fixes.patch` just copies dependencies from Makefile.am.
`LIBS` variable (in upstream version) becomes equal to `-lpthreads`.

From generated `Makefile.am`:
```make
am__append_4 = -lws2_32 -lpsapi -liphlpapi -lshell32 -luserenv
...
LIBS = -lpthread  $(am__append_4)
```

From `configure.ac`:
```ac
 AS_IF([test "x$PKG_CONFIG" != "x"], [
    AC_CONFIG_FILES([libuv.pc])
])
AC_CONFIG_FILES([Makefile])
```

Therefore, `Makefile` is generated after `libuv.pc`. Variables from `Makefile` don't affect `config.status` and `libuv.pc`.

The following work:
```
$ LIBS="-lpthread -lws2_32 -lpsapi -liphlpapi -lshell32 -luserenv" ./config.status --recheck
$ ./config.status libuv.pc
```
But how to extract that `LIBS` variable from `Makefile`?

### DIFF
--- a/index.html
+++ b/index.html
@@ -1910,6 +1910,10 @@ local-pkg-list: $(LOCAL_PKG_LIST)</pre>
         <td class="website"><a href="http://libusb.org/">LibUsb-1.0</a></td>
     </tr>
     <tr>
+        <td class="package">libuv</td>
+        <td class="website"><a href="http://libuv.org/">libuv</a></td>
+    </tr>
+    <tr>
         <td class="package">libvpx</td>
         <td class="website"><a href="http://code.google.com/p/webm/">vpx</a></td>
     </tr>
--- a/src/libuv-1-fixes.patch
+++ b/src/libuv-1-fixes.patch
@@ -4,24 +4,34 @@ See index.html for further information.
 Contains ad hoc patches for cross building.
 
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
-From: Boris Nagaev <bnagaev@gmail.com>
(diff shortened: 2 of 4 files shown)


tonytheodore:
I think the `LIBS` list needs to be moved to `configure.ac` so that it's available to all the `AC_CONFIG_FILES`.

First I tried with `AC_CHECK_LIB` but that won't work as expected due to [differences in call signatures that require appropriate headers to be included](https://lists.gnu.org/archive/html/autoconf/2013-12/msg00019.html).

The below seems to work for having a single list and should be suitable for upstream:
https://github.com/tonytheodore/mxe/blob/LuaAndC-libuv/src/libuv-1-fixes.patch

starius:
It works for me.
New files created by the package: https://gist.github.com/6b86d95f3aa1cb942a360eb71a25c932
@tonytheodore, can you contribute it in a pull request, please?

tonytheodore:
See https://github.com/LuaAndC/mxe/pull/5. There's extra merges in there that rebasing should take care of.

starius:
@tonytheodore, thanks! I merged your pull request.

tonytheodore:
Thanks!

starius:
The patch was sent to upstream: https://github.com/libuv/libuv/pull/841

Issue #1221, opened 2016-02-16
How can I uninstall mxe


userow:
How can I uninstall mxe ?

starius:
Do you want to uninstall whole MXE or individual packages?

userow:
First of all - whole.

I've installed it for testing purposes.

TimothyGu:
How did you install it? Through Debian packages or locally?

If locally, then

```sh
rm -rf mxe
```

userow:
OS X, through make / makefile after installing of required tools through MacPorts.

starius:
> `rm -rf mxe`

It doesn't work because of read-only directory:

```bash
$ rm -rf mxe/
rm: cannot remove ‘mxe/usr/readonly/.gitkeep’: Permission denied
```

To remove it, you can use the following commands:

```bash
$ chmod +w mxe/usr/readonly
$ rm -rf mxe
```

I propose to do keep this directory read-only only while building to prevent such a problem in future. Instead of running `chmod -w mxe/usr/readonly` at creation time, we should run it before each build and run `chmod +w mxe/usr/readonly` after each build.

tonytheodore:
@starius wrote:
> I propose to do keep this directory read-only only while building to prevent such a problem in future. 

Agreed, however that may cause (silent) issues with parallel builds. It should be created in `tmp-<pkg>-<target>` instead.

@userow, you'll also want to revert any changes to `$PATH`.

starius:
There are also problems with git in `usr/` created by `build-pkg`:
```
$ git checkout master 
warning: unable to unlink readonly/.gitkeep: Permission denied
```

PR #1292, opened 2016-04-11
add "make shell" command

starius:
It provides interactive shell with the environment of MXE build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -460,6 +460,10 @@ $(NONET_LIB): $(TOP_DIR)/tools/nonetwork.c | $(PREFIX)/$(BUILD)/lib/.gitkeep
 	@echo '[build nonetwork lib]'
 	@$(BUILD_CC) -shared -fPIC $(NONET_CFLAGS) -o $@ $<
 
+.PHONY: shell
+shell:
+	$(PRELOAD) $(SHELL)
+
 define PKG_TARGET_RULE
 .PHONY: $(1)
 $(1): $(PREFIX)/$(3)/installed/$(1)


tonytheodore:
Nice! That's a great approach that has many possibilities - we could change build rules to shell variables instead of `make` variables to allow easy copy-pasting for tutorials/debugging...

A couple of little tweaks in https://github.com/LuaAndC/mxe/pull/4. It would be good to move the creation of `WINEPREFIX` out of `mxe-conf` and add it as dependency of `shell`, but that can be done separately.

starius:
> It would be good to move the creation of WINEPREFIX out of mxe-conf and add it as dependency of shell, but that can be done separately.

See https://github.com/mxe/mxe/issues/1221#issuecomment-185110607
Removing MXE tree is complicated because of readonly directory.

tonytheodore:
How would you like to handle these types of merges:
  1. use the merge button
  2. person merging rebases and pushes from local, manually closing pull request
  3. person proposing rebases and force-push to update request, then merge with button

I'm beginning to think that rebasing to reduce merges is a losing battle and the [new squash feature is even worse](https://github.com/blog/2141-squash-your-commits).

tonytheodore:
4. Disable "squash" and use the merge button ;)

Thanks!

starius:
I prefer option 3: person proposing rebases and force-push to update request, then merge with button.

> new squash feature is even worse.

Agreed. Squashing kills atomic nature of commits. I'd prefer to have "Rebase and merge" instead (without squash!).

> rebasing to reduce merges

A merge commit after a pull request to mxe:master is important, as it "closes" the pull request. It serves as a separator in the history. I think, that pull requests (followed by merge commits) should be the only source of new commits in `master`.

A merge commit after a cross- pull request (e.g., 8499a63d82e3a75578ec92bcfe4c837b9dbda082) are not so important and can be removed using `git rebase` after merging them with "Merge" button. Merge commits from `master` should be avoided and replaced with rebase, as they break a sequence of commits of the pull request with unrelated commits from `master`.

Turning to rebase-before-merge in mxe:master, it is useful not because it avoids a merge commit (see above why), but because it **groups the commits of the pull request**, simplifying the output of `git log --graph`. It is easy to read when commits from different pull requests do not intermix and the number of leading `| ` is low (ideally is equal to 1).

PR #1293, opened 2016-04-11
libsoup: disable gssapi

tonytheodore:
On OSX, native `/usr/bin/krb5-config` is detected and tries to enable gssapi.


starius:
Tested on all 4 targets.

tonytheodore:
Thanks!

PR #1295, opened 2016-04-12
add native build of `patch` for darwin

tonytheodore:
same as #1170, MacPorts has `gpatch` but Homebrew and Rudix don't


tonytheodore:
This seems like a good test of the new "Confirm squash and merge" button. It's okay for single commits, but now seems to be the default.

PR #1298, opened 2016-04-12
Update SFML to the version 2.3.2

bluszcz:
There is available a new version already: http://www.sfml-dev.org/download/sfml/2.3.2/

It builds correctly:

 ```
> file ./usr/i686-w64-mingw32.static/bin/test-sfml.exe 
./usr/i686-w64-mingw32.static/bin/test-sfml.exe: PE32 executable (console) Intel 80386, for MS Windows
 ```
I have tried also executing binary file with wine - it worked out correctly unveiling a red eclipse.

### DIFF
--- a/src/sfml.mk
+++ b/src/sfml.mk
@@ -3,8 +3,8 @@
 
 PKG             := sfml
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 2.3.1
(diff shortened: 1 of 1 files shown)


tonytheodore:
Can you rebase this into a single commit? Note that `make update-package-foo` will change both at the same time.

starius:
Tested on all 4 targets.

bluszcz:
I was not aware that command, will PR soon correct one.

PR #1299, opened 2016-04-13
Add Vc

boudewijnrempt:
Vc is a library offering SIMD Vector Classes for C++

https://github.com/VcDevel


starius:
Thank you! I have some thoughts on this pull request.

  * MXE provides cmake and a wrapper for it. Please use `'$(TARGET)-cmake'` instead of `cmake -DCMAKE_TOOLCHAIN_FILE='$(CMAKE_TOOLCHAIN_FILE)' -DCMAKE_BUILD_TYPE=Release`.
  * Do you need `-DCMAKE_SYSTEM_PROCESSOR=x86`?
  * Please separate build and install steps. Build with `-j '$(JOBS)'`, install with `-j 1`.
  * Remove an empty line before `endef`.
  * The updater doesn't work. `make update-package-vc` tries to update to version 87b0525 and fails to download a file.
  * Shared targets install `lib/libVc.a`, but they should install `bin/libVc-xx.dll` and `lib/libVc.dll.a`.
  * Please provide `*.pc` file to facilitate the use of the library. You can use files from `usr/i686-w64-mingw32.static/lib/pkgconfig/` as examples. See also https://secure.freedesktop.org/~dbn/pkg-config-guide.html
  * Could you add file `src/vc-test.cpp` with a sample application using this library (and pkg-config for flags) and install it to `'$(PREFIX)/$(TARGET)/bin/test-vc.exe'`, please?

boudewijnrempt:
> Thank you! I have some thoughts on this pull request.
>
>  *  MXE provides cmake and a wrapper for it. Please use '$(TARGET)-cmake' instead of cmake -DCMAKE_TOOLCHAIN_FILE='$(CMAKE_TOOLCHAIN_FILE)' -DCMAKE_BUILD_TYPE=Release.

Okay -- I'm afraid I just blindly copied the eigen3.mk file, since that's a library I'm familiar with :-)

>  *  Do you need -DCMAKE_SYSTEM_PROCESSOR=x86?

Yes, this is a bit dumb, but vc checks for that (or amd) to determine that it's not being built on ARM.

>  *  Please separate build and install steps. Build with -j '$(JOBS)', install with -j 1.
>  *  Remove an empty line before endef.
>  *  The updater doesn't work. make update-package-vc tries to update to version 87b0525 and fails to download a file.
>  *  Shared targets install lib/libVc.a, but they should install bin/libVc-xx.dll and lib/libVc.dll.a.

I'm not sure what's up with that...

>  *  Please provide *.pc file to facilitate the use of the library. You can use files from usr/i686-w64-mingw32.static/lib/pkgconfig/ as examples. See also https://secure.freedesktop.org/~dbn/pkg-config-guide.html
>  *  Could you add file src/vc-test.cpp with a sample application using this library (and pkg-config for flags) and install it to '$(PREFIX)/$(TARGET)/bin/test-vc.exe', please?

I will try that, but I'm a bit stressed at the moment because we need to release real soon now...

-- 
Boudewijn Rempt | http://www.krita.org, http://www.valdyas.org

starius:
>> Do you need `-DCMAKE_SYSTEM_PROCESSOR=x86`

> Yes, this is a bit dumb, but vc checks for that (or amd) to determine that it's not being built on ARM.

So, you can pass `x86_64` in case of amd64:

```make
$(if $(findstring x86_64,$(TARGET)),x86_64,x86)
```

PR #1300, opened 2016-04-13
vtk update

muellni:
Replaces #1279.
Update package vtk to version 7. Replaces package vtk6.
This version links to Qt5 instead of Qt4.


starius:
Maybe it is better to keep vtk6 as it is (renamed to vtk_qt4) with Qt4 to compatibility with old applications using Qt4 and vtk?

muellni:
> Maybe it is better to keep vtk6 as it is (renamed to vtk_qt4) with Qt4 to compatibility with old applications using Qt4 and vtk?

Fine for me. I removed it, because I thought you'd prefer less package variants.

muellni:
@starius : here's the `vtk_qt4` package you proposed

tonytheodore:
> Maybe it is better to keep vtk6 as it is (renamed to vtk_qt4)

I think it's better to keep it as it is - no renaming. The ideal situation I see is:

 - vtk: remove package with deprecation warning to use vtk6
 - vtk6: no change with qt4 backend
 - vtk7: new package with qt5

This has minimal impact for any current users of `vtk` and there are no surprises.

The `*_qt*` variants are better handled by a generalisation of #579, I don't have any WIP on this but I am coalescing thoughts on dependencies vs output variants.

muellni:
replaced by #1527

PR #1301, opened 2016-04-13
Add LibRaw

boudewijnrempt:
LibRaw is a raw image decoder library. The package used is a combination
of http://www.libraw.org/data/LibRaw-0.17.1.tar.gz and https://github.com/LibRaw/LibRaw-cmake.

The LibRaw people are not maintaining the cmake build system for their
library, however, the autotools version doesn't build for Windows because
of inconsistent dll import/export.


starius:
Similar to https://github.com/mxe/mxe/pull/1303
Some changes should be done to merge this.

tonytheodore:
Replaced by #2481

PR #1305, opened 2016-04-13
Update libical

### LINKED ISSUE #1302: package libical is broken
Commit: a7e50dbde1deeaf3a8b104b9094a46c3625dce28
Log: https://gist.github.com/8ba59cde9a8f0b9d9b70f5b1a1eb9753
It can be related to https://github.com/mxe/mxe/pull/1290 (cmake update).

starius:
The list of files installed by libical updated: https://gist.github.com/010f93ae6d49aff2af3e4b772605ed4d

close #1302

### DIFF
--- a/src/libical-1-fixes.patch
+++ b/src/libical-1-fixes.patch
@@ -1,0 +1,133 @@
+This file is part of MXE.
+See index.html for further information.
+
+Contains ad hoc patches for cross building.
+
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Boris Nagaev <bnagaev@gmail.com>
+Date: Wed, 13 Apr 2016 20:54:43 +0300
+Subject: [PATCH] fix definitions of gmtime/gmtime_r
+
+Fix the following error:
+config.h:508:37: error: expected initializer before '?' token
+
+The error is caused by the following combination:
+
+ #include <unistd.h>
+ #define gmtime_r(tp,tmp) ((gmtime(tp))?(*(tmp)=*(gmtime(tp)),(tmp)):0)
+ #include <time.h>
+
+time.h declares and defines gmtime_r. Because of the macro, the definition
+of gmtime_r is broken. This commit moves time.h before the macro define.
+
+diff --git a/config.h.cmake b/config.h.cmake
+index 1111111..2222222 100644
+--- a/config.h.cmake
++++ b/config.h.cmake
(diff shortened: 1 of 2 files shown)


src/libical-1-fixes.patch
```diff
@@ -0,0 +1,133 @@
...
+
+diff --git a/examples/CMakeLists.txt b/examples/CMakeLists.txt
+index 1111111..2222222 100644
+--- a/examples/CMakeLists.txt
++++ b/examples/CMakeLists.txt
+@@ -23,19 +23,11 @@ add_dependencies(doesnothing ical icalss icalvcal)
+ if(NOT STATIC_ONLY)
+   target_link_libraries(doesnothing ical icalss icalvcal)
+ else()
+-  if(NOT WIN32)
```
tonytheodore: Can `if(NOT WIN32)` be changed to `if(UNIX OR MINGW)` or `if(NOT MSVC)`?

src/libical-1-fixes.patch
```diff
@@ -0,0 +1,133 @@
...
+ #endif
+-#include <time.h>
+ 
+ /* localtime_r - thread safe localtime() really only needed on Unix */
+ #if !defined(HAVE_LOCALTIME_R)
+
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Boris Nagaev <bnagaev@gmail.com>
+Date: Thu, 14 Apr 2016 01:33:47 +0300
+Subject: [PATCH] test, examples: fix linking with static library
```
tonytheodore: Alternative is to disable the examples and tests: https://github.com/LuaAndC/mxe/pull/6/commits/780fc7139fd775b3c9d934966de3f3a63a63b922

starius: It is a useful option for upstream, but we should not use it in MXE, because building tests and examples serves as a test for linking errors.

src/libical-1-fixes.patch
starius: I tried `if(NOT MSVC)` and it didn't work:

```
** No rule to make target `lib/libical.a', needed by `bin/doesnothing.exe'.  Stop.
```

It creates `build/bin/libical.a` instead of `build/lib/libical.a`.

I think, that right approach is to let cmake pick the path.

starius:
Rebased.

tonytheodore:
Thanks!

PR #1306, opened 2016-04-16
add GNU Tar to requirements and specify OS X packages

tonytheodore:
needed for `build-pkg.lua`, installed as:
 - `gtar` on Homebrew
 - `gnutar` on MacPorts and Rudix

so will be selected by `tool` function


starius:
Thanks!

PR #1307, opened 2016-04-18
update qbittorrent to 3.3.4

starius:
The following patches were included in the upstream:

  * convert includes like <Windows.h> to lowercase
    https://github.com/qbittorrent/qBittorrent/pull/4505
  * fix library list
    https://github.com/qbittorrent/qBittorrent/pull/4824


pavelvat:
What about problem with "std::" prefix, like this https://github.com/mxe/mxe/issues/1104?
https://github.com/qbittorrent/qBittorrent/blob/master/src/base/net/smtp.cpp#L522

starius:
@pavelvat, `std::snprintf` is used only in `src/base/net/smtp.cpp`. This file doesn't include `libintl.h` (which is the source of `std::` problem), that is why qbittorrent is not affected. To check it, I used the following commands:

```
$ make shell
$ cd gits/qbittorrent-3.3.4/src/base/net
$ i686-w64-mingw32.static-g++ -E smtp.cpp -o /tmp/smtp.i -I ../../../../../usr/i686-w64-mingw32.static/qt/include/ -I ../../../../../usr/i686-w64-mingw32.static/qt/include/QtCore/ -I ../.. -I ../../../../../usr/i686-w64-mingw32.static/qt/include/QtNetwork/
$ grep libintl.h /tmp/smtp.i
```

tonytheodore:
Thanks!

PR #1312, opened 2016-04-20
Use correct cmake variable for static assimp library.

m-ou-se:


starius:
Thank you!

Without this commit static targets install `bin/libassimp.dll` and `lib/libassimp.dll.a` as if they are shared targets. With this commit, static targets install `lib/libassimp.a` as expected.

PR #1322, opened 2016-04-30
add package luv to luarocks plugin

starius:
i686-w64-mingw32.shared was tested on Windows XP with the [official example](https://github.com/luvit/luv/blob/master/README.md) of TCP echo server.


tonytheodore:
x86_64-w64-mingw32.shared on Windows 10 also works with echo server. Thanks!

PR #1323, opened 2016-05-02
jsoncpp: fix build with GCC < 6.x

tehnick:
Fix after 0bc73f739d7a8f2bc4e5100da62a0d894cbc9e38.
There is no -Werror=shift-negative-value in GCC 4.9.x.


starius:
See also https://github.com/mxe/mxe/pull/1324
I think, we should update that table in that very commit which fixes a problem.

starius:
Works for me both with gcc6 plugin and without it.

PR #1324, opened 2016-05-02
update table of packages broken by gcc6 plugin

starius:
@tehnick, @tonytheodore, I have updated the table of broken packages.

Changes:

  * libical is not broken
  * guile, ocaml-lablgtk2 and qtwebkit are broken
  * for qtwebkit only i686-w64-mingw32.shared is broken, but x86_64-w64-mingw32.shared works. This package is shared-only.

Log (only first pass of build-pkg): https://gist.github.com/starius/8b3c9547b89d3c0ca29762adcf3a7bb5


tonytheodore:
Thanks, I've [added note](https://github.com/mxe/mxe/commit/3d816111a9caa185acef50f2c1c692d64e69edcd) following https://github.com/mxe/mxe/pull/1323#issuecomment-216201665

tehnick:
> I've added note following #1323 (comment)

In fact GCC 6.x related fix was in 0bc73f7. And it may be interesting for users who want to use GCC 6.x. Commit a1cc019 is done only for compatibility with GCC 4.9 and almost useless for GCC 6.x related issues.

PR #1327, opened 2016-05-04
Gcc6 related fixes

tehnick:


tehnick:
All changes were tested with old GCC.

tonytheodore:
Changes to `patch` [are failing on OSX](https://gist.github.com/tonytheodore/ffed9cff73cac2111d02e7051e2e1fee)

starius:
Why option --binary was added to `patch`? Patches can be applied now
without this option and it works for both Linux and Windows line ends.
Files with Windows line ends keep Windows line ends in a patch, but other
lines of a patch (e.g., `diff --git a/ofstd/libsrc/ofstd.cc
b/ofstd/libsrc/ofstd.cc`) have Linux line ends.

On Thu, May 5, 2016 at 11:45 PM, Tony Theodore <notifications@github.com>
wrote:

> Changes to patch are failing on OSX
> <https://gist.github.com/tonytheodore/ffed9cff73cac2111d02e7051e2e1fee>
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/mxe/mxe/pull/1327#issuecomment-217341436>
>



-- 


Best regards,
Boris Nagaev

tonytheodore:
> Why option --binary was added to `patch`? 

The `freeimage` patch appears to need it:
```
Failed to build package freeimage for target x86_64-w64-mingw32.static!
patching file Makefile.fip
(Stripping trailing CRs from patch; use --binary to disable.)
patching file Makefile.mingw
Hunk #1 FAILED at 67 (different line endings).
1 out of 1 hunk FAILED -- saving rejects to file Makefile.mingw.rej
make[1]: *** [build-only-freeimage_x86_64-w64-mingw32.static] Error 1
```

With `--binary` enabled, there is another issue if `git` isn't set for `autocrlf = input`

```
Failed to build package freeimage for target i686-w64-mingw32.static!
patching file Makefile.fip
(Stripping trailing CRs from patch.)
patching file Makefile.mingw
Hunk #1 FAILED at 67.
1 out of 1 hunk FAILED -- saving rejects to file Makefile.mingw.rej
make[1]: *** [build-only-freeimage_i686-w64-mingw32.static] Error 1
```

The `postgresql` patch (unrelated to this PR) fails with `--binary` either way `git` is configured. There could be other unrelated packages failing similarly.

I'd say `--binary` should be removed and the `freeimage` patch corrected so it doesn't need it. Oddly, the patch already contains a section with `crlf` that applies correctly.

tonytheodore:
@tehnick see https://github.com/tehnick/mxe/pull/1. ([just the relevant commits without other merges from master](https://github.com/tonytheodore/mxe/compare/8b89d6b...ffc0656))

tonytheodore:
> there is another issue if git isn't set for autocrlf = input

Actually, that seems solved when the patches are created with `make export-patch-*`

tehnick:
Done.

tehnick:
> Actually, that seems solved when the patches are created with make export-patch-*

Could you write more details about this or give a link to documentation?

tonytheodore:
There are some patch helpers in https://github.com/mxe/mxe/blob/master/patch.mk. To fix the patch, I tried round-tripping it through those to see what happens first:
 - manually add `From 000000....` line so `git` sees it as a commit (using `vi` instead of normal editor which mangles patches)
 - `make init-git-freeimage` creates repo in gits/FreeImage directory
 - `make import-patch-freeimage` applies patch with `git am --keep-cr`
 - `make export-patch-freeimage` exports patchset with `git format-patch`

Voila - it now works!

tehnick:
@tonytheodore
Great tool! It should be more convenient than manually editing. I'll try to use it later.

Is pull request is ready for merging now?

tonytheodore:
> Is pull request is ready for merging now?

Yes, full build with old gcc succeeds and all the fixed packages work with gcc6.

Thanks!

PR #1329, opened 2016-05-04
proj4: libproj-0.dll is checked by GDAL if the platform is windows

rkanavath:
For mingw, proj4 creates libproj-9.dll and GDAL library consider
this as missing dll. GDAL has a workaround to use PROJSO env
variable that allows to set proj-9.dll. But it is better if
proj library could simply output the dll with the other name

https://trac.osgeo.org/gdal/browser/branches/2.1/gdal/ogr/ogrct.cpp#L86


tonytheodore:
How do the other downstream projects (libgeotiff openscenegraph ossim) handle this?

rkanavath:
ossim and openscenegraph does not depends on proj4 library. I don't how geotiff libraries will work with this patch.
do you want to propose the fix into gdal/ogr/ogrct.cpp ?

tonytheodore:
> I don't how geotiff libraries will work with this patch.

It is detected correctly, thanks.

PR #1333, opened 2016-05-04
mxe-conf. set WIN32 flag is missing in toolchain conf

rkanavath:
without set(WIN32 TRUE) in the toolchain conf. some builds are failing.

currently
if(WIN32)
<cross compile won't reach here >
endif()

if(UNIX)
<shouldn't have reached here. But it does anyway in cross compiling
endif()

changing first loop to:
if(WIN32 OR CMAKE_CROSSCOMPILING)
<cross compile reaches here >
endif()

### DIFF
--- a/src/mxe-conf.mk
+++ b/src/mxe-conf.mk
@@ -24,6 +24,7 @@ define $(PKG)_BUILD
     touch '$(CMAKE_TOOLCHAIN_DIR)/.gitkeep'
     (echo 'set(CMAKE_SYSTEM_NAME Windows)'; \
      echo 'set(MSYS 1)'; \
+     echo 'set(WIN32 TRUE)'; \
      echo 'set(BUILD_SHARED_LIBS $(if $(BUILD_SHARED),ON,OFF))'; \
      echo 'set(LIBTYPE $(if $(BUILD_SHARED),SHARED,STATIC))'; \
      echo 'set(CMAKE_PREFIX_PATH $(PREFIX)/$(TARGET))'; \


tonytheodore:
It should be [set automatically by cmake](https://cmake.org/cmake/help/latest/variable/WIN32.html). Do you have an example where this fails?

tonytheodore:
Can you check the output of:
```sh
cat > CMakeLists.txt <<EOF 
if(WIN32)
   message (win32)
endif()
if(UNIX)
   message (unix)
endif()
EOF

i686-w64-mingw32.static-cmake
```

it should be:
```
== Using MXE wrapper: /Users/tonyt/dev/mxe/usr/bin/i686-w64-mingw32.static-cmake
== Using MXE toolchain: /Users/tonyt/dev/mxe/usr/i686-w64-mingw32.static/share/cmake/mxe-conf.cmake
== Using MXE runresult: /Users/tonyt/dev/mxe/usr/share/cmake/modules/TryRunResults.cmake
loading initial cache file /Users/tonyt/dev/mxe/usr/share/cmake/modules/TryRunResults.cmake
-- The C compiler identification is GNU 4.9.3
-- The CXX compiler identification is GNU 4.9.3
-- Check for working C compiler: /Users/tonyt/dev/mxe/usr/bin/i686-w64-mingw32.static-gcc
-- Check for working C compiler: /Users/tonyt/dev/mxe/usr/bin/i686-w64-mingw32.static-gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Users/tonyt/dev/mxe/usr/bin/i686-w64-mingw32.static-g++
-- Check for working CXX compiler: /Users/tonyt/dev/mxe/usr/bin/i686-w64-mingw32.static-g++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
win32
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/tonyt/dev/mxe/tmp
```

rashadkm:
what is the MXE wrapper?
Using MXE wrapper: /Users/tonyt/dev/mxe/usr/bin/i686-w64-mingw32.static-cmake

rashadkm:
i am using cmake 3.4.0 downloaded from cmake download site. version 3.4.0,  Linux x86_64
But I have the correct CMAKE_TOOLCHAIN_FILE=/path/to/mxe-conf.cmake

https://cmake.org/cmake/help/latest/variable/WIN32.html
the page says "Set to true when the target system is Windows." . So I thought maybe it is neeed in the toolchain file.

tonytheodore:
> what is the MXE wrapper?

The wrapper script ensures you are using the same version of cmake and options that MXE is built and configured with. It's much the same as any prefixed tool (like *-pkg-config or even *-gcc), you could set those options manually but it would be very cumbersome and error prone.

> the page says "Set to true when the target system is Windows."

Probably a quirk of English, that means "cmake will set to true when the target system is Windows", not the literal reading which implies it should be set.

> So I thought maybe it is needed in the toolchain file.

If there's any doubt, just run a small test to observe the actual behaviour.

rkanavath:
i ran a small test and it passed only using if(WIN32 or CMAKE_CROSSCOMPILING)

but running now seems fine. maybe using latest upstream fixed the issue. Thanks.

PR #1335, opened 2016-05-05
openssl: update 1.0.2g --> 1.0.2h

tonytheodore:
See:
https://www.openssl.org/news/secadv/20160503.txt

tested on all four targets but not downstream packages


tonytheodore:
Already done: https://github.com/mxe/mxe/commit/945e76fc6ca3a8ab6acb6655749525fa43c4fca6

PR #1338, opened 2016-05-06
Fix cpp-netlib cmake directory.

TobiX:
For find_package(cppnetlib) to work, the directory for the cmake files
must be called the same as the files installed into the directory.

Installing directly into the cmake directory would also be an option...


starius:
Tested on i686-w64-mingw32.static.
Thanks!

Issue #1311, opened 2016-04-19
dll files in lib/


pavelvat:
After https://github.com/mxe/mxe/pull/1210 some dll files was built and installed to mxe/usr/i686-w64-mingw32.shared/lib/:
```
$ find /home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/lib/ -name *.dll
/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/lib/aspell-0.60/tex-filter.dll
/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/lib/aspell-0.60/texinfo-filter.dll
/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/lib/aspell-0.60/nroff-filter.dll
/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/lib/aspell-0.60/context-filter.dll
/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/lib/aspell-0.60/email-filter.dll
/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/lib/aspell-0.60/sgml-filter.dll
```

starius:
Is `bin/aspell-0.60/*.dll` the correct place or these files?

pavelvat:
I don't know.

TimothyGu:
I believe they are so-called "plugins," that instead of being linked to, are loaded at runtime. I believe the convention is to leave them in `lib/`.

pavelvat:
>I believe they are so-called "plugins," that instead of being linked to, are loaded at runtime. I believe the convention is to leave them in lib/.

Ok.

PR #1340, opened 2016-05-06
nsis: fix build for x86_64-w64-mingw32 targets

tehnick:
@starius @tonytheodore
I have began porting `nsis` pkg to `x86_64-w64-mingw32` targets.
Most of changes are trivial. But work is not finished yet.
I have stuck on this build failure:
http://www.susepaste.org/view/raw/29773209
Unfortunately, my knowledges of Assembler in a common and for x86 in particular are equal to zero.
Any advices?

Do we really need file `./usr/<target>/share/nsis/Bin/makensisw.exe`?
I have tested `make nsis -k -i` and almost all other files are successfully installed.
Only this binary file and `Contrib/Graphics` files are absent.

### DIFF
--- a/src/nsis.mk
+++ b/src/nsis.mk
@@ -26,7 +26,7 @@ define $(PKG)_BUILD
         `[ -d /usr/local/lib ]     && echo APPEND_LIBPATH=/usr/local/lib` \
         $(if $(findstring x86_64-w64-mingw32,$(TARGET)),\
             SKIPPLUGINS='System') \
-        SKIPUTILS='NSIS Menu' \
+        SKIPUTILS='MakeLangId,Makensisw,NSIS Menu,UIs,zip2exe' \
         NSIS_MAX_STRLEN=8192 \
         install
     $(INSTALL) -m755 '$(PREFIX)/$(TARGET)/bin/makensis' '$(PREFIX)/bin/$(TARGET)-makensis'
--- a/src/nsis.mk
+++ b/src/nsis.mk
@@ -17,15 +17,17 @@ define $(PKG)_UPDATE
 endef
 
 define $(PKG)_BUILD
+    $(if $(findstring x86_64-w64-mingw32,$(TARGET)),\
+        $(SED) -i 's/pei-i386/pei-x86-64/' '$(1)/SCons/Config/linker_script')
     cd '$(1)' && scons \
         MINGW_CROSS_PREFIX='$(TARGET)-' \
         PREFIX='$(PREFIX)/$(TARGET)' \
         `[ -d /usr/local/include ] && echo APPEND_CPPPATH=/usr/local/include` \
         `[ -d /usr/local/lib ]     && echo APPEND_LIBPATH=/usr/local/lib` \
+        $(if $(findstring x86_64-w64-mingw32,$(TARGET)),\
+            SKIPPLUGINS='System') \
         SKIPUTILS='NSIS Menu' \
         NSIS_MAX_STRLEN=8192 \
         install
     $(INSTALL) -m755 '$(PREFIX)/$(TARGET)/bin/makensis' '$(PREFIX)/bin/$(TARGET)-makensis'
 endef
-
-$(PKG)_BUILD_x86_64-w64-mingw32 =
--- a/src/nsis.mk
+++ b/src/nsis.mk
@@ -3,8 +3,8 @@
 
 PKG             := nsis
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 2.50
-$(PKG)_CHECKSUM := 3fb674cb75e0237ef6b7c9e8a8e8ce89504087a6932c5d2e26764d4220a89848
+$(PKG)_VERSION  := 2.51
+$(PKG)_CHECKSUM := 43d4c9209847e35eb6e2c7cd5a7586e1445374c056c2c7899e40a080e17a1be7
 $(PKG)_SUBDIR   := nsis-$($(PKG)_VERSION)-src
 $(PKG)_FILE     := nsis-$($(PKG)_VERSION)-src.tar.bz2
 $(PKG)_URL      := http://$(SOURCEFORGE_MIRROR)/project/nsis/NSIS 2/$($(PKG)_VERSION)/$($(PKG)_FILE)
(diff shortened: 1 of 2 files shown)


tehnick:
Now this pull request should be ready for merging.

I have tested builds for x86_64-w64-mingw32.static and i686-w64-mingw32.static targets with GCC 6.1.0 and 4.9.3.

tehnick:
Any comments?

src/nsis.mk
```diff
@@ -17,15 +17,16 @@ define $(PKG)_UPDATE
...
+    $(if $(findstring x86_64-w64-mingw32,$(TARGET)),\
+        $(SED) -i 's/pei-i386/pei-x86-64/' '$(1)/SCons/Config/linker_script')
     cd '$(1)' && scons \
         MINGW_CROSS_PREFIX='$(TARGET)-' \
         PREFIX='$(PREFIX)/$(TARGET)' \
         `[ -d /usr/local/include ] && echo APPEND_CPPPATH=/usr/local/include` \
         `[ -d /usr/local/lib ]     && echo APPEND_LIBPATH=/usr/local/lib` \
-        SKIPUTILS='NSIS Menu' \
+        SKIPPLUGINS='System' \
+        SKIPUTILS='NSIS Menu,Makensisw,zip2exe' \
```
TimothyGu: This takes away features originally included in the 32-bit build. Is it _very_ difficult to get them working on 64-bit? If so, please only exclude them on 64-bit builds.

tehnick:
@TimothyGu @tonytheodore @starius 
Could you explain why do we need `share/nsis/Bin/*.exe` and `share/nsis/Plugins/*.dll` at all? I do not see situations where they might be useful.
If user wants to launch nsis related executables in Wine (s)he may use programs from official web site. As far as I see only `usr/bin/<target>-makensis` files are used in MXE.

tonytheodore:
I know very little about `nsis` and only [just discovered it now builds on the latest versions of OSX](https://github.com/mxe/mxe/commit/cb06e059b23ef08f31eafd1dae1ea1417b102198).

In general, a (conditional) working subset is better than no support at all, but this package is something of a special case. If an installer couldn't display images (for example) I'd be very reluctant to use it. All 64-bit versions of windows still run 32-bit binaries so the very ugly workaround is to just use the 32-bit version to create installers.

Does the `nsis` created installer depend on the binaries under `share/nsis/*`? Does it somehow package those in the installer if special features are activated or are they simply unused artefacts of a cross-compile?

tonytheodore:
> Could you explain why do we need share/nsis/Bin/*.exe and share/nsis/Plugins/*.dll at all?

@pavelvat, do you have any insight on the features actually used from `nsis`? Are these binaries used at all in the deployed installer?

pavelvat:
For nsis package from Debian:
```
ls /usr/share/nsis/Bin/
RegTool.bin
``` 
so *.exe are not needed.
```
ls /usr/share/nsis/Plugins/
AdvSplash.dll  BgImage.dll  InstallOptions.dll  Math.dll       nsExec.dll  Splash.dll     System.dll   UserInfo.dll
Banner.dll     Dialer.dll   LangDLL.dll         nsDialogs.dll  NSISdl.dll  StartMenu.dll  TypeLib.dll  VPatch.dll
```
so *.dll is needed. But the path to the plug-ins is not clear: "mxe/usr/i686-w64-mingw32.shared/share/nsis/Plugins" or "mxe/usr/share/nsis/Plugins".

For example, EiskaltDC++ use additional ShellExecAsUser plug-in from http://nsis.sourceforge.net/ShellExecAsUser_plug-in.
https://github.com/eiskaltdcpp/eiskaltdcpp/blob/master/windows/EiskaltDC%2B%2B.nsi#L58-L59
https://github.com/eiskaltdcpp/eiskaltdcpp/blob/master/windows/EiskaltDC%2B%2B.nsi#L113-L115
(I had disabled it because nsis.mk not contain additional Plugins from http://nsis.sourceforge.net/Category:Plugins)
In order to ```makensis``` from Debian work correctly is required download ShellExecAsUser plug-in by link http://nsis.sourceforge.net/mediawiki/images/c/c7/ShellExecAsUser.zip and unpack file ShellExecAsUser.dll to /usr/share/nsis/Plugins/. Then command ```makensis EiskaltDC++.nsi``` will build ShellExecAsUser.dll inside the resulting installer file.

pavelvat:
>But the path to the plug-ins is not clear: "mxe/usr/i686-w64-mingw32.shared/share/nsis/Plugins" or "mxe/usr/share/nsis/Plugins".

I have tested plug-in ShellExecAsUser.dll with EiskaltDC++ from https://github.com/mxe/mxe/pull/1165 for path "mxe/usr/i686-w64-mingw32.shared/share/nsis/Plugins" it has worked correctly. I had also updated https://github.com/mxe/mxe/issues/1311 (problem only occurs with aspell).

tehnick:
> Does the `nsis` created installer depend on the binaries under `share/nsis/*`? Does it somehow package those in the installer if special features are activated or are they simply unused artefacts of a cross-compile?

I have not looked on this yet due to lack of time. I'll try to find time at this weekend.

> so *.exe are not needed.

Great! So we do not need makensisw.exe and zip2exe.exe.
More over: we may disable MakeLangId.exe

> so *.dll is needed.

So only `System.dll` is questionable now.
(Its build is broken due Assembler usage as I wrote above.)

> But the path to the plug-ins is not clear: "mxe/usr/i686-w64-mingw32.shared/share/nsis/Plugins" or "mxe/usr/share/nsis/Plugins".

From my point of view it should be `usr/<target>/share/nsis/Plugins/`.

> I had disabled it because nsis.mk not contain additional Plugins from http://nsis.sourceforge.net/Category:Plugins

We may add `nsis-<plugin-name>` packages for necessary plugins if they are FOSS.

pavelvat:
@tehnick Why nsis-2.50? nsis-2.51 is available https://sourceforge.net/projects/nsis/.

tehnick:
> Why nsis-2.50? nsis-2.51 is available

Because even before my changes patch for `nsis` was quite big and it could not be applied to `nsis-2.51` as is. For newer `nsis` release this patch should be updated and it requires additional time...

tehnick:
Ok. I have updated `nsis` to newer version.
Is this PR ready for merging now?

tehnick:
Now rebased.

tonytheodore:
Thanks! Builds on all four targets with both gcc 6.1.0 and 4.9.3.

PR #1345, opened 2016-05-10
gdal: install gdal.pc and gdal-data

rkanavath:


tonytheodore:
The installed `gdal.pc` doesn't have any dependency libs:

```
$ i686-w64-mingw32.static-pkg-config --libs gdal
-L/Users/tonyt/dev/mxe/usr/i686-w64-mingw32.static/lib -lgdal
```

You may be better off using `*-gdal-config --dep-libs` but this PR looks okay otherwise.

Thanks!

rkanavath:
Thanks.

PR #1348, opened 2016-05-17
openthreads: build in  src/OpenThreads

rkanavath:
openthreads should not do rebuild of osg. it should only build src/OpenThreads

### DIFF
--- a/src/openthreads.mk
+++ b/src/openthreads.mk
@@ -25,5 +25,5 @@ define $(PKG)_BUILD
         -D_OPENTHREADS_ATOMIC_USE_WIN32_INTERLOCKED=1 \
         '$(1)'
 
-    $(MAKE) -C '$(1).build' -j '$(JOBS)' install VERBOSE=1
+    $(MAKE) -C '$(1).build/src/OpenThreads' -j '$(JOBS)' install VERBOSE=1
 endef


tonytheodore:
Are you sure osg is built? The patch adding `OPENTHREADS_STANDALONE` option skips the other subdirectories - at least it looks like it's supposed to.

rkanavath:
i didn't had the patch when tried in my fork. This is why it failed. 

Sorry for the noise. BTW, if building inside src/OpenThreads/ could remove the patch then is that okay?
I am not against keeping patches, but if there is an easy way then i wish.

tonytheodore:
Removing the patch is good, though I suspect it also suppresses detection of other required dependencies of osg. Make sure you test with a clean build and no other packages installed.

rkanavath:
okay. i will test from a fresh checkout and report back.

rkanavath:
suppress in dependencies is good. but I don't think changing build directory would do that. It still goes detect all osg dependencies. Can we override the dependencies in openthreads.mk. It only requires gcc.

tonytheodore:
The dependencies are not hard requirements so it simply bypasses them. Patch removed in https://github.com/mxe/mxe/commit/d57e22f049c7153cdcf256013056b53574501918.

PR #1350, opened 2016-05-18
glm: add rules to download/build version 0.9.7.4

yipdw:
This pull request contains build rules for glm 0.9.7.4 (http://glm.g-truc.net/0.9.7/index.html).

glm is a header-only library, so it can be vendored in projects that use it; however, some libraries (e.g. CEGUI 0.8.7) expect to find it in the build environment.

The test program should generate the output

```
mat4x4((0.672222, 0.000000, 0.000000, 0.000000), (0.000000, 0.896295, 0.000000, 0.000000), (0.000000, 0.000000, -0.501001, -0.500000), (0.000000, 0.000000, -0.200200, 0.000000))
```

### DIFF
--- a/src/glm-test.cpp
+++ b/src/glm-test.cpp
@@ -1,0 +1,32 @@
+/*
+ * This file is part of MXE.
+ * See index.html for further information.
+ *
+ * This file is a test program for glm, adapted from the code sample at
+ * http://glm.g-truc.net/0.9.7/index.html.
+ */
+
+#include <glm/gtc/matrix_transform.hpp> // glm::translate, glm::rotate, glm::scale, glm::perspective
+#include <glm/mat4x4.hpp> // glm::mat4
+#include <glm/vec3.hpp> // glm::vec3
+#include <glm/vec4.hpp> // glm::vec4
+#include <glm/gtx/string_cast.hpp> // so we can print a calculation result
+#include <iostream>
+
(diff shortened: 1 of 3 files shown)


yipdw:
I've updated the build rules to use `$(TARGET)-cmake`, which seems like a more modern alternative to `cmake -DCMAKE_TOOLCHAIN_PATH=...`.

src/glm-test.cpp
```diff
@@ -0,0 +1,32 @@
...
+    return Projection * View * Model;
+}
+
+int main()
+{
+    glm::mat4 m = camera(0.0f, glm::vec2(0.0f, 0.0f));
+    std::cout << glm::to_string(m) << std::endl;
+
+    return 0;
+}
```
yipdw: GLM also has an official test suite, but it is not installed as part of `make install`: it's designed to be run via `make test`.  I suppose this could be integrated into the build file, but I am unsure how to do so: we cannot guarantee the existence of e.g. WINE on systems.

It may be possible to install `cppcheck` and the GLM test harness into the target `bin` directory, which will let the user run the GLM test suite for themselves.

tonytheodore:
We specifically disable `wine` during builds and don't generally build test harnesses for most packages. A test that linking works is usually sufficient.

Looks good, thanks!

Issue #1364, opened 2016-05-26
Documentation for usage of plugins


rr-:
The instructions mentioned in PR #935 should be probably included also in https://github.com/mxe/mxe/blob/master/plugins/README.md - at very least, the info on `MXE_PLUGIN_DIRS` variable.

tonytheodore:
Fixed in #1418.

PR #1366, opened 2016-05-26
docs: add mini-tutorial on plugins

rr-:
Proposed fix for #1364 

This probably should make its way in one form or another to `gh-pages` branch, as well as `MXE_VERBOSE` variable which is extremely useful.


tonytheodore:
Thanks, it's a good starting point.

PR #1367, opened 2016-05-28
Allow the host version of gettext to be compiled on systems with older versions of libxml2

dgburr:
gettext requires libxml2 >= 2.9.1.  However some distros provide an older version of this library (e.g. Debian 7.10 includes libxml 2.8.0).  This patch is based upon the fix described in https://savannah.gnu.org/bugs/?46844 which has now been committed upstream and will be included in the next release of gettext.

### DIFF
--- a/src/gettext-1-libxml-check.patch
+++ b/src/gettext-1-libxml-check.patch
@@ -1,0 +1,99 @@
+This file is part of MXE.
+See index.html for further information.
+
+This patch allows gettext to be compiled on systems which have a version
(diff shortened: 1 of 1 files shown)


tonytheodore:
Mmm, cross builds of gettext don't require libxml2, would a similar `--without-libxml2-prefix` option be more appropriate?

dgburr:
Hi @tonytheodore,

Why not use `--with-included-libxml` instead?

dgburr:
Created pull request #1368 with the alternative approach discussed here

PR #1368, opened 2016-05-29
Always use the included version of libxml

dgburr:
Alternative fix to mxe/mxe#1367 - always use the version of libxml2 included in gettext instead of trying to use the system version


tonytheodore:
Thanks!

PR #1370, opened 2016-05-30
[WIP] qbittorrent: fix qmake config for win32-g++

tonytheodore:
May only be an OSX issue or something to do with my setup.

See https://github.com/tonytheodore/mxe/pull/7#issuecomment-222419989

PR #1371, opened 2016-06-01
qt: symlink mkspecs/default

tonytheodore:
On OSX, the default mkspec doesn't set `win32-g++` correctly.
The following workarounds are possible:
 - set `QMAKESPEC` env var
 - use `-spec` command line option
 - set `CONFIG+=`
 - symlink mkspecs/default

The internal build of qt always seems to use the `-spec` option,
but that doesn't solve the default case - the symlink approach used
on linux appears to always work.

closes #1370

### DIFF
--- a/src/qt.mk
+++ b/src/qt.mk
@@ -73,6 +73,13 @@ define $(PKG)_BUILD
     $(MAKE) -C '$(1)' -j 1 install
     ln -sf '$(PREFIX)/$(TARGET)/qt/bin/qmake' '$(PREFIX)/bin/$(TARGET)'-qmake-qt4
 
+    # symlink mkspecs/default if it isn't already
+    # required on OSX to mimic linux installation
+    [[ -L  '$(PREFIX)/$(TARGET)/qt/mkspecs/default' ]] || \
+    rm -rf '$(PREFIX)/$(TARGET)/qt/mkspecs/default' && \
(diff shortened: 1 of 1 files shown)


tonytheodore:
Commit message updated.

src/qt.mk
```diff
@@ -73,6 +73,13 @@ define $(PKG)_BUILD
     $(MAKE) -C '$(1)' -j 1 install
     ln -sf '$(PREFIX)/$(TARGET)/qt/bin/qmake' '$(PREFIX)/bin/$(TARGET)'-qmake-qt4
 
+    # symlink mkspecs/default if it isn't already
+    # required on OSX to mimic linux installation
+    [[ -L  '$(PREFIX)/$(TARGET)/qt/mkspecs/default' ]] || \
+    rm -rf '$(PREFIX)/$(TARGET)/qt/mkspecs/default' && \
+    ln -s  '$(PREFIX)/$(TARGET)/qt/mkspecs/win32-g++-4.6' \
```
tehnick: Why not just `ln -sf`?

tonytheodore: `ln -sf` won't replace an existing directory. It probably doesn't need to check if it's a already a link though.

tehnick: Hmm, really. I was inattentive.

Issue #1259, opened 2016-03-16
Ensure plugins work with all make targets


tonytheodore:
From #1234:
- cleanup-style
- cleanup-deps-style
- versions.json

(see also #1015)

tonytheodore:
Also the automatic `plugins/native/$(OS_SHORT_NAME)`

tonytheodore:
ongoing

PR #1372, opened 2016-06-02
fix native plugins when additional dirs are specified

tonytheodore:
When plugins are specified on the command line, automatic native plugins
are not included (currently for darwin and wheezy):

```
$ make gmsl-print-MXE_PLUGIN_DIRS MXE_PLUGIN_DIRS=plugins/apps/
MXE_PLUGIN_DIRS = plugins/apps/
```

See:
https://www.gnu.org/software/make/manual/make.html#Override-Directive
https://github.com/mxe/mxe/issues/1259#issuecomment-217376756


starius:
Looks good to me.

PR #1375, opened 2016-06-05
Add plugin file to compile some libraries with Qt 5 only

saiarcot895:
Poppler and OSG can be built with Qt 5 instead of Qt 4 when using the plugin file.


starius:
Thank you for the plugin! Can you fix the following issues, please?

1. You don't have to specify `poppler_BUILD_SHARED`, as it is already defined in `src/poppler.mk`
2. How does `openscenegraph` choose between qt4 and qt5? What happens if both qt4 and qt5 are available?
3. Can you avoid/minimize code duplications in `poppler_BUILD` section using a variable or a macro? The only difference is `--enable-poppler-qt5`.

saiarcot895:
Done (for all three), and I added shared build support for `poppler` as well, since it was a matter of using a variable in there. I also simplified the override file so that it only removes `qt` from the build dependencies and adds `qtbase`, so the main build dependencies can be changed without the override file breaking (hopefully).

For both `poppler` and `openscenegraph`, I'm checking to see if `qtbase` is in the build dependencies, and if so, use qt5. If the override plugin is in effect, then qt5 will always be used (even if qt4 and qt5 are installed).

`openscenegraph` has a CMake variable that lets you choose the desired qt version, which I'm setting. Sadly, I don't think it will error out if it doesn't find the qt libraries for some reason.

starius:
Looks good to me. Not tested yet.

> Sadly, I don't think it will error out if it doesn't find the qt libraries for some reason.

You can test it yourself: add a test which checks output of `$(TARGET)-nm libosgQt.a | grep something-qt4-or-qt4-specific`

saiarcot895:
Updated. I took the easy way out and added the `REQUIRED` flag to the cmake build file.

saiarcot895:
Verified it compiles correctly on `i686-w64-mingw32.shared`.

starius:
I tried to build with and without the plugin in clean clones. Both builds are broken.

Without the plugin. Command `make -j 1 JOBS=3 poppler openscenegraph`. Package failed: poppler. Log: https://gist.github.com/b51e9948595931479ecb2eed8f04f1be

With the plugin. Command `make -j 1 JOBS=3 poppler openscenegraph MXE_PLUGIN_DIRS=plugins/qt5-deps/`. Package failed: icu4c. Log: https://gist.github.com/0a86a18f5ada466071e10dd5314503ae File `tmp-icu4c-i686-w64-mingw32.static/icu/source/configure`: https://gist.github.com/42a0ff21fad359ec127d5c350b1cfd33 (Can be unrelated.)

saiarcot895:
Fixed the first error (forgot to dereference a variable). As for the second error, I have no idea why `icu4c` is failing. I'll do a test build tonight.

tonytheodore:
`icu` error looks like an `autotools` issue and very likely fixed by #1372.

saiarcot895:
Confirmed `poppler` and `openscenegraph` compile for `i686-w64-mingw32.static` after making another fix to the makefiles.

starius:
As of a3c8d4b5eb194efe7c8d060923d0cbad51e80ebc + https://github.com/mxe/mxe/pull/1372, both builds are still broken:

  * without the plugin, poppler: https://gist.github.com/052f28577bfa6358a184086ada6db1f9
  * with the plugin, openthreads: https://gist.github.com/176751407e0ea6dfb3c841435d335a8f

src/poppler.mk
```diff
@@ -24,14 +24,14 @@ define $(PKG)_BUILD
...
         --disable-silent-rules \
-        --disable-shared \
-        --enable-static \
         --enable-xpdf-headers \
-        --enable-poppler-qt4 \
+        $(if $(filter qtbase,$($(PKG)_DEPS)), \
+          --enable-poppler-qt5 \
+          --disable-poppler-qt4, \
+          --disable-poppler-qt4 \
+          --enable-poppler-qt5) \
```
tonytheodore: those enable/disable options are the same.

saiarcot895: Fixed. Hopefully, it's right now.

saiarcot895:
Regarding openthreads, in `openthreads.mk`, the only dependency listed is `gcc`. However, the CMake file also checks to see if poppler is installed and, if so, runs a compile test.

From a clean build, that error wouldn't occur because openthreads would be compiled early on, before poppler is compiled, and so that compile test wouldn't run. However, when openthreads is compiled after poppler is compiled, the compile test would occur.

openscenegraph sets the `POPPLER_HAS_CAIRO_EXITCODE` variable to 0, which is what I've added in into the openthreads makefile.

starius:
As of 8e58fb9363c910757c299163cc9fd019a73b47bf, openthreads is broken in the build without the plugin: https://gist.github.com/815ba3d8b43be5276c86e0fb0d4197ac

tonytheodore:
For openthreads, I'd suggest reverting https://github.com/mxe/mxe/commit/d57e22f049c7153cdcf256013056b53574501918. I only tested with a clean build that the dependencies weren't needed, not the effect of them being present.

saiarcot895:
The openthreads fix is in the commit after the one tested, 5b52ff4bc16bdc2e88f7e85338a7cfbffe918b43.

With the addition of `POPPLER_HAS_CAIRO_EXITCODE`, I don't think that commit needs to be reverted. I just did a build of openthreads with some other openscenegraph dependencies present, and it compiled.

starius:
5b52ff4bc16bdc2e88f7e85338a7cfbffe918b43 + revert of d57e22f049c7153cdcf256013056b53574501918 + #1372 = builds on all 4 targets with and without the plugin.

starius:
5b52ff4bc16bdc2e88f7e85338a7cfbffe918b43 + #1372 = builds on all 4 targets with and without the plugin.

revert of d57e22f049c7153cdcf256013056b53574501918 is not needed for build.

saiarcot895:
Anything else I need to do for this pull request?

starius:
Looks good to me! Thanks!

PR #1376, opened 2016-06-05
Fix hyperscan and update to 4.2.0

starius:
I have tested `i686-w64-mingw32.static` in Windows XP in VirtualBox after disabling AXV2 instructions by specifying `-DCMAKE_C_FLAGS="-march=core2" -DCMAKE_CXX_FLAGS="-march=core2"` to cmake.

Could someone test it in native Windows on machine with AVX2 support, please?

To test, run `test-hyperscan.exe some-pattern some-file`, e.g. `test-hyperscan.exe a C:\boot.ini` and make sure that there are some hits:

```
Match for pattern "a" at offset 57
Match for pattern "a" at offset 69
Match for pattern "a" at offset 82
Match for pattern "a" at offset 93
...
```

To make sure everything works, full test suite can be used. It is not installed to `bin` directory, but it can be extracted from build directory (`tmp-<target>/hyperscan-4.2.0.build/bin/unit-hyperscan.exe`). It requires file `./unit/hyperscan/bad_patterns.txt` which can be copied from source directory. I created a [pack](http://redjohn.tk/mxe/hyperscan-test.zip) of unit tests for all targets.

**UPD**. Tests pass on native Windows.

### DIFF
--- a/src/hyperscan-1-fixes.patch
+++ b/src/hyperscan-1-fixes.patch
@@ -55,22 +55,22 @@ index 1111111..2222222 100644
  SET(hs_HEADERS
      src/hs.h
      src/hs_common.h
-@@ -935,7 +927,6 @@ endif()
+@@ -979,15 +971,12 @@ endif()
  # we want the static lib for testing
  add_library(hs STATIC ${hs_SRCS} $<TARGET_OBJECTS:hs_exec>)
  
 -add_dependencies(hs ragel_Parser)
- add_dependencies(hs autogen_compiler autogen_teddy_compiler)
- 
+-
  if (NOT BUILD_SHARED_LIBS)
-@@ -944,7 +935,6 @@ endif()
+ install(TARGETS hs DESTINATION lib)
+ endif()
  
  if (BUILD_STATIC_AND_SHARED OR BUILD_SHARED_LIBS)
      add_library(hs_shared SHARED ${hs_SRCS} $<TARGET_OBJECTS:hs_exec_shared>)
 -    add_dependencies(hs_shared ragel_Parser)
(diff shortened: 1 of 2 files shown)


tonytheodore:
Tests from the zip file pass on [Win10 x64 in VMWare that appears to have AVX2 support](https://gist.github.com/tonytheodore/b907e2757592878c67f7ad3fba03b608#file-cpuz-win10-64-vmware-L51-L64).

src/hyperscan-1-fixes.patch
```diff
@@ -5998,6 +5959,19 @@ index 1111111..2222222 100644
...
+diff --git a/src/nfa/mcclellan_common_impl.h b/src/nfa/mcclellan_common_impl.h
+index 1111111..2222222 100644
+--- a/src/nfa/mcclellan_common_impl.h
++++ b/src/nfa/mcclellan_common_impl.h
+@@ -26,7 +26,7 @@
+  * POSSIBILITY OF SUCH DAMAGE.
+  */
+ 
+-#if defined(__INTEL_COMPILER) || defined(__clang__) || defined(_WIN32) || defined(__GNUC__) && (__GNUC__ < 4)
++#if defined(__INTEL_COMPILER) || defined(__clang__) || NATIVE_WIN32 || defined(__GNUC__) && (__GNUC__ < 4)
```
tonytheodore: What is `NATIVE_WIN32`? Should that be [`defined(_MSC_VER)`](https://sourceforge.net/p/predef/wiki/Compilers/#microsoft-visual-c)?

tonytheodore: Ignore that, I see it's defined earlier in the patch.

tonytheodore:
Looks good, thanks!

PR #1378, opened 2016-06-08
Add plugin for compiling mingw-w64 with secure APIs enabled.

saiarcot895:
This is based off of #897.

### DIFF
--- a/src/gcc.mk
+++ b/src/gcc.mk
@@ -76,7 +76,8 @@ define $(PKG)_BUILD_mingw-w64
         --host='$(TARGET)' \
         --prefix='$(PREFIX)/$(TARGET)' \
         --enable-sdk=all \
-        --enable-idl
+        --enable-idl \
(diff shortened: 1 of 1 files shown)


tonytheodore:
A plugin is probably overkill for this case, you could just set the variable directly, i.e
```
make mingw-w64_SEC_API=1
# instead of
make MXE_PLUGIN_DIRS=...
```

I'd opt for something more general though, adding `$(mingw-w64-headers_CONFIGURE_OPTS)` to the `configure` invocation and calling with:

```
make mingw-w64-headers_CONFIGURE_OPTS=--enable-secure-api
```

saiarcot895:
Updated implementation. This variable can then be set from settings.mk, right?

tonytheodore:
> This variable can then be set from settings.mk, right?

Yes, from settings.mk or command line.

Looks good, can you rebase into a single commit?

saiarcot895:
Done

tonytheodore:
Thanks!

PR #1380, opened 2016-06-10
libjpeg-turbo: update from 1.4.1 to 1.5.0

tehnick:
Drop unneeded patches.


tehnick:
I have just rebased pull request onto current master.

Any comments?

tonytheodore:
Builds successfully on all four targets, thanks!

PR #1382, opened 2016-06-10
New package: ghostscript

darealshinji:
Since the upstream developers don't seem to care much about supporting MinGW, and especially not about cross-compiling, and since they use a kind of ugly build system, a lot of stuff needed to be patched.
The patch is based on the patches from the [MinGW package](https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-ghostscript), plus it enables cross-compiling.
If this gets merged, [libspectre](https://github.com/darealshinji/mxe/tree/libspectre-new) could be added too, which is a wrapper library for ghostscript.

### DIFF
--- a/src/ghostscript.mk
+++ b/src/ghostscript.mk
@@ -1,0 +1,76 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := ghostscript
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 9.19
+$(PKG)_NODOTVER := $(subst .,,$($(PKG)_VERSION))
+$(PKG)_CHECKSUM := f67acdcfcde1f86757ff3553cd719f12eac2d7681a0e96d8bdd1f40a0f47b45b
+$(PKG)_SUBDIR   := $(PKG)-$($(PKG)_VERSION)
+$(PKG)_FILE     := $(PKG)-$($(PKG)_VERSION).tar.bz2
+$(PKG)_URL      := https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs$($(PKG)_NODOTVER)/$($(PKG)_FILE)
+$(PKG)_DEPS     := gcc dbus fontconfig freetype lcms libiconv libidn libjpeg-turbo libpaper libpng openjpeg tiff zlib
+
+define $(PKG)_UPDATE
+    $(WGET) -q -O- 'http://ghostscript.com/Releases.html' | \
+    $(SED) -n 's:.*GPL_Ghostscript_::p' | \
+    $(SED) -n 's:\.html.*::p'
+endef
+
+define $(PKG)_BUILD
+    cd '$(SOURCE_DIR)' && rm -rf freetype jpeg lcms2 libpng openjpeg tiff zlib
+    cd '$(SOURCE_DIR)' && $(LIBTOOLIZE) --force --copy --install
+    cd '$(SOURCE_DIR)' && autoconf -f -i
+    cd '$(BUILD_DIR)' && $(SOURCE_DIR)/configure \
+        $(MXE_CONFIGURE_OPTS) \
+        --disable-contrib \
+        --enable-threading \
+        --enable-fontconfig \
+        --enable-dbus \
+        --enable-freetype \
+        --disable-cups \
+        --enable-openjpeg \
+        --disable-gtk \
+        --with-libiconv=gnu \
+        --with-libidn \
+        --with-libpaper \
+        --with-system-libtiff \
+        --with-ijs \
+        --with-luratech \
+        --with-jbig2dec \
+        --with-omni \
+        --without-x \
+        --with-drivers=ALL \
+        --with-memory-alignment=$(if $(filter x86_64-%,$(TARGET)),8,4)
+    $(MAKE) -C '$(BUILD_DIR)' -j 1 $(if $(BUILD_STATIC),gs.a,so)
+
+    $(INSTALL) -d '$(PREFIX)/$(TARGET)/include/ghostscript'
+    $(INSTALL) '$(SOURCE_DIR)/devices/gdevdsp.h' '$(PREFIX)/$(TARGET)/include/ghostscript/gdevdsp.h'
+    $(INSTALL) '$(SOURCE_DIR)/base/gserrors.h' '$(PREFIX)/$(TARGET)/include/ghostscript/gserrors.h'
+    $(INSTALL) '$(SOURCE_DIR)/psi/iapi.h' '$(PREFIX)/$(TARGET)/include/ghostscript/iapi.h'
+    $(INSTALL) '$(SOURCE_DIR)/psi/ierrors.h' '$(PREFIX)/$(TARGET)/include/ghostscript/ierrors.h'
+
+    $(INSTALL) -d '$(PREFIX)/$(TARGET)/bin'
+    $(INSTALL) -d '$(PREFIX)/$(TARGET)/lib'
+    $(if $(BUILD_STATIC),\
+        $(INSTALL) '$(BUILD_DIR)/gs.a' '$(PREFIX)/$(TARGET)/lib/libgs.a',\
+        $(INSTALL) '$(BUILD_DIR)/sobin/libgs-9.dll' '$(PREFIX)/$(TARGET)/bin/libgs-9.dll' && \
+        $(INSTALL) '$(BUILD_DIR)/sobin/libgs.dll.a' '$(PREFIX)/$(TARGET)/lib/libgs.dll.a')
+
+    $(INSTALL) -d '$(PREFIX)/$(TARGET)/lib/pkgconfig'
+    (echo 'Name: ghostscript'; \
+     echo 'Version: $($(PKG)_VERSION)'; \
+     echo 'Description: Ghostscript library'; \
+     echo 'Cflags: -I"$(PREFIX)/$(TARGET)/include/ghostscript"'; \
+     echo 'Cflags.private: -DGS_STATIC_LIB'; \
+     echo 'Libs: -L"$(PREFIX)/$(TARGET)/lib" -lgs'; \
+     echo 'Requires: libidn libtiff-4 libpng jpeg lcms2 zlib'; \
+     echo '# https://github.com/mxe/mxe/issues/1446'; \
+     echo 'Libs.private: -lm -liconv -lpaper -lopenjp2 -lwinspool';) \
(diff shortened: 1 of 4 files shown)


darealshinji:
Rebased commits into a single one.

src/ghostscript.mk
```diff
@@ -0,0 +1,74 @@
...
+     echo 'Version: $($(PKG)_VERSION)'; \
+     echo 'Description: Ghostscript library'; \
+     echo 'Cflags: -I"$(PREFIX)/$(TARGET)/include/ghostscript"'; \
+     echo 'Libs: -L"$(PREFIX)/$(TARGET)/lib" -lgs'; \
+     echo 'Libs.private: -lm -lidn -liconv -lpaper -ltiff -lpng -ljpeg -lopenjp2 -llcms2 -lz -lwinspool';) \
+     > '$(PREFIX)/$(TARGET)/lib/pkgconfig/ghostscript.pc'
+
+    '$(TARGET)-gcc' \
+        -W -Wall -Werror -pedantic `$(TARGET)-pkg-config --cflags ghostscript` \
+        $(if $(BUILD_STATIC),-DGS_STATIC_LIB, ) \
```
tonytheodore: Move `-DGS_STATIC_LIB` to `Cflags.private:` section of `ghostscript.pc` so it's available to downstream packages. 
Use single invocation of `$(TARGET)-pkg-config --cflags --libs ghostscript`

tonytheodore:
Looking good, builds on all four targets.

src/ghostscript.mk
darealshinji: Done.

src/ghostscript.mk
```diff
@@ -0,0 +1,73 @@
...
+    $(WGET) -q -O- 'http://ghostscript.com/Releases.html' | \
+    $(SED) -n 's:.*GPL_Ghostscript_::p' | \
+    $(SED) -n 's:\.html.*::p'
+endef
+
+define $(PKG)_BUILD
+    cd '$(1)' && rm -rf freetype jpeg lcms2 libpng openjpeg tiff zlib
+    cd '$(1)' && $(LIBTOOLIZE) --force --copy --install
+    cd '$(1)' && autoconf -f -i
+    cd '$(1)' && ./configure $(MXE_CONFIGURE_OPTS) \
```
tonytheodore: put `$(MXE_CONFIGURE_OPTS)` on its own line

src/ghostscript.mk
```diff
@@ -0,0 +1,73 @@
...
+     echo 'Version: $($(PKG)_VERSION)'; \
+     echo 'Description: Ghostscript library'; \
+     echo 'Cflags: -I"$(PREFIX)/$(TARGET)/include/ghostscript"'; \
+     echo 'Cflags.private: -DGS_STATIC_LIB'; \
+     echo 'Libs: -L"$(PREFIX)/$(TARGET)/lib" -lgs'; \
+     echo 'Libs.private: -lm -lidn -liconv -lpaper -ltiff -lpng -ljpeg -lopenjp2 -llcms2 -lz -lwinspool';) \
+     > '$(PREFIX)/$(TARGET)/lib/pkgconfig/ghostscript.pc'
+
+    '$(TARGET)-gcc' \
+        -W -Wall -Werror -pedantic $(if $(BUILD_STATIC),-DGS_STATIC_LIB, ) \
```
tonytheodore: Remove `$(if $(BUILD_STATIC),-DGS_STATIC_LIB, )`, it's in the *pc file

starius:
> echo 'Libs.private: -lm -lidn -liconv -lpaper -ltiff -lpng -ljpeg -lopenjp2 -llcms2 -lz -lwinspool';) \ 

Can you replace it with `Requires:`? MXE has most (or all) of the libraries from this list.

starius:
Can you format the patch with `patch-tool-mxe`, please?

IMHO, if a patch is based on another (external) patch, it is better to keep it as two "commits" in a patch file. When the original patch updates, it will be easy to "rebase" changes against new version of upstream patch. I have separated your changes from the original patch. My patch file includes patches by MINGW-packages followed by further changes found in your initial patch.
https://gist.github.com/7f14838b3487cf8f282f03520b92a0a8

darealshinji:
I've updated ghostscript.mk and added your patch file.


>> echo 'Libs.private: -lm -lidn -liconv -lpaper -ltiff -lpng -ljpeg -lopenjp2 -llcms2 -lz -lwinspool';) \

>Can you replace it with Requires:? MXE has most (or all) of the libraries from this list.

Static linking against libgs will fail if I remove the `Libs.private` line. MXE has all these libraries but not all of them have a pkg-config file.

starius:
> but not all of them have a pkg-config file

Which ones do not have pc-files?
Can you use `Requires` for those who have, please?

starius:
Why do you build with `-j 1` instead of `-j $(JOBS)`?

starius:
I changed your .mk file to build in a separate dir and use new macros `SOURCE_DIR`, `BUILD_DIR` and `TEST_FILE`: https://gist.github.com/c2d4a6346eae98bdcbb6e0bcfadeab8a

darealshinji:
> Which ones do not have pc-files?

libm, libiconv, libpaper, libopenjp2 and libwinspool.

> Can you use Requires for those who have, please?

Like this? -> `Requires: libidn libtiff-4 libpng jpeg lcms2 zlib`

> Why do you build with -j 1 instead of -j $(JOBS)?

The build system doesn't seem to support more than 1 jobs. It'll eventually break at some point during the build.

darealshinji:
Building with `-j $(JOBS)` works on all targets. Still the problem with `BUILD_DIR`.

starius:
> Like this? -> Requires: libidn libtiff-4 libpng jpeg lcms2 zlib

Yes.

Also add a comment with https://github.com/mxe/mxe/issues/1446 near libraries linked directly through Libs.private, please.

> The build system doesn't seem to support more than 1 jobs. It'll eventually break at some point during the build.

Please report this to upstream and add link to the bug report to comment.

> Still the problem with BUILD_DIR.

What problems?

darealshinji:
> Also add a comment with #1446 near libraries linked directly through Libs.private, please.

Is this correct?
``` sh
     echo 'Requires: libidn libtiff-4 libpng jpeg lcms2 zlib'; \
     echo '# https://github.com/mxe/mxe/issues/1446'; \
     echo 'Libs.private: -lm -liconv -lpaper -lopenjp2 -lwinspool';) \
```

> Please report this to upstream and add link to the bug report to comment.

I think this only happened to me on the previous version of ghostscript. This isn't a problem anymore.

> What problems?

Out-of-tree building isn't supported:
```
[build]     ghostscript              i686-w64-mingw32.static                                              

Failed to build package ghostscript for target i686-w64-mingw32.static!
./obj/aux_/echogs -e .dev -w- -l-dev2 -b -s -l-include -l./obj/page -l-obj ./obj/jpeggray ./obj/gdevjpeg.o
./obj/aux_/echogs -e .dev -a-  ./obj/jpeggray -include ./obj/sdcte
make[2]: *** No rule to make target `obj/mag16.dev', needed by `obj/devdevs.tr'.  Stop.
make[2]: Leaving directory `/home/djcj/mxe/tmp-ghostscript-i686-w64-mingw32.static/ghostscript-9.19.build_'
make[1]: *** [build-only-ghostscript_i686-w64-mingw32.static] Error 2
make[1]: Leaving directory `/home/djcj/mxe'
real	0m34.271s
user	0m29.100s
sys	0m1.688s
[log]      /home/djcj/mxe/log/ghostscript_i686-w64-mingw32.static

make: *** [/home/djcj/mxe/usr/i686-w64-mingw32.static/installed/ghostscript] Error 1
```

But it works if I replace the `BUILD_DIR` macros with `SOURCE_DIR`.

starius:
> Shouldn't it be Requires.private?

IMHO, `*.private` is for dependencies not exposed to API. So the answer depends on whether these dependencies are exposed to API of ghostscript.

> Out-of-tree building isn't supported:

Please report this to upstream. I think, we can use in-tree building and provide a link to the report.

darealshinji:
I've tested a native Linux build on the unpatched source. The out-of-tree problem only occurs when the source was configured with `--disable-contrib`.

starius:
I am digging into this issue.

Other issues. I have found the following messages in the log:

```
configure: WARNING: unrecognized options: --enable-static, --disable-shared, --disable-doxygen
```

It doesn't understand static/shared switches.

```
configure: WARNING: Using "native" font scaler which is now deprecated (rather than freetype),
configure: WARNING: Support for this will be removed in a future release
```

Does it mean that it doesn't use freetype? Please, make sure that freetype provided by MXE is used.

starius:
The problem with out-of-tree build was fixed in https://github.com/LuaAndC/mxe/commit/7e218ad3b6240bb3cb2e4689c2c4232169cf0ad9 and https://github.com/LuaAndC/mxe/commit/508809b110bc9b9f7e537aad5afad64aadbb0230 .

It was caused by too precise grep pattern in configure.ac:

```shell
noncontribmakefiles=`find $srcdir -name '*.mak' -print | grep -v '^\./contrib/'`
```

Please `cherry-pick` or `am` my commits and rebase your branch against mxe:master.

darealshinji:
Changes were applied and branch was rebased.

starius:
List of installed files:

  * i686-w64-mingw32.shared/bin/libgs-9.dll
  * i686-w64-mingw32.shared/bin/test-ghostscript.exe
  * i686-w64-mingw32.shared/installed/ghostscript
  * i686-w64-mingw32.shared/lib/libgs.dll.a
  * i686-w64-mingw32.shared/lib/pkgconfig/ghostscript.pc
  * i686-w64-mingw32.static/bin/test-ghostscript.exe
  * i686-w64-mingw32.static/installed/ghostscript
  * i686-w64-mingw32.static/lib/libgs.a
  * i686-w64-mingw32.static/lib/pkgconfig/ghostscript.pc
  * x86_64-w64-mingw32.shared/bin/libgs-9.dll
  * x86_64-w64-mingw32.shared/bin/test-ghostscript.exe
  * x86_64-w64-mingw32.shared/installed/ghostscript
  * x86_64-w64-mingw32.shared/lib/libgs.dll.a
  * x86_64-w64-mingw32.shared/lib/pkgconfig/ghostscript.pc
  * x86_64-w64-mingw32.static/bin/test-ghostscript.exe
  * x86_64-w64-mingw32.static/installed/ghostscript
  * x86_64-w64-mingw32.static/lib/libgs.a
  * x86_64-w64-mingw32.static/lib/pkgconfig/ghostscript.pc

Looks good to me.

starius:
Thanks!

PR #1386, opened 2016-06-11
add tool install-deps

starius:


tools/install-deps
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+
+set -xue
+
+if [[ "$OSTYPE" =~ "linux" ]]; then
+    if which apt-get; then
+        # Debian or Ubuntu
```
darealshinji: Just a little heads-up: PCLinuxOS is an RPM based distro that uses apt as its installation frontend, but the package names are still similar to other RPM distros like Fedora.

starius: It is impossible to make a script working on all distributions. It should work on [supported](http://mxe.cc/#requirements) ones.

I tried to use `lsb_release` instead, but it is not installed by default on some OSes, e.g. Gentoo.

darealshinji: On PCLinuxOS apt calls rpm instead of dpkg, so you could check for dpkg too to at least exclude PCLinuxOS:
`if [ $(which apt-get) ] && [ $(which dpkg) ]; then`

starius: Changed to `if which apt-get && which dpkg; then` to exclude PCLinuxOS.

UniqueFool:
If someone really wanted to add support for other package managers, it might be possible to use a helper function to encapsulate package manager calls/details (e.g. name, arguments etc) in order to prepare support for different distros.

starius:
@UniqueFool, the approach with a helper function is too complex for a shell script, because one have to provide name of each package on each distro (map data structure => not shell). It can be implemented with Lua/Python/Ruby in lieu of Shell. But any of Lua/Python/Ruby has to be installed first, so we have the problem of chicken and egg. We can split the script to installer of Lua/Python/Ruby and Lua/Python/Ruby script installing other dependencies. I am not sure if it is worth it.

tools/install-deps
```diff
@@ -0,0 +1,81 @@
...
+            p7zip patch perl perl-XML-Parser pkg-config python \
+            ruby scons sed unzip wget xz
+        if ! [[ `uname -m` =~ i686 ]]; then
+            zypper install -R \
+                gcc-32bit glibc-devel-32bit libgcc46-32bit \
+                libgomp46-32bit libstdc++46-devel-32bit
+        fi
+    fi
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+    sudo port install \
```
tonytheodore: I'd remove `sudo`. The script should be invoked with the necessary privileges, the `sudo` is just a convenience for copy/paste from docs.

tonytheodore:
`lua` executable is half the size of `ext/config.guess` script so it wouldn't be unreasonable to add pre-built binaries to the git repo to bootstrap a configuration tool that abstracts package mangers... 

This looks ideal for the scope of provisioning automated build machines, the only real improvement would be to generate a `.json` file for `index.html` so that docs don't diverge. That's a low priority though as requirements change slowly and actually running them will expose those changes earlier.

I'd say it's good to merge (after removing `sudo`).

starius:
`sudo` removed. The pull request was rebased.

tonytheodore:
Thanks!

starius:
> lua executable is half the size of ext/config.guess script so it wouldn't be unreasonable to add pre-built binaries to the git repo to bootstrap a configuration tool that abstracts package mangers... 

I think, this solution is not ideal. Binaries are not portable across platforms. We would have to put lua binaries for all supported platforms. Moreover, binaries can not be read by humans. I wish we had lua interpreter written in shell...

darealshinji:
You could try to bootstrap Lua. Setting the default platform in Lua's Makefile to posix should probably work on most platforms. The source code without documentation has 672k.

starius:
@darealshinji, you need a compiler to build Lua from source. This script (or another script in Lua) installs the compiler. This is a chicken-and-egg problem.

PR #1387, opened 2016-06-11
create tmp-*/readonly instead of usr/readonly

### LINKED ISSUE #1221: How can I uninstall mxe

starius:
Removing MXE directory with `rm -rf` used to fail on file `usr/readonly/.gitkeep` because directory `usr/readonly` was readonly. Now readonly directory is created in `tmp-*` directory and `.gitkeep` is not created for it (because `tmp-*` is not under `usr/`). Problems with removing MXE directory are fixed even in case of interrupted build.

fix #1221


tonytheodore:
Nice!

PR #1390, opened 2016-06-12
patch-tool-mxe: rename existing .git directories

starius:
winpcap has directory wpcap/libpcap/.git in its source tree. Git considers wpcap/libpcap to be a submodule, which is unwanted.


patch.mk
```diff
@@ -22,6 +22,8 @@ define INIT_GIT
     # if PKG_SUBDIR is ".", the following will move gits/tmp/pkg
     mv '$(abspath $(GITS_DIR)/tmp/$(1)/$($(1)_SUBDIR))' '$(call GIT_DIR,$(1))'
     rm -rf '$(GITS_DIR)/tmp'
+    # rename existing .git directories if any
+    find '$(call GIT_DIR,$(1))' -name .git | rename 's/\.git$$/_git/'
```
tonytheodore: `rename` isn't available on darwin, `mv` is more portable.

starius: Fixed.

tonytheodore:
Thanks!

PR #1400, opened 2016-06-19
qtbase: fix build with -fontconfig and -system-freetype for MS Windows

tehnick:


mabrand:
Thanks. I pushed this as ac52a067e966f0ed79cd4c2e361180a64f6b9f3b.

It looks like https://codereview.qt-project.org/#/c/151625/ is what broke what you fixed. It assumed that fontconfig wouldn't be used with windows. Could you please propose your fix there? You can add the reviewers and author of the guilty patch to your reviewer list.

By the way, what are the benefits of enabling these options (fontconfig and freetype) on Windows? As far as I know, the standard Qt configuration (done with configure.exe) doesn't enable these. I don't know much about this subject.

tehnick:
> If these fonts are not found in a system, fontconfig will choose a sane alternative (with the most close look) from available ones instead of using the default font.

Probably you may check this difference by launching simple example program in Wine, which is lacks of some MS Windows fonts. But I have faced with this problem very long time ago and maybe I remember some things incorrectly...

tehnick:
@mabrand
I have tested build of qtbase for i686 and x86_64 with gcc6 and gcc4.9, but only for `static` targets and completely forgot about `shared` targets. Sorry. I'll fix it ASAP.

mabrand:
ok. If you have to revise the patch, just update your patch in Qt's Gerrit. I'll grab it from there.

PR #1401, opened 2016-06-20
qtbase: update patch

tehnick:


tehnick:
@mabrand
Now tested for all targets (static/shared and gcc6/gcc4.9).

mabrand:
Thanks. I pushed your patch.

PR #1373, opened 2016-06-03
protobuf: use official release tarball instead of tarball from git tag

tehnick:

PR #1273, opened 2016-03-23
add package googletest and update protobuf (fix #1269)

pavelvat:
Fix #1269. I have tested it for protobuf versions 2.6.1 and 3.0.0-beta-2. For version 3.0.0-beta-2 need to replace line:
```
cd '$(1)' && mv googletest-release-$(googletest_VERSION)/ gtest
```
on:
```
cd '$(1)' && mv googletest-release-$(googletest_VERSION)/ gmock
```

PR #1406, opened 2016-06-28
virtual package googletest is not required anymore

pavelvat:
Virtual package googletest was added in #1273. Since  #1373 package googletest is not required anymore.


starius:
Thanks!

PR #1409, opened 2016-06-28
install-deps: several improvements on Debian Jessie

starius:
Otherwise it will make interactive prompts.


tonytheodore:
Looks good!

PR #1415, opened 2016-07-01
fix races of downloading and verifying

starius:
Sometimes `make download -j 6` fails with errors in downloading
package openthreads (and some other). How to reproduce:

```
    $ sed 's/wget/wget --limit-rate=1M/' -i Makefile
    $ make download-only-plib download-only-openscenegraph \
        download-only-openthreads -j 2
    ...
    Download failed or wrong checksum of package openscenegraph!
```

In previous implementation download pipe (wget, gzip) used to write
its output directly to a target file in pkg/ directory.
If multiple packages (e.g., openscenegraph and openthreads) share
the same file in pkg/ directory, the download can result in failure:
while first package is verifying the checksum, another one overwrites
the file with partially downloaded content.

(Package plib in the example is needed to shift start of download of
openthreads. `--limit-rate=1M` is needed to increase duration of download.)

To fix this problem, wget command was changed to write to a unique
file, which is renamed to a target file afterwards. Renaming is
an atomic action.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -214,7 +214,8 @@ BACKUP_DOWNLOAD = \
     $(WGET) -O- $(PKG_CDN)/`$(call ESCAPE_PKG,$(1))`))
 
 DOWNLOAD_PKG_ARCHIVE = \
-        mkdir -p '$(PKG_DIR)' && ( \
+        mkdir -p '$(PKG_DIR)'/tmp && ( \
+        ( \
             $(WGET) -T 30 -t 3 -O- '$($(1)_URL)' \
             $(if $($(1)_URL_2), \
                 || (echo "MXE Warning! Downloading $(1) from second URL." >&2 && \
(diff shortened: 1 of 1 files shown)


tonytheodore:
Would something like the below be simpler:

```diff
diff --git a/Makefile b/Makefile
index cd45f8c..6f757a2 100644
--- a/Makefile
+++ b/Makefile
@@ -426,7 +426,8 @@ define PKG_RULE
 download-$(1): $(addprefix download-,$($(1)_DEPS)) download-only-$(1)

 .PHONY: download-only-$(1)
-download-only-$(1):
+download-only-$(1): download-only-$($(1)_FILE)
+download-only-$($(1)_FILE)::
        $(and $($(1)_URL),
        @[ -d '$(LOG_DIR)/$(TIMESTAMP)' ] || mkdir -p '$(LOG_DIR)/$(TIMESTAMP)'
        @if ! $(call CHECK_PKG_ARCHIVE,$(1)); then \
```

to avoid downloading the file altogether and help with the race condition when multiple targets are defined since they also use the same file. I haven't tested it much, but sanity check seems to do what I think it would:

```
$ make download-only-plib download-only-openscenegraph download-only-openthreads -j 2
[download]  plib
[download]  openscenegraph
make: Nothing to be done for `download-only-openthreads'.
```

starius:
@tonytheodore, your solution is better. Can you open a pull request for it?

Note, that second `:` fixes warnings "ignoring old commands for target ...". I think, it worth adding a comment with this information to `Makefile` after `download-only-$($(1)_FILE)::`.

PR #1417, opened 2016-07-04
Makefile: note use of target-specific variables

tonytheodore:
Following https://github.com/mxe/mxe/pull/1415#issuecomment-230095906, note use of lesser-known `make` features.


starius:
Thanks!

PR #1418, opened 2016-07-04
Plugin docs and add $(PKG)_PATCHES field

tonytheodore:
Resulting preview: https://github.com/tonytheodore/mxe/tree/plugin-docs/plugins

I've moved the `qt5-deps` to `examples` since I see it as the sort of plugin that people would maintain locally instead of being in the MXE tree - I still have to clarify those cases but I'm happy with this as a first pass.


plugins/README.md
```diff
@@ -1,27 +1,102 @@
...
+$ make all-luarocks MXE_PLUGIN_DIRS=plugins/luarocks
+```
+
+To always use gcc 6:
+
+```
+echo 'override MXE_PLUGIN_DIRS += plugins/gcc6' >> settings.mk
+```
+
+Note that multiple entires in `settings.mk` should not be escaped:
```
starius: > entires

typo? (entries)

pavelvat:
```
For example, if you want to build qbittorrent from the `apps` plugin with
gcc 6 run:

make qbittorrent MXE_PLUGIN_DIRS='plugins/gcc6 plugins/apps'
 ```

It is unclear. Can I use simultaneously GCC 6 for qbittorrent plugin and GCC 4.9.3 for src/*.mk? Currently I am using two separate copy of MXE: ```~/builds/mxe``` for GCC 4.9.3 and ```~/builds/mxe-gcc6/mxe``` for building of plugins through GCC 6.

**Comment by starius** [comment]
> It is unclear. Can I use simultaneously GCC 6 for qbittorrent plugin and GCC 4.9.3 for src/*.mk?

It is not supported. gcc6 and normal gcc share the same package name and target files in usr/. Moreover, qbittorrent depend on packages from src/*.mk, so they should be compiled with one compiler.

**Comment by tonytheodore** [comment]
Thanks, typos fixed and I've changed the examples to be more generic - gcc version mixing is unrelated. Also added more background and a fix to allow overriding patches.

https://github.com/tonytheodore/mxe/tree/plugin-docs/plugins

**Comment by starius** [comment]
Check for `_PATCHES` feature:

Apply [the patch](https://gist.github.com/9d053f7feee75dd7f4c5fc1778b58710) adding custom patch to lua in luarocks plugin.

Build and test:
```
$ make lua MXE_TARGETS=x86_64-unknown-linux-gnu MXE_PLUGIN_DIRS=plugins/luarocks/
$ ./usr/x86_64-unknown-linux-gnu/bin/lua
LUAAAAAAA!!!!
Lua 5.3.3  Copyright (C) 1994-2016 Lua.org, PUC-Rio
>
```

starius:
Looks good to me! Thanks!

pavelvat:
>gcc version mixing is unrelated

You can add small note that mixing GCC versions is not supported, so need used multiple copies of MXE in separate directories.

tonytheodore:
Is that a specific note about "mixing gcc versions" or a general note about what plugins do?

PR #1419, opened 2016-07-05
libwebsockets updated

newnon:


starius:
Could you make a patch from `libwebsockets-1-fixes.patch`, please? Now it is only a diff without commit message and authorship. You can use `./tools/patch-tool-mxe` to make a patch.

pavelvat:
It is deprecated:
```
      cd '$(1)' && cmake \
         -DCMAKE_TOOLCHAIN_FILE='$(CMAKE_TOOLCHAIN_FILE)'
```
See for example: https://github.com/mxe/mxe/blob/master/src/freeglut.mk#L20-26

newnon:
if it possible add please add information what you need, i've just fixed some files and make a patch with some GUI tool it is not from any commit.

starius:
@TimothyGu, you are the author of [original patch](https://github.com/mxe/mxe/blob/3caf14807e3c702dcae2ff25687426f3c22d9fda/src/libwebsockets-1-fixes.patch).
Can you review this change, please?

TimothyGu:
Unassigning myself; I will no longer have time to work on this unfortunately.

tonytheodore:
libwebsockets updated to 2.4.2 in https://github.com/mxe/mxe/commit/b3552f798ccdc03466f78b43fe9c1ab5df6eda18

PR #1421, opened 2016-07-05
chipmunk v7.0.1

newnon:


tonytheodore:
Similar to https://github.com/mxe/mxe/pull/1419#issuecomment-230430384, the patch should be created with `git format-patch` and ideally have some reference to upstream. I'll see if I can document it.

PR #1430, opened 2016-07-08
Update ffmpeg

tehnick:


tonytheodore:
Thanks! Building on all four gcc4 targets and downstream pkgs.

Issue #950, opened 2015-10-21
cross-compile Qt build tools: qmake.exe, rcc.exe...


starius:
See http://lists.nongnu.org/archive/html/mingw-cross-env-list/2015-10/msg00016.html

>  I need that when using the cross-compiled libs on windows platform for building other projects. For example, cmake does check for qmake.exe, rcc.exe etc..

aquiles2k:
Just want to second the need for this feature.

I mostly need this to work with a mxe compiled qt5 (including webkit) on a windows dev machine.

tonytheodore:
Qt's `configure` script doesn't seem to have an easy way to separate out the `qmake` bootstrapping. I imagine the other tools should be easy enough to cross-compile since they use `qmake` (which is known to work for the rest of Qt).

From [comment on host toolchain plugin](https://github.com/mxe/mxe/commit/ff8278ad07ad8fae10fa439969c297190804e4a8#commitcomment-18141312)
> Could this be adapted to also build qt host tools like qmake.exe rcc.exe etc. ?

You could try using the host toolchain on a windows machine to configure a minimal qt. The `configure` step will build `qmake.exe` and it may be straightforward from there to build just the tools.

aquiles2k:
Zapping through some code it seems qmake gets built inside the qtbase module which has a seperate configure and produces Makefiles with qmake targets. I'm not very familiar with the internals of cross-compiling unfortunately and if we already do this stuff, excuse my ignorance :) If not, it might be possible to adapt some of this to build a qmake.exe?

aquiles2k:
Ok, so far this kindof works:
* Building host-toolchain through plugin and copy it to win machine
* Download qt source and place it where u can reach it (i tried it with the whole thing - it will probably also work with just the qtbase module)
* create a batch file to set your path to the toolchain bin and call configure.bat
* qmake.exe gets build - find it in qtbase/bin

I have yet to check if it works with the qt5 version build through mxe - most probably the paths wont fit

tonytheodore:
See #1431

aquiles2k:
Cool, i'll test that next week and how it all comes together when using it as a fully functional qt build usable on the host machine.

tonytheodore:
Ha! I imagine "fully functional" will be the hard part (and this wasn't exactly easy) ;)

tonytheodore:
Qt test program is working - #1436

tonytheodore:
[...]

PR #1431, opened 2016-07-09
add qt5-host-tools example plugin

tonytheodore:
starting point for #950, builds `qmake.exe`, `rcc.exe`, etc


starius:
Tested on all 4 targets.

```
$ find usr/*/qt5-host-tools | gist
```
https://gist.github.com/dbcf9d99d17f34c2eab419e74a86d7e5

Thanks!

PR #1436, opened 2016-07-13
host-toolchain plugin: build qt5 test, cmake, and pkgconf, add `make` binary download

tonytheodore:


tonytheodore:
I'm all finished with this now and my curiosity is satisfied. Further examples require too much development on windows, though the modern tools (noted in `README.md`) are a worthwhile find.

I'll probably cross-compile `make` down the track but that will be a separate exercise.

tonytheodore:
Works on Win7 32/64 and Win10 64.

PR #1439, opened 2016-07-15
Update Wt from 3.3.5 to 3.3.6

starius:


tonytheodore:
Working on all four targets, thanks!

PR #1318, opened 2016-04-27
Adds yaml-cpp

NickNick:

PR #1440, opened 2016-07-15
add yaml-cpp (second attempt)

starius:
Following on from #1318


tonytheodore:
New format looks good. Building on all four targets, thanks!

Issue #1434, opened 2016-07-10
cegui is broken


starius:
Log: https://gist.github.com/c330c284f70af9c04df962197608ccf8
Machine: Debian Jessie
Commit: 2ee728cf99c0fe022e3fa9fddd0cfa41d79c8111

Error message:
```
CEGUIOpenGLRenderTarget.cpp:(.text+0x517): undefined reference to `gluUnProject@48'
CEGUIOpenGLRenderTarget.cpp:(.text+0x5a2): undefined reference to `gluUnProject@48'
CEGUIOpenGLRenderTarget.cpp:(.text+0x60d): undefined reference to `gluProject@48'
CEGUIOpenGLRenderTarget.cpp:(.text+0x678): undefined reference to `gluProject@48'
CEGUIOpenGLRenderTarget.cpp:(.text+0x6e3): undefined reference to `gluProject@48'
CEGUIOpenGLRenderTarget.cpp:(.text+0x9e6): undefined reference to `gluPerspective@32'
CEGUIOpenGLRenderTarget.cpp:(.text+0xa32): undefined reference to `gluLookAt@72'
```

brike:
i got this error too,how to fix it？

starius:
The following patch seems to fix cegui: https://github.com/mxe/mxe/commit/49f7a465cd20302aee3769e98d9a082f5a215330
The idea is to add `-lglu32` to linker options. MXE creates `glu.pc` which adds this option. I added `glu` to requirements of `freeglut`.

I am testing this patch. Then I'll open a PR.

tonytheodore:
Good pickup, I thought those were deprecated functions but [the previous `glut.pc` file included `-lglu32`](https://github.com/mxe/mxe/blob/bfc6be49a459fd9f9f7151530bc10b45c4258f20/src/freeglut-1-fixes.patch#L54).

dl5rcw:
@starius I just noticed the lglu32 issue and then found your fix here. I gave it a try - also to support you with testing - and copied your freeglut-1-fixes.patch fix to the last version a903d59466a6b7da214a38c68be9a146e5491841. Worked for all builds (shared+static, 32+64).

PR #1429, opened 2016-07-08
freeglut: update patch: fix pkgconfig file

tehnick:

PR #1441, opened 2016-07-16
freeglut: add glu to requirements in .pc file

starius:
Fix https://github.com/mxe/mxe/issues/1434
See https://github.com/mxe/mxe/pull/1429
See http://stackoverflow.com/a/17451000


tonytheodore:
Thanks!

PR #1330, opened 2016-05-04
[WIP] qt5: fix pkg-config files and enable debug variants

tonytheodore:
See #1316, build with:
```
make qt5 QT5_BUILD_TYPE=[-release | -debug-and-release | -debug]
```

PR #1442, opened 2016-07-16
qt5: fix pkg-config files and enable debug variants

tonytheodore:
From #1330, uses a more generic customisation variable `$(PKG)_CONFIGURE_OPTS` and defines the `BUILD_TYPE` variables in the main Makefile.

```
make qt5 qtbase_CONFIGURE_OPTS=[-release | -debug-and-release | -debug]
```


tonytheodore:
All three variants succeed on all four targets.

PR #1443, opened 2016-07-17
ocaml-camlimages: fix for giflib >=5.1

tonytheodore:
See: https://github.com/mxe/mxe/pull/1423#issuecomment-233153020

Fixes the build after `giflib` update, but I can't get the current patch to apply with `make import-patch-*` or by hand with `git am`.

Issue #1235, opened 2016-02-23
add issue template


starius:
The most important issues are bug reports. Many bug reports lack build logs. It's hard to solve a problem when important details are missing. GitHub [supports](https://github.com/blog/2111-issue-and-pull-request-templates) adding an issue template. I propose to add one to MXE repo with field "Log file".

See how it looks like:
https://github.com/LuaAndC/mxe/issues/new

tonytheodore:
+1

Would some links to common pastebin services be useful?

starius:
> Would some links to common pastebin services be useful?

Yes. I think, the template can be as follows:

```
Log: https://gist.github.com/
```

starius:
By mistake, I pushed untested commit https://github.com/mxe/mxe/commit/54706f6f3ff2386d75cbf5911ebffc79922b6580 directly to mxe:master.

The commit was reverted. Sorry for noise.

tonytheodore:
Closed by #1456.

PR #1445, opened 2016-07-17
build-pkg: do not omit gitignored files in usr

starius:
See https://github.com/mxe/mxe/pull/1443#issuecomment-233181951

### DIFF
--- a/tools/build-pkg.lua
+++ b/tools/build-pkg.lua
@@ -461,7 +461,7 @@ local function gitCheckout(new_branch, deps, item2index, pass_of_deps)
 end
 
 local function gitAdd()
-    os.execute(GIT .. 'add --all .')
+    os.execute(GIT .. 'add --all --force .')
 end
 
 -- return two lists of filepaths under ./usr/


tonytheodore:
Can you change commit message to "force inclusion of" rather than "do not omit"?

starius:
I have changed the commit message. Now it doesn't fit in [50 characters](http://chris.beams.io/posts/git-commit/#limit-50).

> So shoot for 50 characters, but consider 69 the hard limit.

tonytheodore:
I see, hadn't thought about 50 characters before but shorter can be clearer. How about "build-pkg: track all files"?

starius:
Chenged to `build-pkg: track all files including gitignored`.

tonytheodore:
Thanks!

tonytheodore:
What do you think of a [commit template](http://codeinthehole.com/writing/a-useful-template-for-commit-messages/)? Apparently, we can set [custom `prepare-commit-msg
` hooks](https://www.kernel.org/pub/software/scm/git/docs/githooks.html) for a repo-specific template.

starius:
It is a good idea (as well as [issue template](https://github.com/mxe/mxe/issues/1235)).
Can we add the hook project-wide?

PR #1447, opened 2016-07-19
[WIP] Upgrade CEGUI to 0.8.7

Quintus:
Hi everyone,

I am trying to upgrade the CEGUI packaged with MXE from the outdated 0.7.9 to the current 0.8.7. That wasn't as easy as expected, namely because CEGUI is now built with CMake rather than autotools. There are some things that do not work yet and I need help.

For one, it does not build the libxml2-based module due to a weird duplication of the `-D` flag that makes CMake invoke g++ with `-D-DLIBXML_STATIC`, which (of course) makes it complain about an error in the commandline. Since CEGUI allows other XML parser modules, I'll ignore that issue for a moment and come back to it later.

The more important problem is that the OpenGL renderer module does not compile. It fails to find symbols in GLEW, which it actually should. The build errors look like this:

~~~~~
/home/quintus/repos/privat/projekte/misc/mxe/usr/lib/gcc/i686-w64-mingw32.static/4.9.3/../../../../i686-w64-mingw32.static/lib/../lib/libglew32.a(glew.o):glew.c:(.text+0x3c7): undefined reference to `_imp__wglGetProcAddress@4'
/home/quintus/repos/privat/projekte/misc/mxe/usr/lib/gcc/i686-w64-mingw32.static/4.9.3/../../../../i686-w64-mingw32.static/lib/../lib/libglew32.a(glew.o):glew.c:(.text+0xe47): undefined reference to `_imp__wglGetProcAddress@4'
/home/quintus/repos/privat/projekte/misc/mxe/usr/lib/gcc/i686-w64-mingw32.static/4.9.3/../../../../i686-w64-mingw32.static/lib/../lib/libglew32.a(glew.o):glew.c:(.text+0x18f7): undefined reference to `_imp__wglGetProcAddress@4'
[... and so on...]
~~~~~

These are the infamous `_imp` errors we all know and love when linking static libraries for Windows. GLEW allows us to specify `GLEW_STATIC` to get rid of the `__declspec` causing this in the GLEW headers. I know this, and the MXE `pkg-config` (which CEGUI does not consult, hence my Makefile invokes it manually) provides the proper `-DGLEW_STATIC` flag. But the problem does not go away. Even with `-DGLEW_STATIC` in the commandline, the `_imp` errors still occur. I'm sure there must be an easy way around this, but when debugging this I got confused by the number of GLEW libraries MXE provides, namely:

* libglew32.a
* libGLEW.a
* libglew32s.a
* libGLEWmx.a

The unpatched CEGUI build system finds and links `libGLEW.a`, causing similar `_imp` errors. I then patched it to find `libglew32.a`, (32 reminds me of Win32, so that was a good guess in my opinion), but I still get the `_imp` errors you see above. The other libraries also give `_imp` errors. So which one is the correct one to use now? And why do I get `_imp` errors with GLEW albeit `-DGLEW_STATIC` is provided?

Here's the actual command executed by CMake:

~~~~~
home/quintus/repos/privat/projekte/misc/mxe/usr/bin/i686-w64-mingw32.static-g++    -I/home/quintus/repos/privat/projekte/misc/mxe/usr/i686-w64-mingw32.static/include/freetype2 -DGLEW_STATIC -I/home/quintus/repos/privat/projekte/misc/mxe/usr/i686-w64-mingw32.static/include -DFREEIMAGE_LIB -std=c++03  -fsigned-char -Wall -Wextra -O3 -DNDEBUG   -fsigned-char -Wall -Wextra -shared -o ../../../../bin/libCEGUIOpenGLRenderer-0.dll -Wl,--out-implib,../../../../lib/libCEGUIOpenGLRenderer-0.dll.a -Wl,--major-image-version,2,--minor-image-version,4 -Wl,--whole-archive CMakeFiles/CEGUIOpenGLRenderer-0.dir/objects.a -Wl,--no-whole-archive @CMakeFiles/CEGUIOpenGLRenderer-0.dir/linklibs.rsp
~~~~~

Any hints are appreciated.

Greetings
Marvin


Quintus:
I found the error. It’s the order of the linking flags that matters. `-lGLEW` must come **before** `-lopengl32` and **before** `-lglu32` (tested by modifying CMake's `linklibs.rsp` after generation). The OpenGL renderer links if the order is like this: `-lGLEW -lglu32 -lopengl32`. CMake uses `-lglu32 -lopengl32 -lGLEW`, which exhibits the above error. How do I tell CMake to order the library link flags in exactly that order?

Greetings
Marvin

tonytheodore:
You could try changing the [order of the `cegui_add_dependency ` lines (near the end)](https://bitbucket.org/cegui/cegui/src/e201a330be77bbd7dbb13ee36803cef8a0ffc1d7/cegui/src/RendererModules/OpenGL/CMakeLists.txt?at=v0-8&fileviewer=file-view-default#CMakeLists.txt-55).

Quintus:
I filed a bug report upstream: https://bitbucket.org/cegui/cegui/issues/1131/opengl-renderers-fail-to-crosscompile-due

Quintus:
Thank you for the quick response. I'll try that, but let's wait how the CEGUI devs react.

tonytheodore:
Strange that it's not using `pkg-config` for `freetype` via the [`FindFreetype.cmake` module](https://github.com/mxe/mxe/blob/master/src/cmake/modules/FindFreetype.cmake)

Quintus:
The tinyxml-based XML parser does not compile because `TIXML_USE_ST` is not defined. I opened #1463 for this.

Quintus:
Okay. I have opened a whole set of tickets on the CEGUI tracker now:

* https://bitbucket.org/cegui/cegui/issues/1131/opengl-renderers-fail-to-crosscompile-due
* https://bitbucket.org/cegui/cegui/issues/1132/libxml2-based-xml-parser-fails-to
* https://bitbucket.org/cegui/cegui/issues/1133/tinyxml-based-xml-parser-fails-to
* https://bitbucket.org/cegui/cegui/issues/1134/ceguibase-library-fails-to-crosscompile

On the original problem. @tonytheodore I can confirm your suggestion regarding exchanging the order of the `cegui_add_dependency` lines works. Even better, a CEGUI developer has already [pushed a fix for that](https://bitbucket.org/cegui/cegui/commits/bf699651740a1206c9ce214b91426db1e709ddc4).

Greetings
Marvin

tonytheodore:
@Quintus, #1465 should fix the `freetype` issue.

src/cegui-build-pkgconfig-files.patch
```diff
@@ -0,0 +1,13 @@
+# CEGUI does not build the .pc files when targetting Windows.
+# See upstream ticket: https://bitbucket.org/cegui/cegui/issues/1135
+--- old/CMakeLists.txt	2016-04-28 21:12:57.000000000 +0200
++++ new/CMakeLists.txt	2016-07-29 13:07:15.095432616 +0200
+@@ -599,7 +599,7 @@
+ configure_file( samples/common/include/CEGUISamplesConfig.h.in samples/common/include/CEGUISamplesConfig.h )
+ configure_file( doc/doxygen/doxyfile.in doc/doxygen/doxyfile )
+ 
+-if (UNIX AND NOT APPLE)
++if (NOT(APPLE))
```
tonytheodore: you can use `if (UNIX OR MINGW)` here.

tonytheodore:
Getting closer, it appears to be building both shared and static libs and some changes are needed for the test program to build:
```diff
diff --git a/src/cegui-test.cpp b/src/cegui-test.cpp
index 36b5bf3..1249f81 100644
--- a/src/cegui-test.cpp
+++ b/src/cegui-test.cpp
@@ -8,9 +8,9 @@
  * and makes CEGUI draw an "in-game" window into it.
  ************************************************/

-#include <GL/freeglut.h>
 #include <CEGUI/CEGUI.h>
 #include <CEGUI/RendererModules/OpenGL/GLRenderer.h>
+#include <GL/freeglut.h>

 // We’re lazy
 using namespace CEGUI;
@@ -65,11 +65,11 @@ int main(int argc, char* argv[])

   // Create the hypothetical CEGUI root window
   Window* p_root_window = WindowManager::getSingleton().createWindow("DefaultWindow", "root");
-  gui_context.setRootWindow(p_rootwindow);
+  gui_context.setRootWindow(p_root_window);

   // Create an actual framed window we can look onto
   FrameWindow* p_frame_window = static_cast<FrameWindow*>(WindowManager::getSingleton().createWindow("TaharezLook/FrameWindow", "testWindow"));
-  p_root_window->addChildWindow(p_frame_window);
+  p_root_window->addChild(p_frame_window);
   p_frame_window->setPosition(UVector2(UDim(0.25f, 0), UDim(0.25f, 0)));
   p_frame_window->setSize(USize(UDim(0.5f, 0), UDim(0.5f, 0)));
   p_frame_window->setText("Hello World!");
```

tonytheodore:
See https://github.com/Quintus/mxe/pull/1 for some fixes and style changes.

src/cegui-build-pkgconfig-files.patch
Quintus: Thanks for the hint. I googled on how to check for MinGW and got several sites suggesting to parse the value of the variable describing the compiler. This is much cleaner indeed.

Quintus:
I included your changes and your suggestion. It builds now correctly for me. However, the CEGUI pkg-config files do not mention the flag `-DCEGUI_STATIC`, which is required to use a statically compiled CEGUI (but not a dynamically linked one, obviously). I mentioned this to the CEGUI devs at [this new ticket](https://bitbucket.org/cegui/cegui/issues/1135/pkg-config-files-not-installed-when-using). Given that your changes appear to build CEGUI both in the shared and statically linked variant, I'm not sure how I should add this flag.

Greetings
Marvin

Quintus:
I tried to build the DevIL-based imagecodec as MXE also ships DevIL, but CMake fails to detect IL. It detects ILU, though. When building the renderer, the build fails at the linking stage, and VERBOSE=1 reveals that it doesn't even try to pass `-lIL` or `-lILU`.

Before I annoy the CEGUI people with yet another ticket (I think I have opened 6 tickets by now on this topic), can you check if there's another problem with MXE not setting some CMake variables as it should?

Greetings
Marvin

Quintus:
Regarding the libxml2-based parser build failure, the CEGUI developer suspects a bug in CMake itself:  https://bitbucket.org/cegui/cegui/issues/1132/libxml2-based-xml-parser-fails-to#comment-29547064

Quintus:
Okay, CEGUI devs resolved the pkg-config thing a little different. Upstream commit: https://bitbucket.org/cegui/cegui/commits/67125f71a8e953d93bbbc263e99bfe04655524ba

I won't copy this, our patch is just fine as well. When the commit gets into a stable release, it will do just okay also and the patch can be removed.

The LibXML2 build problems are so far being addressed in two commits:

* https://bitbucket.org/cegui/cegui/commits/2eed615dc8ea8768fb86645788504876bb68eb8c
* https://bitbucket.org/cegui/cegui/commits/a6263f0f4fd53bf92b5cdde210dd3ea7e608de0a

Should I incorporate them as patches?

It still gives linking errors, see the recent discussion [on the CEGUI tracker](https://bitbucket.org/cegui/cegui/issues/1132/libxml2-based-xml-parser-fails-to).

Greetings
Marvin

Quintus:
Thread opened by the CEGUI dev on the CMake mailinglist: http://public.kitware.com/pipermail/cmake/2016-July/063995.html

tonytheodore:
> Given that your changes appear to build CEGUI both in the shared and statically linked variant, I'm not sure how I should add this flag.

I managed to separate the shared and static builds (by hand, not automated yet) and will try to look at the .pc files tomorrow. The `-DCEGUI_STATIC` goes in the `Cflags.private` section and the files also need `Requires` sections populated.

Quintus:
The CEGUI dev promised to make an option available for compiling CEGUI static-only, see here: https://bitbucket.org/cegui/cegui/issues/1132/libxml2-based-xml-parser-fails-to#comment-29553011

Maybe we should just merge this and update to the next CEGUI release once it is out. Looks as if it resolve a number of problems.

tonytheodore:
> I won't copy this, our patch is just fine as well. When the commit gets into a stable release, it will do just okay also and the patch can be removed.

I suspect the upstream patch will fail when cross-compiling on OSX to mingw.

Re. Devil and LibXML2, let's get a complete build of the currently enabled packages working first.

tonytheodore:
> make an option available for compiling CEGUI static-only

Wish I'd seen that a few hours ago...

> Maybe we should just merge this PR and update to the next CEGUI release once it is out.

Probably better to switch to tracking the development branch while it's at the front of mind.

Quintus:
> Re. Devil and LibXML2, let's get a complete build of the currently enabled packages working first.

The current state successfully builds for me.

> Probably better to switch to tracking the development branch while it's at the front of mind.

Does this mean waiting for the 0.8.8 release of CEGUI or is  there some way to target CEGUI's version control system (Mercurial) as the source download?

Greetings
Marvin

tonytheodore:
> The current state successfully builds for me.

It was here too, but without [separation of the shared/static builds](https://github.com/Quintus/mxe/commit/b07b534fd5516fac860868b01406da62834d26e4), it was actually linking against the default shared libs (`gcc` can do that)

> is there some way to target CEGUI's version control system (Mercurial) as the source download?

A mix of [local testing against a repo](https://github.com/mxe/mxe/pull/1470) and [tracking dev branches](https://github.com/Quintus/mxe/commit/c2a1d8000ab984524897bf9c79080e38a8556ab1). We do the latter with github often (even maintain various [git-based mirrors](https://github.com/mirror/mirror)) - bitbucket also seems to offer deterministic tarballs for commits.

> Maybe we should just merge this PR

See https://github.com/Quintus/mxe/pull/2, if that works for you, I'm happy to merge this.

Quintus:
> if that works for you, I'm happy to merge this.

Builds just fine. Thank you for your enormous effort in this issue :+1: 

Greetings
Marvin

tonytheodore:
Great! Thanks for testing and the detailed reports.

Back to your original question, `glew` provides two libs, `libGLEW.a` and `libGLEWmx.a` - `libglew32.a` and `libglew32s.a` are just copies of `libGLEW.a` for compatibility with config systems that look for different names.

Issue #1446, opened 2016-07-19
libm, libiconv, libpaper, libopenjp2 and libwinspool do not install pc-file


starius:
From https://github.com/mxe/mxe/pull/1382#issuecomment-233607349

Progress:
- [ ] libm
- [ ] libiconv
- [ ] libpaper
- [ ] libopenjp2
- [ ] libwinspool
- [ ] add build-pkg check of whether pc-file is installed

darealshinji:
All the mingw libraries come without pc files. But you can generate them.
Here's an example Makefile:
``` make
i686_w64_mingw32_static_libs = \
	aclui activeds adsldpc advapi32 apcups api-ms-win-core-winrt-l1-1-0 \
	api-ms-win-core-winrt-string-l1-1-0 authz avicap32 avifil32 avrt bcrypt bootvid \
	browcli bthprops cabinet cap cfgmgr32 classpnp clfsw32 clusapi cmutil comctl32 \
	comdlg32 connect credui crtdll crypt32 cryptnet cryptsp cryptxml cscapi ctl3d32 \
	d2d1 d3d11 d3d8 d3d9 d3dcompiler_33 d3dcompiler_34 d3dcompiler_35 \
	d3dcompiler_36 d3dcompiler_37 d3dcompiler_38 d3dcompiler_39 d3dcompiler_40 \
	d3dcompiler_41 d3dcompiler_42 d3dcompiler_43 d3dcompiler_46 d3dcompiler_47 \
	d3dcompiler d3dcsx_46 d3dcsxd_43 d3dcsxd d3dim d3drm d3dx10_33 d3dx10_34 \
	d3dx10_35 d3dx10_36 d3dx10_37 d3dx10_38 d3dx10_39 d3dx10_40 d3dx10_41 d3dx10_42 \
	d3dx10_43 d3dx10 d3dx11_42 d3dx11_43 d3dx11 d3dx8d d3dx9_24 d3dx9_25 d3dx9_26 \
	d3dx9_27 d3dx9_28 d3dx9_29 d3dx9_30 d3dx9_31 d3dx9_32 d3dx9_33 d3dx9_34 \
	d3dx9_35 d3dx9_36 d3dx9_37 d3dx9_38 d3dx9_39 d3dx9_40 d3dx9_41 d3dx9_42 \
	d3dx9_43 d3dx9 d3dx9d d3dxof davclnt davhlpr dbgeng dbghelp ddraw delayimp \
[...]

PR #1450, opened 2016-07-23
pkg-config files

darealshinji:
Install pc files for libiconv, libpaper and openjpeg. Update ghostscript's pc file entries.
Should fix some points from https://github.com/mxe/mxe/issues/1446

### DIFF
--- a/src/ghostscript.mk
+++ b/src/ghostscript.mk
@@ -64,7 +64,7 @@ define $(PKG)_BUILD
      echo 'Cflags: -I"$(PREFIX)/$(TARGET)/include/ghostscript"'; \
      echo 'Cflags.private: -DGS_STATIC_LIB'; \
      echo 'Libs: -L"$(PREFIX)/$(TARGET)/lib" -lgs'; \
-     echo 'Requires: libiconv libidn libpaper libtiff-4 libpng jpeg openjp2 lcms2 zlib'; \
+     echo 'Requires: libiconv libidn libpaper libtiff-4 libpng jpeg openjpeg lcms2 zlib'; \
      echo 'Libs.private: -lwinspool';) \
      > '$(PREFIX)/$(TARGET)/lib/pkgconfig/ghostscript.pc'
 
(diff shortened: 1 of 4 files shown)


starius:
Why does package `openjpeg` create both `openjp2.pc` and `openjpeg.pc`?

starius:
For `libiconv` you create .pc file for native target (`$(PKG)_BUILD_$(BUILD)`) instead of cross-target (`$(PKG)_BUILD`). `ghostscript` fails to build because it can't find `libiconv.pc`.

Please create .pc file for cross-target of `libiconv`.

darealshinji:
> Why does package openjpeg create both openjp2.pc and openjpeg.pc?

I wasn't sure if I should name the pc file `openjp2.pc` or `openjpeg.pc`.

starius:
> I wasn't sure if I should name the pc file openjp2.pc or openjpeg.pc.

I [grepped github](https://github.com/search?l=cmake&q=openjpeg.pc&ref=searchresults&type=Code&utf8=%E2%9C%93) and discovered many variants, including `libopenjpeg.pc`. However it is confusing to have a package under different names in one installation. (At the moment, there is only one symlink in the library of packages: `libpng.pc -> libpng16.pc`.) IMHO, we should keep only one of them and `openjpeg.pc` seems the best candidate, because it matches with the package name in MXE.

starius:
`libpaper` is not built for shared targets, so `libpaper.pc` is not created for shared libraries, so `ghostscript` fails.

By the way, why `ghostscript` didn't fail to build on shared libraries without libpaper? Is it an optional dependency?

darealshinji:
Yes, I think it's optional. I think most of the non-optional dependencies are those that are included within the source package.
Should I remove the dependency on libpaper or should I add a conditional check like this?
``` sh
echo 'Requires: libiconv libidn $(if $(BUILD_STATIC),,libpaper) libtiff-4 libpng jpeg openjpeg lcms2 zlib'; \
```

starius:
> echo 'Requires: libiconv libidn $(if $(BUILD_STATIC),,libpaper) libtiff-4 libpng jpeg openjpeg lcms2 zlib'; \

Looks good to me.

[In general](https://github.com/mxe/mxe/issues/1111), optional dependencies should be either:

  * explicitly disabled in cmake/configure options,
  * or treated as normal dependencies

to achieve predictable build artefacts not dependant on build order.

Rinat84:
https://distfiles.dereferenced.org/pkgconf

tonytheodore:
@Rinat84, that appears to be the source tarballs for `pkgconf` - I'm not sure what you intended to link to?

tonytheodore:
These really should come from the upstream projects, we can add *.pc files where it's convenient for us, but not across the board.

Issue #910, opened 2015-10-08
lua installs lua53.dll instead of liblua53.dll


starius:
Is there any reason for it to be `lua53.dll` (not `liblua53.dll`)? Most of libraries in bin/ have `lib` prefix.

tonytheodore:
What's the general convention in the lua community? It's better to match that so feel free to change it rather than modify any downstream build rules.

starius:
[LuaDist](http://luadist.org/) distributes `bin/liblua.dll`.
Cygwin [installs ](https://cygwin.com/cgi-bin2/package-cat.cgi?file=x86_64%2Flua%2Flua-5.1.5-1&grep=dll) `cyglua-5.1.dll` and `liblua5.1.dll.a`.

TimothyGu:
It's probably better to change it to liblua5.3.dll, just to be compatible with Cygwin and Linux (on Ubuntu it's `liblua5.3.so`)

starius:
+1 for `liblua5.3.dll`

PR #1458, opened 2016-07-27
Windows Lua DLLs should not start with "lib"

fiendish:
The canonical Lua DLL file name in Windows is either Lua5x.dll or Lua5.x.dll.

### DIFF
--- a/plugins/luarocks/luarocks-1-fixes.patch
+++ b/plugins/luarocks/luarocks-1-fixes.patch
@@ -81,7 +81,7 @@ index 1111111..2222222 100644
 +   }
 +   defaults.runtime_external_deps_patterns = {
 +      bin = { "?.exe", "?.bat" },
-+      lib = { "?.dll", "lib?.dll" },
++      lib = { "?.dll" },
 +      include = { "?.h" }
 +   }
 +end
--- a/src/lua.mk
+++ b/src/lua.mk
@@ -6,6 +6,7 @@ $(PKG)_IGNORE   :=
 $(PKG)_VERSION  := 5.3.3
 # Shared version and luarocks subdir
 $(PKG)_SHORTVER := $(call SHORT_PKG_VERSION,$(PKG))
+$(PKG)_DLLVER   := $(subst .,,$($(PKG)_SHORTVER))
 $(PKG)_CHECKSUM := 5113c06884f7de453ce57702abaac1d618307f33f6789fa870e87a59d772aca2
 $(PKG)_SUBDIR   := lua-$($(PKG)_VERSION)
 $(PKG)_FILE     := lua-$($(PKG)_VERSION).tar.gz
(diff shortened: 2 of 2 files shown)


fiendish:
This travis failure is entirely unrelated. :\

tonytheodore:
See #910, the convention appears to be `libluaX.Y.dll`?

fiendish:
I think that thread #910 is misguided and harmful. MXE isn't targeting cygwin or Linux and so producing something that conforms to their naming is a mistake. MXE runs in Linux but it is by definition targeting the creation of Windows native binaries.

The official Lua and LuaJIT Windows DLLs are both named lua5x.dll.

Many "batteries included" packages make arbitrary and ill-conceived name changes, possibly for project sandboxing, possibly because they think they know versioning better, and possibly just because it's what they're used to in Linux, that are ultimately quite negative because it breaks drop-in compatibility with other software that is already built and not easy to rebuild/redistribute without forcing the (not completely straightforward) creation of yet-another-shim-dll-that-does-nothing-but-forward-calls. It's a bad practice that causes tons of additional rebuilding and maintenance effort.

(I've amended the PR because this means the SHORT_PKG_VERSION call is also invalid)

fiendish:
no further thoughts?

tonytheodore:
I'm mostly [ambivalent on this](http://martinfowler.com/bliki/TwoHardThings.html), perhaps you should link to the [official binaries](https://sourceforge.net/projects/luabinaries/files/5.3.3/Windows%20Libraries/Dynamic/) drawing particular attention to the [naming convention used to describe the runtimes they're built with](http://luabinaries.sourceforge.net/packaging.html).

Are you seeing build-time or runtime issues? Would a missing *.dll be worse than a *.dll with missing functions? My thoughts lead towards an `mxe-` prefix on all shared libs.

starius:
`lua` and `luarocks` were tested on `i686-w64-mingw32.shared`. Everything is Ok with the build.

I opened https://github.com/mxe/mxe/issues/910 just because `lua` was the only package in MXE installing DLL file without `lib` prefix. I did not know about the official naming conventions of Lua DLLs on Windows and wanted to level `lua` with other packages.

This pull request looks good to me.

fiendish:
> Are you seeing build-time or runtime issues? Would a missing *.dll be worse than a *.dll with missing functions? My thoughts lead towards an mxe- prefix on all shared libs.

This is a runtime thing. The entire point of linking dynamically at runtime is to be able to swap in any other version of the library. This requires your swapped in version to have the same name and, critically, the same code entry points.

fiendish:
(oops. I accidentally hit close. Had to reopen, sorry.)

fiendish:
> Are you seeing build-time or runtime issues? Would a missing *.dll be worse than a *.dll with missing functions? My thoughts lead towards an mxe- prefix on all shared libs.

Here's basically my scenario to explain why renaming DLLs introduces a little bit of unnecessary chaos (at least for me)...

Application A dynamically loads Lua51.dll. Lua code executed within the Lua context of Application A loads llthreads2 and luasocket and does stuff with them. llthreads2 and luasocket are both included in MXE, so I say to myself "This is great. My dreams come true. I can make sure everything is up to date and get some nice no-nonsense no-need-for-msvcr100 dlls all at once!"
 
llthreads2 and luasocket are also both compiled linking against the Lua dll.

Now we see there is a minor problem. llthreads2 and luasocket both look for mxe_whatever_lib_lua_dll_name_foo.dll, and Application A looks for the official name Lua51.dll. We can't have both, because multiple nested Lua runtimes leads to errors.

The possible solutions I can think of then are 
1) Rebuild Application A so that it looks for mxe_whatever_lib_lua_dll_name_foo.dll. This can't be done easily without Windows and an MSVC++ with MFC support and thus 30GB of free drive space and a lot of time (Visual Studio Community is gigantic).
2) Make a dummy stub DLL (a process which has no fewer than 4 different suggested approaches, each of which is somewhat arcane in what exactly differentiates them) called mxe_whatever_lib_lua_dll_name_foo.dll that does nothing but forward all calls over to Lua51.dll
3) Build llthreads2 and luasocket (and others) to look for Lua51.dll in the first place to avoid this mess.

tonytheodore:
Thanks for the explanation, I had assumed that mixing msvcr* versions would causes new problems. If shared libs do, in practice, isolate transitive dependencies, then it makes a lot more sense why people choose them.

PR #1454, opened 2016-07-24
patch.mk: create target subdir if it is multilevel

starius:
Example: xvidcore. Its subdir is xvidcore/build/generic.
This commit adds creation of xvidcore/build prior to moving
a directory to xvidcore/build/generic.

PR #1460, opened 2016-07-27
xvidcore: _SUBDIR is top-level directory

starius:
See https://github.com/mxe/mxe/pull/1454#issuecomment-235508913

**Work-in-progress!** It builds here, but I have not compared lists of files.


tonytheodore:
Looks good to me.

starius:
Tested. It produces the same set of files.

PR #1465, opened 2016-07-29
FindFreetype.cmake: add INCLUDE_DIR

tonytheodore:
From https://github.com/mxe/mxe/pull/1447#issuecomment-235851175, `cegui` expects `INCLUDE_DIR` instead of `INCLUDE_DIRS`.

### DIFF
--- a/src/cmake/modules/FindFreetype.cmake
+++ b/src/cmake/modules/FindFreetype.cmake
@@ -9,4 +9,6 @@ endif()
 
 pkg_check_modules(FREETYPE freetype2)
 
-set(FREETYPE_LIBRARY ${FREETYPE_LIBRARIES}) #For compatibility
+#For compatibility
+set(FREETYPE_LIBRARY ${FREETYPE_LIBRARIES})
+set(FREETYPE_INCLUDE_DIR ${FREETYPE_INCLUDE_DIRS})


Quintus:
I have to say if there's one thing I dislike about CMake, it's the completely unclear naming convention of `INCLUDE_DIR` vs. `INCLUDE_DIRS` and `LIBRARY` vs. `LIBRARIES`. Each module does it differently, and that's confusing. The CMake devs should make a definite decision here really.

Thanks for catching this. I wouldn't have found this to be the cause ever.

Vale,
Quintus

Quintus:
Looking at the diff, this wasn't the problem. Sorry for the noise, but it's still an annoying thing of CMake.

tonytheodore:
@Quintus, to get this to work with https://github.com/Quintus/mxe/pull/1, you'll need to `touch src/mxe-conf.mk`. I've raised a new issue for it #1467.

Quintus:
> Looking at the diff, this wasn't the problem. Sorry for the noise, but it's still an annoying thing of CMake.

Or was it? Damned, I am getting seriously confused. Ignore whatever I have said on this topic, it must be FUD as I have no knowledge of MXE internals.

Sorry again.

>  to get this to work with Quintus#1, you'll need to touch src/mxe-conf.mk. I've raised a new issue for it #1467.

Already applied to my work. Thanks!

Greetings
Marvin

tonytheodore:
So I tried updating `cegui` a few weeks ago for [unrelated reasons](https://github.com/mxe/mxe/pull/1429#issuecomment-231351274) and was completely baffled by this. Normally, a switch to CMake "magically" solves various issues. If it doesn't, finding a plural is much easier than some [alternatives](http://freecode.com/articles/stop-the-autoconf-insanity-why-we-need-a-new-build-system).

Keep going with detailed reports of your investigations.

Issue #1462, opened 2016-07-28
qt3d fails to build


dl5rcw:
On Debian I did a `make qt5` and the built failed for qt3d. Please find the full log here: https://gist.github.com/dl5rcw/7f9e679ebc3f74c7e5b5442643bdf9cd
It is reproducible on the debian build-server.

For your first insight, here is the potential log sniplet
make[5]: Entering directory '/share/mxe/tmp-qt3d-i686-w64-mingw32.static/qt3d-opensource-src-5.7.0/src/quick3d/imports/extras'
make[5]: *** No rule to make target '/share/mxe/tmp-qt3d-i686-w64-mingw32.static/qt3d-opensource-src-5.7.0/lib/libQt53DQuick.a', needed by '../../../../qml/Qt3D/Extras/libquick3dextrasplugin.a'.  Schluss.
make[5]: Leaving directory '/share/mxe/tmp-qt3d-i686-w64-mingw32.static/qt3d-opensource-src-5.7.0/src/quick3d/imports/extras'
Makefile.importsextras:34: recipe for target 'release' failed
make[4]: *** [release] Error 2
make[4]: Leaving directory '/share/mxe/tmp-qt3d-i686-w64-mingw32.static/qt3d-opensource-src-5.7.0/src/quick3d/imports/extras'
Makefile:414: recipe for target 'sub-quick3d-imports-extras-make_first' failed
make[3]: *** [sub-quick3d-imports-extras-make_first] Error 2
make[3]: *** Warte auf noch nicht beendete Prozesse...

dl5rcw:
I found no 'real' issue, so I went for a race condition. I compile with JOBS:= 4 in the settings.mk file.
When I did JOBS:=1, qt3d compiled! The past days, I had KUbuntu do JOBS:= 8 and qt3d was OK.
I suggest to modify qt3d.mk
```
define $(PKG)_BUILD
    cd '$(1)' && '$(PREFIX)/$(TARGET)/qt5/bin/qmake'
    #$(MAKE) -C '$(1)' -j  '$(JOBS)'
    $(MAKE) -C '$(1)' -j  1
    $(MAKE) -C '$(1)' -j 1 install
endef
```
to prevent such ugly behaviour in future. I guess this issue is something to report to qt3d itself.

tonytheodore:
Seems fine with either 6 or 8 jobs. A retry is usually better:
```
[...]

PR #1468, opened 2016-07-30
fix for #1462 where qt3d failed to build on JOBS := 4

dl5rcw:
this is fixing the issue #1462 as proposed by @tonytheodore. Testing was good now.


tonytheodore:
Thanks!

Issue #1452, opened 2016-07-23
use SOURCE_DIR, BUILD_DIR and TEST_FILE in all packages


starius:

starius:
Starting point: [script](https://gist.github.com/92b9cbf5175bcac1d07725802fde5f4a) to get rid of `$(2)` in all `*.mk` files.

PR #1469, opened 2016-07-30
use $(TEST_FILE) instead of $(2).c and other

starius:
The following script was applied:

```bash
set -xue

find src/ plugins/ -name '*.mk' -type f > mks

# special cases for test file:
# ls -d src/*-test* | sed 's@-test.*@@' | sort | uniq --repeated
sed 's@$(2).c@$(PWD)/src/$(PKG)-test.c@g' -i \
	src/{boost,freeimage,hdf5,lapack,sdl,sdl_image,sdl_sound}.mk
sed 's@$(2).f@$(PWD)/src/$(PKG)-test.f@g' -i src/lapack.mk
sed 's@$(PWD)/$(2).pro@$(PWD)/src/$(PKG)-test.pro@g' -i \
	src/qt.mk plugins/examples/custom-qt-min/overrides.mk
sed 's@$(2)-CMakeLists.txt@$(PWD)/src/$(PKG)-test-CMakeLists.txt@g' \
	-i src/{sdl_image,sdl_sound}.mk

# check
for base in $(ls -d src/*-test* | sed 's@-test.*@@' | sort | uniq --repeated); do
	! grep -q '$(2)' "${base}.mk"
done

# other $(2)
sed 's@$(2).cpp@$(TEST_FILE)@g' -i $(cat mks)
sed 's@$(2).cxx@$(TEST_FILE)@g' -i $(cat mks)
sed 's@$(2).c@$(TEST_FILE)@g' -i $(cat mks)
sed 's@$(2).ml@$(TEST_FILE)@g' -i $(cat mks)
sed 's@$(2).f@$(TEST_FILE)@g' -i $(cat mks)

# check
! grep '$(2)' $(cat mks)
```

I am building 208f5cc8508dd70d3f40fcba255e12cf8f1f9bf3 and 17e40d3884dc4b9db5b521c7e017a3fa12366370 to make sure nothing is broken. 

See https://github.com/mxe/mxe/issues/1452

### DIFF
--- a/plugins/examples/custom-qt-min/overrides.mk
+++ b/plugins/examples/custom-qt-min/overrides.mk
@@ -78,7 +78,7 @@ define qt_BUILD
     ln -sf '$(PREFIX)/$(TARGET)/qt/bin/qmake' '$(PREFIX)/bin/$(TARGET)'-qmake-qt4
 
     mkdir            '$(1)/test-qt'
-    cd               '$(1)/test-qt' && '$(PREFIX)/$(TARGET)/qt/bin/qmake' '$(PWD)/$(2).pro'
+    cd               '$(1)/test-qt' && '$(PREFIX)/$(TARGET)/qt/bin/qmake' '$(PWD)/src/$(PKG)-test.pro'
     $(MAKE)       -C '$(1)/test-qt' -j '$(JOBS)'
     $(INSTALL) -m755 '$(1)/test-qt/release/test-qt.exe' '$(PREFIX)/$(TARGET)/bin/'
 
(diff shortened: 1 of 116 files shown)


starius:
Tested with `build-pkg`. Lists of files are identical (except of `print-git-oneline-*`). I have [uploaded](https://gist.github.com/starius/21286cc89e95ce7d89ce073794e92eee) logs and lists of files to gist. By the way, a gist can be cloned and used as normal git repo - this is how I put binary files there.

tonytheodore:
Looks good! 

Do you think we should use variables like `$(TEST_FILE_C)`, `$(TEST_FILE_CXX)` etc. or does that seem like overkill?

starius:
> Do you think we should use variables like $(TEST_FILE_C), $(TEST_FILE_CXX) etc. or does that seem like overkill?

I think, it is like overkill: most packages have only one test file. Those that have many can use full path like `$(PWD)/src/foo-test.c`

tonytheodore:
Thanks!

Issue #183, opened 2013-04-29
Use git repos instead of tarballs for packages?


starks:
Is this possible? I'd rather not use outdated stable releases.

PR #1432, opened 2016-07-10
add tool skeleton.py

starius:

PR #1470, opened 2016-08-02
Makefile: enable local $(PKG)_SOURCE_TREE

tonytheodore:
This is very handy for testing against a local repo, especially when
upstream changes are happening quickly.

See also #183

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,17 +201,25 @@ PKG_PATCHES   = $(if $(findstring undefined,$(origin $(1)_PATCHES)), \
                     $($(1)_PATCHES))
 
 define PREPARE_PKG_SOURCE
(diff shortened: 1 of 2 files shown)


tonytheodore:
Updated to allow inter-package references to local tree e.g gcc using mingw-w64 source.

starius:
Tested.

starius:
IMHO, rename `$(PKG)_SOURCE_DIR` to `$(PKG)_SOURCE_TREE` to separate it from [SOURCE_DIR](https://github.com/mxe/mxe/pull/1432/commits/1d75e1f3537e7263fa83255fafe4afb89457a6bb) for grep.

tonytheodore:
Renamed, added usage, and squashed.

starius:
Thanks!

PR #1474, opened 2016-08-03
libical: disable parallel build

tonytheodore:
See:
https://github.com/libical/libical/issues/238
https://github.com/libical/libical/issues/233

### DIFF
--- a/src/libical.mk
+++ b/src/libical.mk
@@ -20,7 +20,9 @@ define $(PKG)_BUILD
         -DSHARED_ONLY=$(CMAKE_SHARED_BOOL) \
         '$(SOURCE_DIR)'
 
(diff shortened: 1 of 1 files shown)


starius:
IMHO, build with `-j 1` and add a comment with reference to bug reports in upstream.

tonytheodore:
Indeed, retrying masks some subtle issues - the libs are being built twice.

tonytheodore:
Double-build is a known issue: https://github.com/libical/libical/issues/174

Rebased and changed to disable parallel build.

starius:
Could you add a comment with link to https://github.com/libical/libical/issues/174 above `$(MAKE) -C '$(BUILD_DIR)' -j 1 VERBOSE=1`, please?

tonytheodore:
> Could you add a comment with link to libical/libical#174

Done, it doesn't look like an easy fix.

PR #1477, opened 2016-08-04
cegui: fix libxml2 and minizip; enable fribidi and glfw3

tonytheodore:
Boost and SDL2 aren't detected and need more work, other packages are
either outdated or missing from MXE.


tonytheodore:
`build-pkg` reverses the build order and avoids the `minizip` ordering issue. Other packages can be enabled as people find the time.

PR #1480, opened 2016-08-08
host-tools plugin: install prefixed binutils and strip binaries

tonytheodore:
See https://github.com/mxe/mxe/issues/950#issuecomment-238112271


starius:
Tested on `i686-w64-mingw32.static`.

Thanks!

PR #1488, opened 2016-08-16
Fix MySQL and PostgreSQL posix and win32 build

tonytheodore:
Follow-up to #1009, fixes default win32 gcc threads also. Tested `qt` and `qtbase` on `i686-w64-mingw32.static.posix` and `i686-w64-mingw32.static`.


tonytheodore:
Changed to WIP, failing on win32 shared builds.

tonytheodore:
Succeeds on all 8 targets for qt and qtbase.

PR #1490, opened 2016-08-16
glm: update to version 0.9.7.6

yipdw:
This pull request updates glm to 0.9.7.6 (release notes: http://glm.g-truc.net/0.9.7/updates.html). 

The results of `test-glm.exe` on both `i686-w64-mingw32.shared` and `x86_64-w64-mingw32.shared` are identical to each other and to the 0.9.7.4 release.

### DIFF
--- a/src/glm.mk
+++ b/src/glm.mk
@@ -3,8 +3,8 @@
 
 PKG             := glm
 $(PKG)_IGNORE   :=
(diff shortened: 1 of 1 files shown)


tonytheodore:
Thanks! Building succeeds with all four targets and downstream deps (only cegui).

PR #1491, opened 2016-08-16
physfs: enable x86-64 build

yipdw:
This PR contains a patch (submitted to icculus, awaiting feedback) that works around some pointer-to-integer-cast problems.

The resulting x86-64 `test-physfs.exe` produces identical output to the i686 `test-physfs.exe`; furthermore, the application I'm building with PhysicsFS also works properly with the MXE x86-64 physfs build.

The patch was extracted from some fixes I did on the PhysicsFS Mercurial repository, which is why it's in Mercurial format.  (The MXE pipeline can apply it fine, however.)  If desired, I can change it to conform to git patch format.

### DIFF
--- a/src/physfs-1-cast-fixes.patch
+++ b/src/physfs-1-cast-fixes.patch
@@ -1,0 +1,44 @@
+This file is part of MXE.
+See index.html for further information.
+
+# HG changeset patch
+# User David Yip <dwyip@peach-bun.com>
+# Date 1471367884 18000
+#      Tue Aug 16 12:18:04 2016 -0500
+# Branch stable-2.0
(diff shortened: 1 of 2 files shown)


tonytheodore:
Thanks! Succeeds on all four targets.

`git am` style patches are preferred as they make updates easier to track and complex patch sets easier to understand. In this case, the patch will be removed (in some form) in the next update so it's fine the way it is. Could you add a link in the patch to the upstream issue?

yipdw:
Sure thing.  There's no upstream issue (just a set of commits); would that also work?

tonytheodore:
Perfect, thanks.

PR #1492, opened 2016-08-17
Add options to strip toolchain, executables, and libs

### LINKED ISSUE #985: Option to strip (native) binaries
Is it possible to automatically strip binaries before or after they were installed? Especially the gcc binaries in the libexec directories are very huge and I personally don't need the debugging symbols. Maybe an option in settings.mk like `STRIP_INSTALL := yes`?

### LINKED ISSUE #1249: strip $(PREFIX)/$(TARGET)/bin/test-$(PKG).exe
Some ```test-$(PKG).exe``` files have very large size and not used end users. In some cases after ```strip test-$(PKG).exe``` size decreases to ~10-20 times. For example, the package ```sparsehash``` has used strip https://github.com/mxe/mxe/blob/master/src/sparsehash.mk#L25.
https://github.com/mxe/mxe/issues/985 - this will really very useful option. Currently MXE has stripped by default some libraries (for example qt5 and zlib), but doesn't strip many other libraries, it is very uncomfortably.

tonytheodore:
Minimal implementation to strip the [largest files][lf-gist] by
default, mostly made up of gcc/binutils and test programs.

gdal and geos both produce large libraries, but the libs themselves
aren't worth stripping, it's the 20 odd programs produced by gdal
with those libs statically linked that consume the most space.

I'm leaving these undocumented as the defaults seems reasonable and
the interface may well change when we enable debug/release variants.

closes #985
closes #1249

[lf-gist]:https://github.com/mxe/mxe/issues/1249#issuecomment-193392038

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,11 @@ PKGS       := $(call set_create,\
 BUILD      := $(shell '$(EXT_DIR)/config.guess')
 PATH       := $(PREFIX)/$(BUILD)/bin:$(PREFIX)/bin:$(PATH)
 
+# set to empty or $(false) to disable stripping
+STRIP_TOOLCHAIN := $(true)
+STRIP_LIB       := $(false)
+STRIP_EXE       := $(true)
+
 # All pkgs have (implied) order-only dependencies on MXE_CONF_PKGS.
 # These aren't meaningful to the pkg list in index.html so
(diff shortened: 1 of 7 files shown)


starius:
Tested on i686-w64-mingw32.static
Rebased and merged manually: c255ff8b5a4ab6b097cd3d9f76a28a93dcf510d4

PR #1360, opened 2016-05-26
gcc6: fix building upx

rr-:
Fixes #1359. 

Additionally, I've "tested" Darwin target by swapping arguments to `echo` and it turned out the `-Wno-error=unused-local-typedef` was seemingly wrong, as there's no such option - only `-Wunused-local-typedefs` (note plural). At least according to my `man`. This might be GCC 6 thing.

I haven't tested this on GCC 4 and 5. There's a good chance they don't support `-Wmisleading-indentation`, so it would be great if someone confirmed whether this commit works for them on these versions.

PR #1496, opened 2016-08-20
upx: disable -Werror on native builds

tonytheodore:
See: https://github.com/mxe/mxe/pull/1360/files/3ab014bd5672a1c84c6d93630634a2b79cdb9b8c#r75447401

Tested on OSX clang-700.1.81, gcc6 6.1.0_0, and gcc5 5.4.0_0


tonytheodore:
Confirmed both original error and fix on gcc version 4.8.4 (Ubuntu 4.8.4-2ubuntu1~14.04.3).

PR #1499, opened 2016-08-21
Makefile: fix target validation checks

### LINKED ISSUE #1497: MXE_TARGETS not validate, confusing error on update
After updating from an old version, I only got this error message for anything I did:
Makefile:445: **\* recipe commences before first target.  Stop.
This happened because of this line from the old settings.mk:
MXE_TARGETS := i686-w64-mingw32
Changes in the settings.mk format should be tested to either be auto-converted or create an understandable error message, which would probably also be useful in case of typos in the settings.

tonytheodore:
fixes #1497

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,8 @@ include $(EXT_DIR)/gmsl
 
 MXE_TRIPLETS       := i686-w64-mingw32 x86_64-w64-mingw32
 MXE_LIB_TYPES      := static shared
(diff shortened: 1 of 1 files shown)


tonytheodore:
Changed to WIP pending #1500/#1503

tonytheodore:
Rebased and removed reference to `index.html`, now only prints the valid alternate targets.

PR #1504, opened 2016-08-24
fix MXE_GET_GITHUB_TAGS

starius:
Error message:

    $ make update-package-pire
    ...
    bash: -c: line 0: syntax error near unexpected token `|'
    ...

Overlooked in b52d3c0c9c24e7904908dc50f0c7c6f163556fab
See https://github.com/mxe/mxe/pull/1439


tonytheodore:
Thanks!

PR #1503, opened 2016-08-22
Prepare source tree for pointing GitHub Pages to master docs/

starius:
See https://github.com/mxe/mxe/issues/1500

PR #1505, opened 2016-08-25
add native package ragel and use it in hyperscan

starius:
[Ragel](https://www.colm.net/open-source/ragel/) is a tool like bison. Increasing number of packages use it, so it is worth adding it as a part of build tool set.

I also removed large patch of hyperscan which was needed not to ragel.

New file `ragel.mk` was added with [new style Copyright header](https://github.com/mxe/mxe/pull/1503).


src/hyperscan.mk
```diff
@@ -9,6 +9,7 @@ $(PKG)_SUBDIR   := $(PKG)-$($(PKG)_VERSION)
 $(PKG)_FILE     := $($(PKG)_SUBDIR).tar.gz
 $(PKG)_URL      := https://github.com/01org/$(PKG)/archive/v$($(PKG)_VERSION).tar.gz
 $(PKG)_DEPS     := gcc boost
+# $(PKG)_NATIVE_DEPS := ragel
```
tonytheodore: I'll submit a PR for #516 soon, native deps will look something like `$(BUILD)~ragel` (see [gcc](https://github.com/tonytheodore/mxe/blob/2f1d18539ddb333f9595c730e75fba0669d76b77/src/gcc.mk#L12) or [qbittorrent](https://github.com/tonytheodore/mxe/blob/2f1d18539ddb333f9595c730e75fba0669d76b77/plugins/apps/qbittorrent.mk#L11)).

tonytheodore:
Builds successfully on all targets. Only concern is that there's no source repo/bug tracker. From the [website](http://www.colm.net/news/2014/10/24/ragel-now-maintained-by-colm-networks.html):

> As of October 2014, Ragel will be maintained by Colm Networks.  This is a new consulting company founded by Dr. Adrian D. Thurston.

> Since we cannot operate in the open, the git repository for Ragel  will no longer be available. The project will be published as release (and pre-release) tarballs only. On the upside, Ragel will get much more attention.

> The license will remain the same: GPLv2 with an exception for the generated code derived from Ragel source.

src/hyperscan.mk
starius: Currently `$(PKG)_DEPS     := gcc boost $(BUILD)~ragel` does not work:

```
make: *** No rule to make target
`/mxe/usr/i686-w64-mingw32.static/installed/x86_64-unknown-linux-gnu~ragel',
needed by `/mxe/usr/i686-w64-mingw32.static/installed/hyperscan'.  Stop.
```

tonytheodore:
Actually, that's not really our concern - `bzip2` is also a tarball-only release.

Thanks!

starius:
> Only concern is that there's no source repo/bug tracker.

Unfortunately they switched to [the cathedral model](https://en.wikipedia.org/wiki/The_Cathedral_and_the_Bazaar). I did not found any popular fork.

PR #1507, opened 2016-08-25
hyperscan: update from 4.2.0 to 4.3.0

starius:
Changelog: https://github.com/01org/hyperscan/releases/tag/v4.3.0

Patch "simplegrep: open file in binary mode" was merged in upstream: https://github.com/01org/hyperscan/pull/26


tonytheodore:
Succeeds on all four targets, thanks!

PR #1513, opened 2016-08-27
remove qtwebengine

### LINKED ISSUE #1509: Qt WebEngine isn't built (5.7)
See https://github.com/mxe/mxe/issues/747#issuecomment-129187387, same in 5.7:

```
Qt WebEngine on Windows requires MSVC 2013 or MSVC 2015.
QtWebEngine will not be built.
```

tonytheodore:
closes #1509

alternative to #1512


tonytheodore:
Rebased.

I think this is preferable to #1512 as disabling it implies a temporary situation.

starius:
Thanks!

PR #1521, opened 2016-09-11
opencsg: update to 1.4.1

t-paul:
Minor but important upgrade fixing an issue on Intel GPU.

This only updates the makefile and resolves the conflicts applying the qmake project file patch.

From http://opencsg.org/news.html:
_On Intel graphics hardware, the new rendering path in the SCS algorithm in OpenCSG 1.4.0 entirely failed to work. This was apparently due to an incompatibility of the fragment program with the fixed function pipeline used at the vertex stage. OpenCSG 1.4.1 implements a workaround: a vertex program is now used instead. I hope that the rendering bugs on Intel graphics hardware are now really a thing of the past._

### DIFF
--- a/src/opencsg-1-fixes.patch
+++ b/src/opencsg-1-fixes.patch
@@ -2,9 +2,9 @@ This file is part of MXE. See LICENSE.md for licensing information.
 
 Contains ad hoc patches for cross building.
 
-diff -ur OpenCSG-1.4.0.orig/example/example.pro OpenCSG-1.4.0/example/example.pro
---- OpenCSG-1.4.0.orig/example/example.pro	2014-09-15 22:25:33.000000000 +0200
-+++ OpenCSG-1.4.0/example/example.pro	2015-03-22 14:08:27.706916987 +0100
+diff -ur OpenCSG-1.4.1.orig/example/example.pro OpenCSG-1.4.1/example/example.pro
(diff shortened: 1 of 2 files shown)


src/opencsg-1-fixes.patch
```diff
@@ -28,19 +28,20 @@ diff -ur OpenCSG-1.4.0.orig/example/example.pro OpenCSG-1.4.0/example/example.pr
...
  
 -CONFIG += opengl warn_on release
 +CONFIG += opengl warn_on release link_pkgconfig
- INCLUDEPATH += ../include ../
+ INCLUDEPATH += ../include ../ $$INSTALLDIR/include
  CONFIG -= qt
--LIBS += -lGLEW
+-LIBS += -L$$INSTALLDIR/lib -lGLEW
++LIBS += -L$$INSTALLDIR/lib
 +PKGCONFIG += glew glut
```
starius: Is it needed to add -L... to libs? IMHO, changing of pkgconfig implies it. Works for me without that line.

t-paul: This comes from the original `Makefile`, I'm not sure for which scenario it's needed. I just tried to make minimal changes to that file, but if it makes things cleaner I can just remove that.

I've asked the author if he could include that patch upstream and he said he'll consider that, so maybe the patch will not be needed in later versions.

starius:
Looks good to me!
Thanks!

PR #1529, opened 2016-10-02
new package glpk

jweaton:

- install .pc file,
- install bin/test-pkg.exe compiled with flags by pkg-config,
- install .dll to bin/ and .a, .dll.a to lib/,
- use $(TARGET)-cmake instead of cmake,
- build in `$(BUILD_DIR)` instead of `$(SOURCE_DIR)`,
- do not run target executables with Wine,
- do not download anything while building,
- do not install documentation,
- do not install .exe files except test and build systems,

Thanks!


tonytheodore:
added in #1530

PR #1527, opened 2016-09-29
Qt version overlay and VTK update

muellni:
I updated vtk to compile with GCC 6 and came up with some patches:
- make vtk and qwt link to Qt 4 using a plugin overlay `plugins/qt4`. Maybe this would resolve #1300 
- set CMake QT_QMAKE_EXECUTABLE variable to cache, because running `find_package(Qt4)` a second time in CMake somehow forgets the value of QT_QMAKE_EXECUTABLE
- fix VTK GCC 6 compilation, which I did by upgrading the Compiler Version check - this will most likely be fixed in the next VTK release, because I didn't spot the relevant code in https://github.com/Kitware/VTK/tree/master/CMake

I tested
- GCC6 Qt4`make MXE_PLUGIN_DIRS='plugins/gcc6 plugins/qt4'  MXE_TARGETS='i686-w64-mingw32.shared i686-w64-mingw32.static' JOBS=8 -j4 vtk qwt`
- GCC6 Qt5`make MXE_PLUGIN_DIRS=plugins/gcc6  MXE_TARGETS='i686-w64-mingw32.shared i686-w64-mingw32.static' JOBS=8 -j4 vtk qwt`
- GCC4 Qt5`make MXE_TARGETS='i686-w64-mingw32.shared i686-w64-mingw32.static' JOBS=8 -j4 vtk qwt` **Failed**, because my qtbase won't build using GCC 4 with some PCH related error (logfile 
  [qtbase_i686-w64-mingw32.shared.zip](https://github.com/mxe/mxe/files/500516/qtbase_i686-w64-mingw32.shared.zip)
  )
- GCC4 Qt4 `make MXE_PLUGIN_DIRS=plugins/qt4  MXE_TARGETS='i686-w64-mingw32.shared i686-w64-mingw32.static' JOBS=8 -j4 vtk qwt` **Failed**, same PCH problem (logfile 
  [qt_i686-w64-mingw32.static.zip](https://github.com/mxe/mxe/files/500548/qt_i686-w64-mingw32.static.zip))

PR #1534, opened 2016-10-05
pcl: update from 1.7.2 to 1.8.0

starius:
Patch pcl-1-cmake-policy-cmp0017.patch is not needed.
See https://github.com/PointCloudLibrary/pcl/commit/51a230f07eb9738ed

New option -DHAVE_SSSE3_EXTENSIONS_EXITCODE=0 is needed to fix this:

    CMake Error: TRY_RUN() invoked in cross-compiling mode,
    please set the following cache variables appropriately:
    HAVE_SSSE3_EXTENSIONS_EXITCODE (advanced)

Update of vtk to 7 depends on updating pcl to 1.8.0,
because pcl 1.7.2 fails with vtk 7.
See https://github.com/mxe/mxe/pull/1527#issuecomment-251645448


starius:
Tested with i686-w64-mingw32.static

PR #1541, opened 2016-10-22
update default gcc to 5.4.0

starius:
The bug in gcc 5 which was a reason to downgrade to gcc 4 was fixed. See https://github.com/mxe/mxe/issues/964 for more details.

I am building all packages. If everything is OK in the first pass of build-pkg, I will remove WIP from the subject.


starius:
Problems:

cloog fails to build: https://gist.github.com/f550a61371f597a213f0291e53e632f0
itk fails to build: https://gist.github.com/6c2bdd7f56cd78f2172844cf52e1cd4e

starius:
I fixed 'too many sections' error of itk, but now it fails with undefined reference errors:
https://gist.github.com/fe54b59131933ebdb51979b8707cabbb

starius:
cloog failure was fixed by updating it.

starius:
I reported ITK problem to upstream:
https://cmake.org/pipermail/insight-users/2016-October/054015.html

starius:
Latest ITK (4.10.1) builds successfully after [disabling try-run](https://gist.github.com/starius/b7b1ee84faa76de1f52efe3f1959dc1d).

starius:
I am building all packages with build-pkg after applying this patch.

Revealed problems:
[qtwebkit_i686-w64-mingw32.shared](https://gist.github.com/ae3865a7e19c352f5635dda9f42662b3)
[gtkmm2_x86_64-w64-mingw32.static](https://gist.github.com/7b06e3dfa716d2d27ed3ef494b9e08b7)
[gtkmm3_x86_64-w64-mingw32.static](https://gist.github.com/fb95104b48aef172a30fd6ed45f17251)

starius:
gtkmm2 and gtkmm3 problems were fixed in https://github.com/mxe/mxe/issues/1609

starius:
qtwebkit failure was not found when I built all packages in October. I looked at changes since October and found 634d80dc56534cd4fb6a9cec3180f1dda28b51c5 to be a possible reason. I am building qtwebkit with gcc 5.3.0 to confirm this.

starius:
qtwebkit builds successfully with gcc 5.3.0.

The same error message in qtwebkit was observed here: https://sourceforge.net/p/mingw-w64/mailman/message/35080654/ and were explained by an "issue in the mingw-w64 headers". They used mingw-w64 of 5.0.0rc2. Maybe it was fixed in 5.0.0. My next attempt will be to update mingw-w64 to 5.0.0 and see what will happen.

starius:
Updating mingw-w64 did not help.
Build [failed](https://gist.github.com/e3bfe83f91c1dba117843374e8167bc5) with the same error.

tonytheodore:
@starius, could you try removing the gcc-1.patch (at least part 2)? The [intrinsics fix mentioned](https://sourceforge.net/p/mingw-w64/code/6303/) in #1082 seems related to the `_mm_mfence` error. (I'm having other build failures with `qtbase` and can't test further)

starius:
qtbase is broken here as well: https://gist.github.com/starius/361e08e501155d06e18642458168366d

tonytheodore:
qtbase fixed in #1638, testing gcc5 now.

tonytheodore:
Same failure for qtwebkit without the gcc patch.

tonytheodore:
qtwebkit fix https://github.com/LuaAndC/mxe/pull/9

tonytheodore:
hyperscan [fails](https://gist.github.com/tonytheodore/625f4f271bc5503022ea543f437b11d5#file-hyperscan_i686-w64-mingw32-shared) with a [seemingly related intrinsics error](https://gist.github.com/tonytheodore/625f4f271bc5503022ea543f437b11d5#file-cmakeerror-log).

starius:
Hyperscan used to build some time ago in this pull request.
Current failure might be caused by https://github.com/mxe/mxe/pull/1621 which is a bit fragile.

starius:
By the way, hyperscan builds on i686-w64-mingw32.static on Debian Jessie 64 bit.

starius:
hyperscan on i686-w64-mingw32.shared - also good.

@tonytheodore 
Can you check hyperscan on head? Maybe this bug is unrelated to the PR.

tonytheodore:
hyperscan is fine with gcc49 i686-w64-mingw32.static. I can get it to build with gcc54 by adding `-msse4` and `-latomic` flags.

starius:
I don't understand why it is different on Linux and on macOSX.

starius:
qtwebkit is fixed on Debian in afae60c76ff966490bea418bc737347881a71eaf

starius:
@tonytheodore 
Does it work if you use `-mssse3` instead of `-msse4`?
SSSE3 seems to be the minimal requirement for hyperscan.

starius:
I found log entry in my CMakeOutput.log corresponding to failing part in your CMakeError.log:
https://gist.github.com/02a0897df770eb4e4a1a829e388f7549

Interesting part of that is a list of options passed to gcc, but they seem to be the same. The suspicious part of them are `-march=native -mtune=native`. It may "resolve" to different options.

Command `gcc -march=native -Q --help=target` can be used to produce description of what `-march=native` means for this gcc.

For me `/home/mxe/mxe-gcc-5/usr/bin/i686-w64-mingw32.static-gcc -march=native -Q --help=target` returns the following: https://gist.github.com/d0caf4055bfbddbcb687de5f0fa3f797

SSSE3 is enabled there:

```
-mssse3 [enabled]
```

Can you check output of the same command?

tonytheodore:
It seems to need `-msse4` even though the missing declaration is in SSE2.

Was qtwebkit the final issue on Debian?

tonytheodore:
[Interesting output](https://gist.github.com/tonytheodore/719953f8cec78faa8b67c8fabc20b458). 
```
  -mno-sse4                   		[enabled]
[...]
  -msse                       		[disabled]
  -msse2                      		[disabled]
  -msse2avx                   		[disabled]
  -msse3                      		[disabled]
  -msse4                      		[disabled]
  -msse4.1                    		[disabled]
  -msse4.2                    		[disabled]
  -msse4a                     		[disabled]
  -msse5                      		
  -msseregparm                		[disabled]
  -mssse3                     		[disabled]
```

tonytheodore:
[gcc49 has similar output to yours](https://gist.github.com/tonytheodore/719953f8cec78faa8b67c8fabc20b458#file-gcc49).

starius:
Yes, qtwebkit was the only blocker on Debian about a month ago.

I think we should achieve the same output of this command on Mac and Linux,
because it has nothing to do with a build machine and this difference
indicates a bug somewhere in the toolchain.

tonytheodore:
> I think we should achieve the same output of this command on Mac and Linux,

Agreed, though I'm happy to start a new issue for it and not block this. #1613 is a higher priority and we could just make gcc4 a plugin if anyone reports issues.

The final issue is removing the patch (#1082). I think both are safe to remove.

starius:
I added SSE4 flags to Hyperscan.
I am testing all packages with build-pkg for commit 788aeb91bf46e9f32b76121e4c710571d6475276
Can you test on Mac, please?

starius:
GCC 4 plugin added.

tonytheodore:
hyperscan fix https://github.com/LuaAndC/mxe/pull/10

starius:
https://github.com/LuaAndC/mxe/pull/10 merged

starius:
gtkglextmm was broken.
Fixed in https://github.com/LuaAndC/mxe/pull/11 - tested on Linux

starius:
hyperscan is now broken with **gcc4 plugin**:
https://gist.github.com/4e1da5dc7bc3210f13d4a4a3ebc5f10b

tonytheodore:
> hyperscan is now broken with gcc4 plugin

Grrr, that makes sense - `libatomic` was only introduced in gcc5!

I wonder if a reversing patch in the plugin makes sense or if that's just a double workaround. I'm not sure when I'll get a chance to fix #1641 and I don't want to hold this up.

starius:
I've added README for gcc4 plugin.

I am building with plain `make` (1) with and (2) without gcc4 plugin.
If everything but hyperscan builds, I'll merge this.

tonytheodore:
> If everything but hyperscan builds, I'll merge this.

+1

starius:
Everything (but hyperscan) works well.
Rebasing and merging.

tonytheodore:
Nice work, thanks for pushing this forward!

PR #1549, opened 2016-10-23
[WIP] update boost to 1.62

starius:
Patch "Fix {make,jump}_fcontext visibility with mingw" was merged to upstream: https://github.com/boostorg/context/pull/22

Current boost (1.60) is not compatible with openssl 1.1. Boost 1.60 builds successfully, but cpp-netlib [fails](https://gist.github.com/starius/b0a5bc29e0678a4ab153b08d4380fae1) with boost 1.60 and openssl 1.1 because of missing definitions.

I think, we should update boost before updating openssl to 1.1.

WIP because tests are still running.


starius:
blocked on: https://github.com/01org/hyperscan/issues/39

starius:
Shared targets are broken.
Blocked on: https://svn.boost.org/trac/boost/ticket/12205
See https://github.com/boostorg/serialization/pull/42

tonytheodore:
The shared targets are fixed in v1.66, but that breaks some other packages.

tonytheodore:
Replaced with #2043

PR #1556, opened 2016-10-26
gcc5 plugin: update gcc from 5.3 to 5.4

starius:
Gcc 5.3 fails to build with gcc 6.
The error message is:

    cfns.gperf:101:1: error: ‘const char* libc_name_p(const char*, unsigned int)’ redeclared inline with ‘gnu_inline’ attribute

See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69959
See https://github.com/mxe/mxe/pull/1527#issuecomment-256273786


starius:
The sqlite downloading [problem](https://travis-ci.org/mxe/mxe/builds/171554302#L261) is unrelated.

PR #1566, opened 2016-11-10
Upgrade PortAudio to version 19.6.0

mgeier:
I had a little problem at first:

```
$ make portaudio
[using autodetected 2 job(s)]
[check requirements]
Missing requirement: openssl
```

To make it work, I compiled it like this:

    make portaudio DONT_CHECK_REQUIREMENTS=1

But I guess this is unrelated to my changes to the PortAudio package.


starius:
Tested on all four targets. Thanks!

PR #1568, opened 2016-11-11
Enable shared builds for gsl

danielwbn:
I followed this guide: https://github.com/mxe/mxe/wiki/Adding-Shared-Support-to-Packages

Also I followed the pre-commit checklist and made sure the files are there and in the right places.


starius:
Thanks!

PR #1570, opened 2016-11-15
update hyperscan to 4.3.2

starius:
This version of hyperscan adds a fix for boost bug introduced in boost 1.62.0.

See https://github.com/mxe/mxe/pull/1549#issuecomment-255600102
See https://github.com/01org/hyperscan/issues/39#issuecomment-260495946

### DIFF
--- a/src/hyperscan.mk
+++ b/src/hyperscan.mk
@@ -2,8 +2,8 @@
 
 PKG             := hyperscan
 $(PKG)_IGNORE   :=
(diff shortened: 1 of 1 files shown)


starius:
Tested with [boost 1.62](https://github.com/mxe/mxe/pull/1549). Works on static targets. Shared targets are broken by [other bug](https://github.com/mxe/mxe/pull/1549#issuecomment-255601137).

PR #1582, opened 2016-12-02
Update sdl2

tehnick:
Tested for:
* x86_64-w64-mingw32.static and x86_64-w64-mingw32.shared targets built with gcc6
* i686-w64-mingw32.static and i686-w64-mingw32.shared targets built with gcc4.9


starius:
Thanks!

Issue #1422, opened 2016-07-05
Move package metadata from index.html to src/*.mk


tonytheodore:
Following on from #638, I'd like to move all package metadata back to the relevant package makefiles and generate a (say) `.json` manifest to populate the list of packages. With 400 odd packages, it's not feasible to maintain a manually sorted list in html.

First pass of moving "name" and "website" is reasonably straightforward.

The next passes of:
- unifying `build-matrix`
- extending the matrix for future target specs (threads, exceptions, compiler [gcc\* | clang])
- incorporating `tools/install-deps`
- ..."json-ify" all shared lists (doesn't have to be json)

warrant a coherent approach.

@TimothyGu, @TobiX, do you want to tackle this?

starius:
I sent PR https://github.com/mxe/mxe/pull/1602

PR #1602, opened 2016-12-18
Move package metadata from index.html to src/*.mk

starius:
Demo: http://redjohn.tk:2709/#packages

fix https://github.com/mxe/mxe/issues/1422


starius:
Build log: https://gist.github.com/d368c53a35e376e759ca12a54088eace
No packages are broken.

vog:
This looks very good. Thanks for your effort!

In commit 09a390afbf655d01ff0c1e15edc062b78453e563 I wonder why `$(PKG)_DESCR` is left off for some packages. Shouldn't we set it explicitly an empty string, to encourage people filling this out?

starius:
> why $(PKG)_DESCR is left off for some packages

Empty description is interpreted [by Javascript](https://github.com/mxe/mxe/pull/1602/commits/193e4b6c3b78f19da0b9c37f7afb6cc39c42adeb#diff-eae600270579316beb8b8d9b9a5732d4R1155) as the description equal to the package name. More than half of packages have description (or official name, as it is called in index.html) equal to the package name. It worth reducing header size in mk files for such packages. This behavior also encourages using official package name as package name in MXE.

starius:
Rebased.

vog:
Okay, fine for me.

tonytheodore:
Nice work, thanks!

PR #1604, opened 2016-12-18
libdvdetect: add dependency on openssl

starius:
It depends through Requires.private.
Broken build: https://gist.github.com/c1606f613d18fbcdf938941145f5fbbb


nschlia:
That one escaped me. THX

PR #1617, opened 2016-12-28
Update poppler.

saiarcot895:
Update poppler version to 0.50.0, and update patch to cover Qt 5.


starius:
Tested on all four targets. Looks good to me!

There are many HTML files installed to `<target>/share/gtk-doc/html/poppler/`, but they were there before, so this is not a degradation, so it does not block this pull request. I will see how we can address this.

Issue #1288, opened 2016-04-09
some dll files from gstreamer-plugins not installed into $(PREFIX)/$(TARGET)/bin/


pavelvat:
Currently build log of clementine from https://github.com/mxe/mxe/pull/1277 contain messages about not found some gstreamer-plugins into $(PREFIX)/$(TARGET)/bin/:
```
install: не удалось выполнить stat для '/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/bin/libgstasf.dll': Нет такого файла или каталога
install: не удалось выполнить stat для '/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/bin/libgstcdio.dll': Нет такого файла или каталога
install: не удалось выполнить stat для '/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/bin/libgstfaad.dll': Нет такого файла или каталога
install: не удалось выполнить stat для '/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/bin/libgstgdp.dll': Нет такого файла или каталога
install: не удалось выполнить stat для '/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/bin/libgstlame.dll': Нет такого файла или каталога
install: не удалось выполнить stat для '/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/bin/libgstmad.dll': Нет такого файла или каталога
install: не удалось выполнить stat для '/home/pavel/builds/mxe/usr/i686-w64-mingw32.shared/bin/libgstopus.dll': Нет такого файла или каталога

```
List of dll gstreamer-plugins installed into ```$(PREFIX)/$(TARGET)/bin/```:
```
libgstadaptivedemux-1.0-0.dll
libgstadder.dll
libgstalaw.dll
libgstallocators-1.0-0.dll
libgstalphacolor.dll
libgstalpha.dll
libgstapetag.dll
libgstapp-1.0-0.dll
libgstapp.dll
libgstaudio-1.0-0.dll
libgstaudioconvert.dll
libgstaudiofx.dll
libgstaudioparsers.dll
libgstaudiorate.dll
libgstaudioresample.dll
libgstaudiotestsrc.dll
libgstauparse.dll
libgstautodetect.dll
libgstavi.dll
libgstbadbase-1.0-0.dll
libgstbadvideo-1.0-0.dll
libgstbase-1.0-0.dll
libgstbasecamerabinsrc-1.0-0.dll
[...]

PR #1277, opened 2016-03-26
add package clementine_qt4 to mxe/plugins/apps

pavelvat:
Problems:
1) #1269 #1273
2) (non critical) clementine-1-fixes.patch contain workaround that allow 64-bit build, but if left mouse click on menu "Tools->Vizualizations" it lead to crash:

```
-#  ifdef _WIN64
+#  if defined(_WIN64) && !defined(__MINGW32__)
typedef __int64 ptrdiff_t;
```

3) (non critical) It is not clear how to calculate the only required gstreamer plugins, currently music playback leads to crash. It can be fixed by manually copy all gstreamer plugins to subdirectory "gstreamer-plugins" and also copy all gstreamer plugins to directory with clementine.exe and re-run ./copydlldeps.sh.
4) (non critical) clementine-1-fixes.patch contain fix for detecting pthread:

```
 find_library(PTHREAD
-  pthreadGC2
+  pthread
 )
```

 because `clementine` uses pthreads  POSIX 1003.1-2001 from https://www.sourceware.org/pthreads-win32/. Maybe need to add package for Pthreads-w32 in MXE?
5) (non critical) "$(PKG)_BUILD_STATIC =" because some of `clementine` dependencies are not available for targets: `i686-w64-mingw32.static` and `x86_64-w64-mingw32.static`

PR #1622, opened 2017-01-02
fix moving gstreamer dlls (#1288)

pavelvat:
see #1288, https://github.com/mxe/mxe/pull/1277#issuecomment-255899589

### DIFF
--- a/src/gst-plugins-ugly.mk
+++ b/src/gst-plugins-ugly.mk
@@ -25,4 +25,9 @@ define $(PKG)_BUILD
         --docdir='$(1)/sink' \
         --with-html-dir='$(1)/sink'
     $(MAKE) -C '$(1)' -j '$(JOBS)' install
+
+    # some .dlls are installed to lib - no obvious way to change
+    $(if $(BUILD_SHARED),
+        mv -vf '$(PREFIX)/$(TARGET)/lib/gstreamer-1.0/'*.dll '$(PREFIX)/$(TARGET)/bin/'
+    )
 endef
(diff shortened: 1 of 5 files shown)


src/gst-plugins-ugly.mk
```diff
@@ -25,4 +25,8 @@ define $(PKG)_BUILD
         --docdir='$(1)/sink' \
         --with-html-dir='$(1)/sink'
     $(MAKE) -C '$(1)' -j '$(JOBS)' install
+
+    # some .dlls are installed to lib - no obvious way to change
+    $(and $(BUILD_SHARED),
```
starius: * Why `and`, not  `if`?
* Can you add indent, please?
```makefile
$(if ...,
    ...
)
```

pavelvat: Because I had just copied it from gstreamer.mk.
gstreamer.mk  doesn't use a indent for it.

Do you insist on change "and->if"?

starius: > gstreamer.mk doesn't use a indent for it.

I think, we can keep the current version, but it should be fixed someday. Maybe in another mass-cleanup effort like https://github.com/mxe/mxe/pull/1621

> Do you insist on change "and->if"?

I don't insist, but I am curious why `and` is used. It is inconsistent with other code. Is there a rationale of using it rather than `if`?

pavelvat: >Is there a rationale of using it rather than if?

I don't know. These changes was introduced by @tonytheodore in https://github.com/mxe/mxe/pull/1157

pavelvat:
I had changed and->if.

starius:
Looks good to me!

starius:
Thanks!

PR #1623, opened 2017-01-02
add package gst-libav

pavelvat:
gst-libav is required by clementine for playing wma and ape formats.

### DIFF
--- a/src/gst-libav.mk
+++ b/src/gst-libav.mk
@@ -1,0 +1,25 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := gst-libav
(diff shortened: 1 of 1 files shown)


src/gst-libav.mk
```diff
@@ -0,0 +1,25 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := gst-libav
+$(PKG)_WEBSITE  := http://gstreamer.freedesktop.org/
```
starius: Maybe, https://gstreamer.freedesktop.org/modules/gst-libav.html ?

starius:
Tested on all 4 targets.

starius:
Thanks!

PR #1626, opened 2017-01-06
add package qtsparkle (build with Qt5)

pavelvat:
Qt5 version of qtsparkle are required for https://github.com/clementine-player/Clementine/tree/qt5

Problems with linking for static targets: https://gist.github.com/pavelvat/1497b6ce639ff1f12baaad14c92c2c9a
Previously I fixed similar problem by listing all dependent libraries: https://github.com/eiskaltdcpp/eiskaltdcpp/blob/master/eiskaltdcpp-qt/CMakeLists.txt#L268-L291
Maybe anybody to know how to make universal static build (i.e. for MXE Qt5, for 
mxe/plugins/examples/custom-qt-min, for non MXE Qt5) without listing all dependent libraries in CMakeLists.txt?

### DIFF
--- a/src/qtsparkle-1-fixes.patch
+++ b/src/qtsparkle-1-fixes.patch
@@ -1,0 +1,20 @@
+This file is part of MXE. See LICENSE.md for licensing information.
+
+Contains ad hoc patches for cross building.
+
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Pavel Vatagin <pavelvat@gmail.com>
+Date: Sat, 28 Jan 2017 03:05:58 +0300
+Subject: [PATCH] disable build exampleapp
+
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 1111111..2222222 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -34,4 +34,4 @@ SET(CMAKE_C_FLAGS    "${CMAKE_C_FLAGS} -Wall")
+ SET(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -Wall")
+ 
+ add_subdirectory(src)
+-add_subdirectory(exampleapp)
++#add_subdirectory(exampleapp)
--- a/src/qtsparkle-test.cpp
+++ b/src/qtsparkle-test.cpp
@@ -1,0 +1,17 @@
+/*
+ * This file is part of MXE. See LICENSE.md for licensing information.
+ */
+
+#include <qtsparkle-qt5/Updater>
+#include <QWidget>
+#include <QUrl>
+
+int main()
+{
+	QWidget w;
+	qtsparkle::Updater* updater = new qtsparkle::Updater(
+		QUrl("http://www.example.com/sparkle.xml"), &w);
+	updater->SetVersion("1.0");
+
+	return 0;
+}
--- a/src/qtsparkle.mk
+++ b/src/qtsparkle.mk
@@ -1,0 +1,48 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := qtsparkle
+$(PKG)_WEBSITE  := https://github.com/davidsansome/qtsparkle
+$(PKG)_DESCR    := Qt auto-updater lib
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 4c852e57061d7928573afdf88f04f89d85167477
+$(PKG)_CHECKSUM := 6b8500de51c6a8dd402663fed99bced0588e5be50cfe8474f6d3b46f92025934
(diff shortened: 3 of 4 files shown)


pavelvat:
I disabled build of exampleapp.

The build qtsparkle-test.cpp through pkgconfig for i686-w64-mingw32.static is successful:
```
$ ls -l test-qtsparkle.exe 
-rwxr-xr-x 1 pavel pavel 17448960 Jan  6 21:21 test-qtsparkle.exe
```

pavelvat:
use $(PKG)_BUILD_COMMON for packages qtsparkle_qt4 and qtsparkle (similar to https://github.com/mxe/mxe/pull/1646).

Problems:
line for qtsparkle:
```echo 'Requires: Qt5Core Qt5Network Qt5Widgets'```
cannot be shared with line for qtsparkle_qt4:
```echo 'Requires: QtCore QtGui QtNetwork QtXml'```

I tried to use:
```
$(if $(findstring OFF,@qtsparkle_use_qt4@),
echo 'Requires: Qt5Core Qt5Network Qt5Widgets';,
echo 'Requires: QtCore QtGui QtNetwork QtXml';
)
```
It doesn't work.

**Comment by tonytheodore** [comment]
> Maybe anybody to know how to make universal static build (i.e. for MXE Qt5, for
mxe/plugins/examples/custom-qt-min, for non MXE Qt5) without listing all dependent libraries in CMakeLists.txt?

#1645 will help with mxe-based projects and is based on build options so will work for custom-qt-min also. Plugins still need to be specified, I didn't take the time to automatically add these as they cause extra bloat:

https://doc.qt.io/qt-5/plugins-howto.html#static-plugins
> The defaults are tuned towards an optimal out-of-the-box experience, but may unnecessarily bloat the application. It is recommended to inspect the linker command line built by qmake and eliminate unnecessary plugins.

**Comment by starius** [comment]
`i686-w64-mingw32.static` builds here (https://github.com/mxe/mxe/pull/1626/commits/5d50c4a684176875e7deaf1ad84b3ab2fe3bb9bd), the test binary is linked.

**Comment by pavelvat** [comment]
>the test binary is linked.

Because you have already installed qt and qttools. For clean MXE installation ```make qtsparkle_qt4``` will unsuccessful.

Problem https://github.com/mxe/mxe/pull/1626#issuecomment-274979563 is still not solved.

I use similar construction in https://github.com/mxe/mxe/pull/1165:
```
        $(if $(findstring x86_64,$(TARGET)),,
            $(TARGET)-makensis -Dstatic=32 '$(1).build/EiskaltDC++.nsi';
            $(INSTALL) '$(1).build/EiskaltDC++-2.4.0-x86-static.exe'    '$(PREFIX)/$(TARGET)/apps/$(PKG)/nsis'
         )
```
It is work correct.
The difference is that #1165 has used another type of variable: $(TARGET) instead @qtsparkle_use_qt4@.

**Comment by pavelvat** [comment]
If I understand correctly, the command:
```
$(PKG)_BUILD = $(subst @qtsparkle_use_qt4@,OFF, \
               $(subst @qtsparkle_version_suffix@,-qt5, \
               $($(PKG)_BUILD_COMMON)))
```
is parsing ```$(PKG)_BUILD_COMMON``` and replace each entry @qtsparkle_use_qt4@ to the string OFF.
I tried to use various variants with ```findstring```, ```filter```, ```ifeq```, for example:
instead line:
```echo 'Requires: Qt5Core Qt5Network Qt5Widgets'; \```
this line:
```echo 'Requires: $(if $(ifeq(@qtsparkle_use_qt4@,OFF)),Qt5Core Qt5Network Qt5Widgets,QtCore QtGui QtNetwork QtXml)'; \```
But each of them have given either always true or always false for ```make qtsparkle``` and ```make qtsparkle_qt4```.

**Comment by pavelvat** [comment]
I think that right solution is make section ```# create pkg-config file``` twice in ```qtsparkle.mk``` and ```qtsparkle_qt4.mk``` in $(PKG)_BUILD instead $(PKG)_BUILD_COMMON.

**Comment by tonytheodore** [comment]
Have you tried adding a placeholder:
```
echo 'Requires: @qtsparkle_pc_requires@'; \
```
with another `subst` call to populate the modules?

pavelvat:
>Have you tried adding a placeholder

It works correctly. Thanks.

tonytheodore:
> Because you have already installed qt and qttools. For clean MXE installation make qtsparkle_qt4 will unsuccessful.

I can't see any issue from a clean install - was it only related to the cmake-based example app?

Otherwise, this looks ready to merge.

pavelvat:
>I can't see any issue from a clean install

It was only related with line ```'Requires: Qt5Core Qt5Network Qt5Widgets``` in file qtsparkle.pc (produced by qtsparkle_qt4.mk). Package qttools is not installed for clean install ```make qtsparkle_qt4```

tonytheodore:
Thanks!

PR #1636, opened 2017-01-17
lpeg: update checksum

starius:
The package was changed in upstream:
http://lua-users.org/lists/lua-l/2017-01/msg00160.html


tonytheodore:
Thanks!

PR #1638, opened 2017-01-18
gnutls: disable tests

tonytheodore:
tests install a dummy libcrypt32 which clobbers the w32api

see: https://github.com/mxe/mxe/pull/1541#issuecomment-273373973


tonytheodore:
qtbase shared build now completes.

Issue #1356, opened 2016-05-24
Is it possible to add package liblastfm in MXE from non original git repo?


pavelvat:
https://github.com/davidsansome/liblastfm

tonytheodore:
There's nothing wrong in principle with using a fork, however both appear to be unmaintained. How about https://github.com/lastfm/liblastfm ?

pavelvat:
>There's nothing wrong in principle with using a fork, however both appear to be unmaintained

Ok. I need liblastfm library for https://github.com/clementine-player/Clementine

https://github.com/davidsansome/liblastfm is maintained by main author of Clementine, it may contain some specific functionality required for Clementine.

PR #1643, opened 2017-01-22
add package liblastfm_qt4

pavelvat:
https://github.com/mxe/mxe/issues/1356
Clementine from Debian has used this library.


tonytheodore:
Very little to review other than the rules can be consolidated https://github.com/mxe/mxe/commit/39a48fd13d30617954dc493044d85a4036b8c2b9.

Thanks!

PR #1630, opened 2017-01-09
add package clementine to mxe/plugins/apps

pavelvat:
required https://github.com/mxe/mxe/pull/1626 (qtsparkle is optional dependency. But required package qttools is listed only in qtsparkle.mk and is not listed in clementine.mk)

problems:
1)build is successful, but crash on launch. For order to workaround it you can copy of settings from clementine-qt4 directory (c:\users\USER_NAME\\.config\Clementine) to clementine-qt5 (c:\users\USER_NAME\\.config\Clementine-qt5). Other way: use patch from https://github.com/clementine-player/Clementine/issues/5297
2)clementine-qt5 currently doesn't play any music files and radio streams (gstreamer related errors).

PR #1644, opened 2017-01-22
add package liblastfm

pavelvat:
Qt5 build.
patch is required to separate headers of lastfm for Qt4 and Qt5 builds.
#1356
Clementine from Debian has used this library.

### DIFF
--- a/src/liblastfm-1-fixes.patch
+++ b/src/liblastfm-1-fixes.patch
@@ -1,0 +1,30 @@
+This file is part of MXE. See LICENSE.md for licensing information.
+
+Contains ad hoc patches for cross building.
+
+From 480e2ec663ef867e7892dbbc624737cae176c717 Mon Sep 17 00:00:00 2001
+From: Dominik Schmidt <dev@dominik-schmidt.de>
+Date: Fri, 6 Feb 2015 01:55:40 +0100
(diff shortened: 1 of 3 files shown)


tonytheodore:
Static builds can be enabled with some pkg-config tricks https://github.com/mxe/mxe/commit/b5cc7857e417e1db3c24351cf2c108c70efb61be, something similar could be done for libechonest. Clementine builds but I can't get it to run.

Thanks!

pavelvat:
>something similar could be done for libechonest.

It is not required, because The Echo Nest was acquired by Spotify, and echonest servers are shut down. echonest support was removed from current master-branch of Clementine. I will make PR to delete package libechonest from MXE since release of Clementine-1.3.2.

Something similar could be done for qtsparkle. I'll do it later.

>Clementine builds but I can't get it to run.

from https://github.com/mxe/mxe/pull/1630#issue-199459252:
build Clementine (Qt5) is successful, but crash on launch. For order to workaround it you can copy of settings from clementine-qt4 directory (c:\users\USER_NAME\\.config\Clementine) to clementine-qt5 (c:\users\USER_NAME\\.config\Clementine-qt5). Other way: use patch from clementine-player/Clementine#5297

Issue #1642, opened 2017-01-22
Qt5 + cmake doesn't seem to work at all


uwehermann:
**TL;DR: MXE+Qt5+cmake doesn't seem to work.**

Hi, I've been trying for quite a while to make sigrok (www.sigrok.org) Windows packages using Qt5 + MXE, without any success unfortunately (our current packages use Qt4, which has also been a challenge to implement as well, but that's somewhat working at least).

For our purposes we're depending on a number of specific things that need to work together: 

- MXE target "i686-w64-mingw32.static.posix" (we need C++11 + threads etc.)
  (and static builds of everything else in MXE and outside MXE, of course)
- cmake (used by our PulseView frontend)
- Qt5 (we cannot and don't want to remain stuck with Qt4 forever, for many different reasons, including features from Qt5 we want, and bugfixes from Qt5 we need)

Within Qt5 we need the following: Qt5Core, Qt5Gui, Qt5Svg and various static plugins (directly or indirectly) such as qsvg, QWindowsIntegrationPlugin and such.

I've spent countless days trying to get a successful build, but I didn't get anywhere near a successful build that will also work at runtime.

So my question is whether I'm doing something wrong or whether there are bugs in MXE e.g. regarding i686-w64-mingw32.static.posix+Qt5+cmake that would prevent this from working?

In order to get something reproducible that demonstrates the problem without too many dependencies on all kinds of other stuff we do in sigrok, I've created this demo repo here:

https://github.com/uwehermann/qt5helloworld

It contains a simple project that I generated in qtcreator on Linux which basically just shows an empty window. I added the simplest-possible CMakeLists.txt file for buiding it, and two shell scripts to demonstrate exactly how I'm building.

- The Linux build script works fine, I get an executable and I can run the executable successfully.
- The MXE script doesn't yield a successful build, there are many linker errors (this is with MXE git as of today, only "make qt5", no Qt4 in the MXE paths anywhere).

I've tried all kinds of things in CMakeLists.txt, the source code itself, random MXE patching, nothing has worked sufficiently well so far.

Can someone please verify that the repo I provided does indeed build fine on Linux and doesn't build with MXE?

If yes, any patches against MXE and/or that repo to make this simple project build and run successfully would be really appreciated. It would at least be a basis for further work (I'm pretty sure the static Qt5 plugins will be the next major issue causing headaches after that).

Please let me know if you need any more info. Thanks!

tonytheodore:
[...]

PR #1645, opened 2017-01-23
[WIP] Qt 4&5: improve cmake static linking

tonytheodore:
tested on i686-w64-mingw32.shared i686-w64-mingw32.static

minimal example for #1642


tonytheodore:
pkg-config libs are now added to the cmake modules so no special changes are required in `CMakeLists.txt` for static linking of dependencies.

Plugins still require changes and there's no way I can see to easily auto-configure them for cmake.

tonytheodore:
Added pkg-config static lib dependencies for Qt4 and unified the tests. Tested qt and qt5 on i686-w64-mingw32.static.posix and i686-w64-mingw32.shared, both build and the test programs run correctly. 

Downstream packages still need testing, but workarounds like the [vtk update](https://github.com/mxe/mxe/pull/1527/commits/718fc84c5f22066879ff305ab253eed8759a7f73) will no longer be necessary.

tonytheodore:
Downstream packages all okay and enabling static builds for new packages is much simpler.

I'm happy with this now and think it's good to merge.

tonytheodore:
Back to WIP pending #1650, patches apply but platform libs have changed.

tonytheodore:
Fixed in #2453

PR #1648, opened 2017-01-29
mingw-w64: add --enable-secure-api

starius:
This is needed to update glib to 2.51.1. It uses strerror_s,
which is guarded by MINGW_HAS_SECURE_API.

See http://win-builds.org/bugs/index.php?do=details&task_id=79
See https://sourceforge.net/p/tdm-gcc/bugs/305/


tonytheodore:
+1 I don't think it will break WinXP compatibility but I think it's not worth spending time on.

Side note: glib 2.51.1 is a development version and we usually stick to the even numbered releases.

starius:
Rebased

tonytheodore:
Thanks!

Issue #1647, opened 2017-01-27
Update Gtk 3


mdickie:
MSys 2 provides a newer version of gtk.

starius:
I tried to [update gtk3](https://gist.github.com/starius/1148df405c5fe0ea51ffec24a9ae0f5e#file-) but it requires updating glib first:
```
Package dependency requirement 'glib-2.0 >= 2.49.4' could not be satisfied.
Package 'glib-2.0' has version '2.44.1', required version is '>= 2.49.4'
```

I am trying to update `glib`.

starius:
Next issue is cairo. gtk3 configure fails with the following message:

```
checking for cairo_win32_surface_create_with_format in -lcairo... no
configure: error: cairo_win32_surface_create_with_format is not found in cairo library
You need a cairo snapshot 1.15.x or 1.16.x release or newer
```

Latest release of cairo is 1.14.8. Cairo [1.15.x's snapshot](https://www.cairographics.org/snapshots/) is available, but snapshots are considered unstable.

starius:
Help is needed.

gtk3 3.22.7 is built but its test [fails with linking errors](https://gist.github.com/7a736cb61f60de9a6bc7f1789f22c535).

My current state is fcbe7e30651b5c3be0cbd95541308cd5cbe393f7 + https://gist.github.com/637ee595aa389b31b9c0fe1525c8f254

I traced symbol `_imp__epoxy_glViewport`. It comes from package `libepoxy`, file `usr/i686-w64-mingw32.static/include/epoxy/wgl_generated.h`. The function is declared with `EPOXY_IMPORTEXPORT`, which comes from `usr/i686-w64-mingw32.static/include/epoxy/gl.h` and is defined to `__declspec(dllimport)`. It seems to be responsible for adding `_imp_` prefix to the symbol, breaking the linking. This is where the help is needed. Why is it defined to `__declspec(dllimport)` on a static target? How to the the error?

tonytheodore:
Something like the below (untested) is usually required:

```diff
diff --git a/include/epoxy/gl.h b/include/epoxy/gl.h
index 0be3a66..2a40766 100644
--- a/include/epoxy/gl.h
+++ b/include/epoxy/gl.h
@@ -71,7 +71,15 @@ extern "C" {
 #endif

 #ifndef EPOXY_IMPORTEXPORT
-#define EPOXY_IMPORTEXPORT __declspec(dllimport)
+#  ifndef EPOXY_STATIC
[...]

PR #1652, opened 2017-01-31
libepoxy: fix linking in static targets and add test

starius:
See https://github.com/mxe/mxe/issues/1647#issuecomment-275967915

### DIFF
--- a/src/libepoxy.mk
+++ b/src/libepoxy.mk
@@ -19,7 +19,11 @@ endef
 
 define $(PKG)_BUILD
     cd '$(1)' && autoreconf -fi -I'$(PREFIX)/$(TARGET)/share/aclocal'
-    cd '$(1)' && ./configure \
+    cd '$(1)' && \
+        CFLAGS='$(if $(BUILD_STATIC),-DEPOXY_STATIC,-DEPOXY_SHARED -DEPOXY_DLL)' \
+        ./configure \
         $(MXE_CONFIGURE_OPTS)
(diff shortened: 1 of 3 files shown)


src/libepoxy.mk
```diff
@@ -19,7 +19,17 @@ endef
 
 define $(PKG)_BUILD
     cd '$(1)' && autoreconf -fi -I'$(PREFIX)/$(TARGET)/share/aclocal'
-    cd '$(1)' && ./configure \
+    cd '$(1)' && \
+        CFLAGS='$(if $(BUILD_STATIC),-DEPOXY_STATIC,-DEPOXY_SHARED -DEPOXY_DLL)' \
+        ./configure \
         $(MXE_CONFIGURE_OPTS)
     $(MAKE) -C '$(1)' -j '$(JOBS)' install $(MXE_DISABLE_CRUFT)
+    $(SED) 's/Cflags:/Cflags: -DEPOXY_$(if $(BUILD_STATIC),STATIC,SHARED)/' \
```
tonytheodore: Does the *.pc file also want `-DEPOXY_DLL`

starius: `-DEPOXY_DLL` means "we are building the DLL now", so it should be set while building the package itself (resulting in `dllexport`) and unset when the package is used by other packages (resulting in `dllimport`).

starius:
Rebased

src/libepoxy.mk
tonytheodore: Thanks for the explanation, it never sticks in my mind.

PR #1653, opened 2017-01-31
update gtk3 to 3.22.7, glib to 2.50.2, cairo to snapshot 1.15.4

starius:
fix https://github.com/mxe/mxe/issues/1647


src/glib.mk
```diff
@@ -4,16 +4,16 @@ PKG             := glib
...
+$(PKG)_CHECKSUM := be68737c1f268c05493e503b3b654d2b7f43d7d0b8c5556f7e4651b870acfbf5
 $(PKG)_SUBDIR   := glib-$($(PKG)_VERSION)
 $(PKG)_FILE     := glib-$($(PKG)_VERSION).tar.xz
 $(PKG)_URL      := http://ftp.gnome.org/pub/gnome/sources/glib/$(call SHORT_PKG_VERSION,$(PKG))/$($(PKG)_FILE)
 $(PKG)_DEPS     := gcc dbus gettext libffi libiconv pcre zlib
 
 define $(PKG)_UPDATE
     $(WGET) -q -O- 'http://git.gnome.org/browse/glib/refs/tags' | \
-    $(SED) -n "s,.*tag/?id=\([0-9]\+\.[0-9]*[02468]\.[^']*\).*,\1,p" | \
+    $(SED) -n 's,.*glib-\([0-9][^"]*\)\.tar.*,\1,p' | \
```
tonytheodore: Shouldn't we keep the filter for even numbered (release) versions? `glib` is rarely trivial to update and I wouldn't want to cause any wasted time on a dev version.

src/cairo.mk
```diff
@@ -3,11 +3,11 @@
 PKG             := cairo
 $(PKG)_WEBSITE  := http://cairographics.org/
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 1.14.6
-$(PKG)_CHECKSUM := 613cb38447b76a93ff7235e17acd55a78b52ea84a9df128c3f2257f8eaa7b252
+$(PKG)_VERSION  := 1.15.4
```
tonytheodore: Using the snapshot seems fine, from the release notes:
```
This new snapshot incorporates changes over the past year since the
1.15.2 snapshot, including all the fixes from the 1.14 release series.
```

src/glib.mk
starius: Fixed. The change of updater was moved to a separate commit.

src/cairo.mk
starius: Ok, then we can go with a snapshot.

starius:
Rebased

tonytheodore:
Thanks!

PR #1607, opened 2016-12-22
apps: add ricochet 1.1.4

starius:

PR #1656, opened 2017-02-02
ricochet depends on specific parts of Qt5, not all

starius:
The list of components was deduced from official build instructions:
https://github.com/ricochet-im/ricochet/blob/master/BUILDING.md#fedora

See https://github.com/mxe/mxe/pull/1607#pullrequestreview-19498314


tonytheodore:
Thanks!

PR #1657, opened 2017-02-03
boost: add test for boost.context linking

tonytheodore:
See http://lists.nongnu.org/archive/html/mingw-cross-env-list/2017-02/msg00002.html

Appears to build and run:
image


starius:
Tested on 64 bit Linux (wheezy).
The test works under wine.

starius:
Thanks!

PR #1668, opened 2017-02-08
add qtspeech module

tonytheodore:
See http://lists.nongnu.org/archive/html/mingw-cross-env-list/2017-02/msg00013.html

### DIFF
--- a/src/qtspeech.mk
+++ b/src/qtspeech.mk
@@ -1,0 +1,22 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := qtspeech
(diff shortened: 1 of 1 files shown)


tonytheodore:
Appears to be disabled for mingw: http://lists.nongnu.org/archive/html/mingw-cross-env-list/2017-02/msg00015.html

tonytheodore:
Seems to be working now: http://lists.nongnu.org/archive/html/mingw-cross-env-list/2017-07/msg00000.html and builds fine for `i686-w64-mingw32.static`.

PR #1671, opened 2017-02-12
upgrade to postgresql 9.6.2

mabrand:
It's been a long time since MXE last saw a postgresql upgrade. This new version builds fine without the old patches. I'm not sure about which hacks in the .mk file could be removed.


mabrand:
This seems to have been merged automatically by MXEBot, which I did not expect. Did I do something wrong?

tonytheodore:
Last time I looked at updating it (~9.4.5), the only patches that were still required were the static/shared ones - though they may have been fixed in recent versions.

Not sure about the PR, some new github feature I guess.

mabrand:
Do you think this happened because the source for the pull request was in the same repo?

tonytheodore:
> pull request was in the same repo?

Possibly a combination of that and something to do with `travis.yml`?

tonytheodore:
> still required were the static/shared ones

The shared libs seem to be installed in the correct locations now, but the shared lib is still built for static targets.

mabrand:
Do you see an easy way to prevent that? I've tried various things, but I'll spare you the details.

tonytheodore:
No easy way last time I looked at it - the [relevant parts involve minimal patching](https://gist.github.com/tonytheodore/aa9b91e673169a6832a6fb22d04bf2ef) but the corresponding builds rules are too arcane.

mabrand:
I haven't been able to crack this nut yet.  https://github.com/mxe/mxe/issues/1676 reported the postgresql update broke qtbase. Reverting it for now seems like the best option.  713c5ecad7dcc51679e218ac0ace1f1d51a6b97a

PR #1672, opened 2017-02-12
libsamplerate: use BUILD_DIR/SOURCE_DIR, disable progs and docs

tonytheodore:


tonytheodore:
@mabrand, this was also just auto-merged by MXEBot, must be caused by pushing to mxe repo.

PR #1675, opened 2017-02-15
glib 2.50.2 native build fails, libmount dependency missing

phillipberndt:
Recent versions of glib require libmount on Linux. If it is unavailable,
configure fails with

    checking libmount/libmount.h presence... no
    checking for libmount/libmount.h... no
    configure: error: *** Could not find libmount

This commit disables libmount by patching `--disable-libmount` into the configure command line.

(FYI, the windows build is unaffected because libmount is only enabled by default on Linux.)


tonytheodore:
Thanks!

PR #1677, opened 2017-02-17
Minor updates

tehnick:


tehnick:
Also could you remove libsamplerate-tidy branch if it is not required anymore?

starius:
Tested on all four targets.

The only new file installed is `<target>/bin/ffplay.exe` (in addition to existing `ffmpeg.exe` and `ffprobe.exe`). It works fine for me.

starius:
Thanks!

PR #1678, opened 2017-02-17
index.html: add Sibuserv to the list of projects which use MXE

tehnick:


starius:
Looks good! Maybe add a short description "Cross-compilation build server"?

tehnick:
I don't think this is necessary. At least not now. Maybe one day, when I'll finish and release Web UI and the project will become a complete continuous integration system.

starius:
Thanks!

Issue #1629, opened 2017-01-08
Download failed or wrong checksum of package nettle


AlekseyDurachenko:
Build failed: "Download failed or wrong checksum of package nettle!"
Logs: http://pastebin.com/j8M8Hhww
Version: d29230191319d993fba5f1e2fa1adb3ef804c2ae

PR #1681, opened 2017-02-18
add script updating packages backup from cron

starius:


tonytheodore:
Should this be added to `tools/s3-fetch-and-sync` or the normal backup cron?

P.S. You can modify the backup cron with the `mxe` user:
```
ssh starius@mirror.mxe.cc
sudo su mxe
crontab -e
```

starius:
> Should this be added to tools/s3-fetch-and-sync or the normal backup cron?

Ideally these two should be merged, but there things to do first:

  * file naming scheme. [The backup managed by update_backup.{py,sh}](https://gitlab.com/starius/mxe-backup2/) uses different filename scheme (original filename + "_" + checksum) to avoid errors if a file changes without changing filename. Can we change tools/s3-fetch-and-sync to do so as well (providing current filenames as well for existing files for backwards compatibility)?
  * Backup service should be stateless. Currently deploying new instance of it requires downloading files from a cloud provider (I wrote script tools/backup_from_s3.py for that). I think, using a central git for tracking state of the backup is a good idea. Moreover, https://gitlab.com/starius/mxe-backup2/ was intended to be such a repo, not to serve files (but [happened](https://github.com/mxe/mxe/pull/1628) to serve files as there was [no other alternative](https://github.com/mxe/mxe/issues/1629)).
  * Separate machines and different clouds to avoid single failure domain. If one machine (or file hosting) is down, another one will backup new files and provide them for download. After we merge file backup code we can add another cloud provider in addition to AWS to host files (e.g. GCP). Then we can make sure that each commit of MXE is and will be downloadable in future.

Issue #1674, opened 2017-02-13
behaviour MXE_GET_GITHUB_TAGS


pavelvat:
I have detected problem with update ```libsoup```. After replace line ```$(PKG)_IGNORE := libsoup-pre%``` to ```$(PKG)_IGNORE :=``` the command ```make update-package-libsoup``` output:
```
NEW      libsoup  2.54.0.1 --> libsoup-pre214-branch-base
```
The reason of that: https://github.com/mxe/mxe/blob/master/Makefile#L176
If replace ```$(SORT) -V``` to  ```$(SORT) -n``` then the command ```make update-package-libsoup``` output:
```
NEW      libsoup  2.54.0.1 --> 2.57.1
```

Full string for tests:
```
wget -q -O- 'https://api.github.com/repos/GNOME/libsoup/git/refs/tags/' | sed -n 's#.*"ref": "refs/tags/\([^"]*\).*#\1#p' | sed 's,^,,g' | sort -n | tail -1
```

starius:
With `$(PKG)_IGNORE := libsoup-pre%` (master version), the output is
```
IGNORED  libsoup  libsoup-pre214-branch-base
```

Probably ignored versions should be sorted out before `head -1` in `define UPDATE`

pavelvat:
look at output this command:
```
wget -q -O- 'https://api.github.com/repos/GNOME/libsoup/git/refs/tags/' | sed -n 's#.*"ref": "refs/tags/\([^"]*\).*#\1#p' | sed 's,^,,g' | sort -n
```

I think that need to add a toggle "sort type" to call of function:
```
define $(PKG)_UPDATE
    $(call MXE_GET_GITHUB_TAGS, repo/libname)
endef

```

starius:
> look at output this command:

There are multiple types of strings there:
  * gnome-2-12-base
  * LIBSOUP_1_99_20
  * libsoup-2-0-branch-base
  * LIBSOUP_2_28_2_THIS_TIME_FOR_REAL
  * libsoup-hacking-branch-base
  * SOUP_0_5_4
  * 2.51.90
  * 2.49.91.1

The tag naming system is organized not ideally. We could deal with that by adding additional filtering (like `grep -e '^[0-9]\+\.[0-9]\+\.[0-9]\+$'`). It can be applied after `sed` and before `sort`.

tonytheodore:
[...]

PR #1686, opened 2017-02-23
update jsoncpp

pavelvat:
The section $(PKG)_UPDATE was updated (previously it didn't work). After this we have the same problem with  ```$(SORT) -V``` <-> ```$(SORT) -n``` as in https://github.com/mxe/mxe/issues/1674.


starius:
Thanks!

PR #1690, opened 2017-03-04
URL updates

vszakats:
Mostly with the goal of securing the transport.

Also removed two exec attributes set for non-executable files.


starius:
https://www.xvid.org/ doesn't work. Can you replace it with https://www.xvid.com/ ?

starius:
https://ftp.sunet.se/pub/multimedia/graphics/ImageMagick/ImageMagick-6.9.0-0.tar.xz doesn't work (as well as previous URL, http://ftp.sunet.se/pub/multimedia/graphics/ImageMagick/ImageMagick-6.9.0-0.tar.xz). I think we can drop this backup URL.

vszakats:
@starius Done, thanks a lot for noticing this!

vszakats:
@starius `imagemagick` keeps only the latest revision there, so this is expected - but unrelated to the protocol upgrade itself. I think this is best to be addressed in a separate commit.

starius:
https://www.cloog.org/ is empty.

starius:
@vszakats 
Thanks for updates!

https://ftp.sunet.se/pub/multimedia/graphics/ImageMagick/ is also 404. I think, the site stopped mirroring ImageMagick releases.

vszakats:
I'm fixing `cloog` as well, thanks!
This is a known working URL for Imagemagick:
https://www.imagemagick.org/download/ImageMagick-7.0.5-0.tar.xz
From:
https://github.com/Homebrew/homebrew-core/blob/master/Formula/imagemagick.rb

starius:
I checked all changed URL. They look OK.

vszakats:
Thank you! I may do some more upgrades later if you don't mind, these were just the more obvious ones. `cloog` restored to HTTP.

starius:
Thanks!

PR #1694, opened 2017-03-05
More URL updates

vszakats:


vszakats:
That's all for the moment.

starius:
gmsl.html is synced from upstream [cvs](https://sourceforge.net/p/gmsl/code/) so changes of this file from this pull request will be lost when it is synced next time. Can you send the patch to the upstream, please? https://sourceforge.net/p/gmsl/_list/tickets

vszakats:
Reverted those.

Here's the diff for reference, it's only minor changes, so I think it's fine as it is.
```diff
--- gmsl.html	2017-03-07 15:51:48.000000000 +0000
+++ gmsl-new.html	2017-03-05 11:29:35.000000000 +0000
@@ -10,8 +10,8 @@
 string manipulation, integer arithmetic, associative arrays, stacks,
 and debugging facilities.&nbsp; The GMSL is released under the BSD License.<br>
 <br>
-<a href="http://sourceforge.net/projects/gmsl/">[Project Page]</a> <a href="http://sourceforge.net/project/showfiles.php?group_id=129887">[Download]</a>
-<a href="http://sourceforge.net/forum/forum.php?forum_id=443916">[Discussion
+<a href="https://sourceforge.net/projects/gmsl/">[Project Page]</a> <a href="https://sourceforge.net/projects/gmsl/files/">[Download]</a>
+<a href="https://sourceforge.net/p/gmsl/discussion/443916/">[Discussion
 Forum]</a><br>
 <h2>Using GMSL</h2>
 The two files needed are <span style="font-family: monospace;">gmsl</span>
@@ -696,18 +696,18 @@
 </table>
 <span style="font-family: monospace;"></span><br>
 <hr>
-Copyright (c) 2005-2014 <a href="http://www.jgc.org/">John Graham-Cumming</a>.<br>
+Copyright (c) 2005-2014 <a href="https://www.jgc.org/">John Graham-Cumming</a>.<br>
 <hr style="width: 100%; height: 2px;">
 <table style="width: 100%; text-align: left;" border="0" cellpadding="2" cellspacing="2">
   <tbody>
     <tr>
       <td style="width: 50%;">John Graham-Cumming's work on this
-project was sponsored by <a href="http://www.electric-cloud.com/">Electric
+project was sponsored by <a href="https://electric-cloud.com/">Electric
 Cloud, Inc</a>.<br>
-      <a href="http://www.electric-cloud.com/"><img alt="" src="http://gmsl.sf.net/ec_logo.gif" style="border: 0px solid ; width: 223px; height: 47px;"></a><br>
+      <a href="https://electric-cloud.com/"><img alt="" src="http://gmsl.sf.net/ec_logo.gif" style="border: 0px solid ; width: 223px; height: 47px;"></a><br>
       </td>
       <td align="right">
-      <p><a href="http://sourceforge.net/"><img src="http://sourceforge.net/sflogo.php?group_id=129887&amp;type=1" alt="SourceForge.net Logo" border="0" height="31" width="88"></a></p>
+      <p><a href="https://sourceforge.net/"><img src="https://sourceforge.net/sflogo.php?group_id=129887&amp;type=1" alt="SourceForge.net Logo" border="0" height="31" width="88"></a></p>
       </td>
     </tr>
   </tbody>
```

starius:
https://marathon.sourceforge.io redirects to https://alephone.lhowon.org/

starius:
https://pdfgrep.sourceforge.io redirects to https://pdfgrep.org/

starius:
https://www.xs4all.nl/~thebeez/4tH redirects to https://thebeez.home.xs4all.nl/4tH/

vszakats:
Updated to follow the redirects.

pavelvat:
>armadillo: update sha256

Checking control sum is needed for security. Are you sure that the file armadillo-6.400.3.tar.gz was not compromised?
reason: https://sourceforge.net/p/arma/news/2017/03/-notice-of-removal-of-old-armadillo-releases-/

This PR has too much "garbage" commits. You need squash commits:
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits

vszakats:
@pavelvat Restored the original checksum and updated the URL to point to the correct file. (I have zero information about this project or why they have multiple packages under the same version number.)

vszakats:
I plan to squash commits once any potential problems are found and fixed. (GitHub now also has a "Squash and Merge" option at merge time, so it can be done automatically.)

vszakats:
Squashed now.

starius:
config.guess is also external as well as gmsl

starius:
For the record:
https://gsoap2.sourceforge.io/ redirects to http://gsoap2.sourceforge.io which redirects to http://www.genivia.com/dev.html which timeouts.
I think, we can do nothing about it.

starius:
https://harfbuzz.sourceforge.io/ returns 404

starius:
Please rebase the PR to resolve conflicts.

starius:
https://hci.iwr.uni-heidelberg.de/vigra also returns 404.

vszakats:
I just noticed that all `wget` download commands had certificate checks disabled via the `--no-check-certificate` option. This effectively disables the authentication feature of HTTPS, so it is considered a severe security gap. Since this would make the use of secure connection (and thus this PR) pointless, I've removed this option. Let's see how the tests go. If there is any URL that requires this option, I think it's better to revert/switch to a cleartext HTTP alternative in those specific cases only.

```
$ grep -R '\--no-check-certificate' *
Makefile:WGET        = wget --no-check-certificate \
tools/compat-init.sh:WGET="wget --no-check-certificate
```

starius:
> it is considered a severe security gap

Package files are checked against checksums after downloading, so `--no-check-certificate` is not such a big deal in that place.

starius:
I think, we should re-add `--no-check-certificate` options of `wget` for verified downloading of packages:
```diff
diff --git a/Makefile b/Makefile
index 6fb880e..a9aec5f 100644
--- a/Makefile
+++ b/Makefile
@@ -248,19 +248,19 @@ ESCAPE_PKG = \
 
 BACKUP_DOWNLOAD = \
     (echo "MXE Warning! Downloading $(1) from backup." >&2 && \
-    ($(WGET) -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' $(PKG_MIRROR)/`$(call ESCAPE_PKG,$(1))` || \
-    $(WGET) -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' $(PKG_CDN)/`$(call ESCAPE_PKG,$(1))` || \
-    $(WGET) -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' $(GITLAB_BACKUP)/`$(call ESCAPE_PKG,$(1))`_$($(1)_CHECKSUM)))
+    ($(WGET) --no-check-certificate -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' $(PKG_MIRROR)/`$(call ESCAPE_PKG,$(1))` || \
+    $(WGET) --no-check-certificate -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' $(PKG_CDN)/`$(call ESCAPE_PKG,$(1))` || \
+    $(WGET) --no-check-certificate -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' $(GITLAB_BACKUP)/`$(call ESCAPE_PKG,$(1))`_$($(1)_CHECKSUM)))
 
 DOWNLOAD_PKG_ARCHIVE = \
     $(if $($(1)_SOURCE_TREE),\
         true\
     $(else),\
         mkdir -p '$(PKG_DIR)' && ( \
-            $(WGET) -T 30 -t 3 -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' '$($(1)_URL)' \
+            $(WGET) --no-check-certificate -T 30 -t 3 -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' '$($(1)_URL)' \
             $(if $($(1)_URL_2), \
                 || (echo "MXE Warning! Downloading $(1) from second URL." >&2 && \
-                    $(WGET) -T 30 -t 3 -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' '$($(1)_URL_2)')) \
+                    $(WGET) --no-check-certificate -T 30 -t 3 -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' '$($(1)_URL_2)')) \
             $(if $(MXE_NO_BACKUP_DL),, \
                 || $(BACKUP_DOWNLOAD)) \
         ) && cat '$(PKG_DIR)/.tmp-$($(1)_FILE)' \
```

This doesn't add security risk, because downloaded files are verified against checksums. The advantage of having this option is more reliability on old systems lacking a root cert or some new crypto algorithm used by a webserver. Adding it to `WGET` var is not ideal because it also affects non-verified downloads in `$(PKG)_UPDATE`.

vszakats:
If there are systems that don't have correct CA bundles, I'd argue this option should be enabled explicitly there and only there, but it should operate securely by default. This option won't help with missing crypto algo, as it servers to disable server certificate verification (one pillar of TLS), but leaves the transfer protocol encrypted.

The checksum is good, but it's only good as long as you verify it against a value you received from a verified source. If you download a package via HTTP or `--no-check-certificate` and calculate the checksum locally, it doesn't fulfil it's intended purpose.

vszakats:
Notice that by using `--no-check-certificate`, the client will not verify if the certificate offered by the server matches the domain name at all, nor will it verify if it's a valid certificate. It means that any certificate will be accepted, even self-signed one, or one created for any other site, or any invalid ones.
This, in effect it means that the "encryption" in HTTPS is simple security theatre and almost just as insecure as cleartext HTTP or FTP. But, it's even worse because everything _looks_ secure, except that nothing is.

starius:
> The checksum is good, but it's only good as long as you verify it against a value you received from a verified source.

The checksums are in the MXE source in all `src/*.mk` files. See also the definition of [CHECK_PKG_ARCHIVE](https://github.com/mxe/mxe/blob/49ab4c13ea17aa0830aa933386c75ced9d27c5cb/Makefile#L239) macro.

Try to change `$(PKG)_CHECKSUM` field in any `src/*.mk` file and run `make download-only-<that-pkg>` to make sure that the check works.

vszakats:
I understand that. The question is how these checksums are formed/calculated? If they are calculated after getting the package via an insecure link, it can be spoofed/compromised.

But, it's not only the checksums that are a concern here. When using `--no-check-certificate`, anybody will also be able to listen to and decrypt your traffic, it's enough to create a fake certificate for HTTPS requests. While the user believes that things are encrypted and secure.

vszakats:
`--no-check-certificate` is in effect implementing in `wget` the infamous 'goto fail' bug found in Apple's HTTPS stack 3 years ago: <https://www.imperialviolet.org/2014/02/22/applebug.html> <https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/>

starius:
> The question is how these checksums are formed/calculated?

That is a good point! In majority of cases they are calculated with `make update-checksum-<pkg>` or `make update-package-<pkg>` both of which depend on downloading of packages. It is not sufficient just to securely download HTML pages in `$(PKG)_UPDATE` -- we need also securely download package files themselves. So my patch above is sub-optimal, because it makes downloading of package files themselves in `make update-checksum-<pkg>` insecure.

Ideally we should download securely in context of `make update-checksum-<pkg>` and with `--no-check-certificate` in regular downloads. This seems to be non-trivial and can be moved to other pull-request (and this one will be only about switching URLs to https). What do you think?

vszakats:
This sounds better, though I'd personally only enable `--no-check-certificate` on outdated systems where it's actually required, regardless of the purpose of the download. I don't believe this option should be used for optimisation purposes.

I've committed a patch here to add support for env var named `MXE_WGET_OPTIONS` where custom options can be passed to `wget` calls. This is still pretty bad, as it allows to silently switch off security, but it can be used on outdated systems to pass `--no-check-certificate`.

starius:
I moved fix of armadillo URL to https://github.com/mxe/mxe/pull/1701 not to mix it with hundreds of URL updates to http.

vszakats:
Would you mind changing the protocols to `https://` in armadillo to avoid conflicting with this update?

tonytheodore:
> outdated systems to pass --no-check-certificate

Can outdated systems fall back to downloading from mirrors (or we can add a native build of wget)?

starius:
> Would you mind changing the protocols to https:// in armadillo to avoid conflicting with this update?

Done

> Ops, a bit more than that, here's the full armadillo patch:

The first change is rather distant, so Git should resolve this automatically.

vszakats:
I'm not sure how CA bundles are passed to `wget`, but it may also a solution to get a fresh CA bundle from a known good source (f.e. <https://curl.haxx.se/ca/cacert.pem>) and explicitly pass that to `wget`. If `wget` doesn't `curl` surely supports such scenario.

starius:
There is no security risk of even using plain HTTP downloads as long as results are verified with checksums. Debian APT works this way.

> envvar named MXE_WGET_OPTIONS where custom options can be passed to wget calls

This is not a backwards-compatible change. Consider an automated system with outdated software. Adding such an option will require a human to figure out about it and to putting `--no-check-certificate` there. Do we really want it? I think, we should remove `--no-check-certificate` for the code updating checksums and keep it for regular downloads. It will add security and will not break backward compatibility.

> Can outdated systems fall back to downloading from mirrors

I really like the idea! If we remove `--no-check-certificate` from `WGET` and add it only to `BACKUP_DOWNLOAD`, then we will support old systems and stop insecure downloads during updating checksums.

> or we can add a native build of wget

IMHO, we can avoid this.

starius:
I merged https://github.com/mxe/mxe/pull/1701
Can you rebase this PR, please?

vszakats:
Debian APT works differently. There, there is digitally signed file (`/apt/debian/dists/wheezy/Release` and `/apt/debian/dists/wheezy/Release.gpg`) that holds the checksums for the `Packages` file, while in turn holds the checksums for the actual `.deb` files, so the transport can be insecure. I'd still hold to the fact that it's not only about checksums, it's also about privacy. I'd prefer my packages downloads to not be visible for anybody else on the same WiFi network, for an obvious example.

vszakats:
As for adding insecurity for `BACKUP_DOWNLOAD`, it's very easy to attack: it's enough to block access to any original source. Also, as per the log, the backup is often used in cases when the primary secure download is perfectly valid secure URL. F.e.: https://travis-ci.org/mxe/mxe/builds/209310747#L184

Probably it'd be good idea to refresh certificates when starting the Travis builds via `sudo update-ca-certificates` (or other secure/official method). This needs `sudo: required` in `.travis.yml`.
It may also help to migrate to Ubuntu 14.04 LTS (Trusty) (from current 12.02), which is offered since last autumn as a beta. This needs `dist: trusty` in `.travis.yml`.

starius:
> or we can add a native build of wget

I wanted to rethink this proposal, but then I realized that it would introduce chicken-and-egg problem.

starius:
> As for adding insecurity for BACKUP_DOWNLOAD, it's very easy to attack: it's enough to block access to any original source.

+1
Downloading from backups should be explicitly disabled while updating checksums.

> Also, as per the log, the backup is often used in cases when the primary secure download is perfectly valid.

Some upstream sites are down sometimes. E.g. sourceforge returns 404 from time to time.

vszakats:
Backup downloads are a good thing. I only argue that they should be downloaded via a secure channel.

starius:
Backup downloads seem useless while updating packages because packages are added to backups after updating them in the MXE source. I moved disabling of --no-check-certificate while updating packages to https://github.com/mxe/mxe/pull/1702

> I'd still hold to the fact that it's not only about checksums, it's also about privacy. I'd prefer my packages downloads to not be visible for anybody else on the same WiFi network, for an obvious example.

Given https://github.com/mxe/mxe/pull/1702 we can add an option to disable --no-check-certificate entirely to support this use-case. However this seems to be different from securing the build process. Even without `--no-check-certificate` those people in the middle can figure out which sites you access from DNS requests, IP addresses and SNI fields of HTTPS. And they [can](https://www.theregister.co.uk/2016/08/29/chinese_ca_hands_guy_base_certificates_for_github_florida_uni/) do man-in-the-middle attack against any site including Github. Given a list of sites you access they can figure out that you use MXE and then get full list of URLs from its source. The only protection from this I know is putting something (Tor/VPN/SSH proxy) between you and sites.

vszakats:
I understand that using HTTPS won't protect from _all_ imaginable threats but it's the current best we can do with ubiquitous and standard technology. Reflecting to your MITM example, WoSign has been since suspended/put on probation for actions like these.

I'd argue that we should use available best practices as far as the HTTPS protocol and current CA system allows. It will still protect from a lot of very obvious attacks and give better privacy than without.

starius:
Can you define the threat model, please? In https://github.com/mxe/mxe/pull/1702 the attack on privacy is possible only if an attacker blocks access to all sites used for downloading packages except those three backup servers. To do that, she already needs knowledge that you use MXE. Given she knows you use MXE, what new information can she receive from MiTM attack on connection between you and backup servers?

vszakats:
A MiTM attacker can see and alter all my traffic cleartext. Including the exact packages and package versions downloaded. It can be revealing information. It can likely be used with something else combined.

I'd reverse the question though: If there are some known outdated systems that have problems handling current live certificates, is it optimal to disable default security for everyone just to address that, or wouldn't it be better to fix said outdated systems (if possible), or enable such workaround on those systems only.

vszakats:
In a test done using Trusty, the number of fallbacks to backup went down from [77](https://travis-ci.org/mxe/mxe/builds/209323078) to 9.

PR: https://github.com/mxe/mxe/pull/1703
Travis log: https://travis-ci.org/mxe/mxe/builds/209388739

vszakats:
Most of those 9 were valid problems, where the URL is broken even when testing it using different method.

Committed a few related homepage fixes and bumped a version in this PR: https://github.com/mxe/mxe/pull/1704

starius:
> Including the exact packages and package versions downloaded.

Given the attacker knows about MXE (which she is likely to know about because she blocked all non-backup servers and kept backup servers unblocked) this information corresponds to MXE commit hash on one-to-one basis. I.e. the attacker knows your MXE commit.

> It can likely be used with something else combined.

Again, if she already know about MXE, she can deploy her attack against all recent MXE commits. A user is likely to use latest head, so if there is an attack benefiting from knowledge of exact MXE commit, it can be successfully used against majority of users. Can you provide an example of such attack, btw?

> If there are some known outdated systems that have problems handling current live certificates, is it optimal to disable default security for everyone just to address that, or wouldn't it be better to fix said outdated systems (if possible), or enable such workaround on those systems only?

If https://github.com/mxe/mxe/pull/1702 hurts security of build for everyone, then it should not be used (and `--no-check-certificate` should be just removed). My point is that with https://github.com/mxe/mxe/pull/1702 there is no more risk for the build than with just removal of `--no-check-certificate`.

vszakats:
#1702 does two different things: It restores security for non-backup downloads, which is perfect and very much welcome. And it removes security from backup downloads, which is however not following good security practices.

starius:
> And it removes security from backup downloads, which is however not following good security practices.

Currently that PR provides plain HTTP downloads (with unprotected HTTPS redirects in case of GitLab) for backup downloads. Downloaded files are verified against checksums. I don't see how it can be insecure in terms of content of downloaded files. If this does not follow security practices, then APT used in Debian also do not follow good security practices.

>  I'd prefer my packages downloads to not be visible for anybody else on the same WiFi network

I thought on this argument a lot and I think that hiding a list of software used by the system can be classified as [security through obscurity](https://en.wikipedia.org/wiki/Security_through_obscurity) which is known to be far from good security practice. Moreover using of strong HTTPS doesn't hide this list from a smart attacker, so it is similar to other well known security-through-obscurity approaches that seem to work but in practice provide very small security if any. I think, we should have clear goals and non-goals in terms of security. Is hiding a list of used packages from ISP one of our goals?

starius:
This pull request looks good to me.

starius:
Many thanks!

vszakats:
Thank you very much for the merge!

On hiding traffic: It's primarily a _privacy_ matter, not a security one. Not revealing more information than necessary is usually good practice. As for 'security through obscurity' it's also not as black and white as it may seem; see some valid points in this article: https://danielmiessler.com/study/security-by-obscurity/.

PR #1702, opened 2017-03-09
disable --no-check-certificate while updating packages

starius:
Regular downloads of packages are verified by checksums, so
--no-check-certificate doesn't compromise the build system,
but the checksums themselves are often updated with update-checksum-%
which in turn calls the regular package download mechanism, so there
is a possibility of downloading and sealing a poisoned file.

On the one hand, old systems may still rely on --no-check-certificate,
so it is not nice to completely disable it for regular downloads.
However keeping this option enabled for backup servers only is enough
to support such systems because of the fallback mechanism.
On the other hand, download from a backup doesn't make sense while
updating a package, because the package is definetely not in the backup yet.

So --no-check-certificate is now enabled only for backup servers
and backup servers are disabled while updating packages.

See https://github.com/mxe/mxe/pull/1694#issuecomment-285324739

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ DEFAULT_MAX_JOBS   := 6
 SOURCEFORGE_MIRROR := downloads.sourceforge.net
 PKG_MIRROR         := http://s3.amazonaws.com/mxe-pkg
 PKG_CDN            := http://d1yihgixbnrglp.cloudfront.net
-GITLAB_BACKUP      := https://gitlab.com/starius/mxe-backup2/raw/master/
+GITLAB_BACKUP      := http://gitlab.com/starius/mxe-backup2/raw/master/
 
 PWD        := $(shell pwd)
 SHELL      := bash
--- a/Makefile
+++ b/Makefile
@@ -35,8 +35,7 @@ PATCH      := $(shell gpatch --help >/dev/null 2>&1 && echo g)patch
 SED        := $(shell gsed --help >/dev/null 2>&1 && echo g)sed
 SORT       := $(shell gsort --help >/dev/null 2>&1 && echo g)sort
 DEFAULT_UA := $(shell wget --version | $(SED) -n 's,GNU \(Wget\) \([0-9.]*\).*,\1/\2,p')
-WGET        = wget --no-check-certificate \
-                   --user-agent='$(or $($(1)_UA),$(DEFAULT_UA))'
+WGET        = wget --user-agent='$(or $($(1)_UA),$(DEFAULT_UA))'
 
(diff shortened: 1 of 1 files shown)


tonytheodore:
> On the one hand, old systems may still rely on --no-check-certificate,
so it is not nice to completely disable it for regular downloads.

Do we know what these old systems are - would they really have trouble with certificates on prominent sites like AWS and GitLab? End users may not be able to upgrade beyond WinXP and developers may not have access to the latest hardware, but developers should be able to keep their own systems updated.

starius:
> Do we know what these old systems are - would they really have trouble with certificates on prominent sites like AWS and GitLab?

All my systems do not need `--no-check-certificate`. Maybe we should ask other users in the mailing list to see if this is a case for everybody.

> trouble with certificates on prominent sites like AWS and GitLab

When I checked last time, IE 6 could not even open GitHub which is also prominent :-)

`--no-check-certificate` was added in eda23ec881b784b77d402ac6576256b26bdbbb99. I can't find context of this change.

@vog 
Can you tell more about this change, please?

tonytheodore:
See:
http://lists.nongnu.org/archive/html/mingw-cross-env-list/2010-05/msg00015.html

It appears to have been instigated by a self-signed certificate, arguably we should fail in those cases, especially with [Lets Encrypt](https://letsencrypt.org/) being so easy.

starius:
We can disable

vog:
The rationale behind this chance was that it is more important to be able to download these files. Since we have our own checksums, it doesn't matter if the HTTPS certificate is valid or not. It doesn't even matter whether we use SSL at all. Back then, most downloads were plain HTTP anyway.

However, now that we have our own mirror, we can afford to be more picky with upstream. Moreover, there's still the initial download after a new version, from which our checksum is initially created.

So I agree with removing `--no-check-certificate`, and with using HTTP for our mirrors, as long as we use HTTPS with strong certificate check for initial downloads (`make update-...`) whenever possible.

tehnick:
> On the one hand, old systems may still rely on --no-check-certificate, so it is not nice to completely disable it for regular downloads.

New systems may rely on ```--no-check-certificate``` too in cases when web-sites are using non-trusted SSL certificates (for example, certificates from WoSign and StartCom) or even self-signed certificates. So it is good to have a reserve variant.

vog:
@tehnick While this is indeed possible, I believe this is just a temporary problem, if at all. HTTPS sites with broken certificates will want to replace them as quickly as possible. Moreover, given the success of Let's Encrypt we probably won't see anything else in the future, while "classic" CAs will only service niches such as webshops and websites of large companies.

PR #1704, opened 2017-03-09
libsvm 3.22

vszakats:
3.20 is no longer available on the website.


starius:
Tested on all 4 targets.
Thanks!

PR #1705, opened 2017-03-09
.travis.yml: upgrade to Trusty CI image

vszakats:
This upgrade results in significantly fewer download errors and fallbacks to backup server.
https://github.com/mxe/mxe/pull/1694#issuecomment-285391552

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: c
+dist: trusty
 sudo: false
 
 script:


starius:
This decrease seems to be occasional. I [rebuilt the pull request](https://travis-ci.org/mxe/mxe/builds/209481498) and the number of backup downloads was 9. The same number of backup downloads [happened](https://travis-ci.org/mxe/mxe/builds/210192686) in a commit in master. I think, the real reason of them is that SF sometimes returns 404.

vszakats:
You're right, sf.net had a some bad days this week where port 443 didn't respond on `downloads.sourceforge.net`. Let's close this.

Issue #1708, opened 2017-03-10
Bash script to install packages on macOS (and BSDs) securely


vszakats:
This is not an issue, just an announcement that I wrote a script that is able to install a list of packages, together with their dependencies from the debian APT repository. Necessary signature and checksums are all verified.

https://github.com/vszakats/harbour-core/blob/master/package/mpkg_win_dl_mxe.sh

Example:
```bash
$ mpkg_win_dl_mxe.sh mxe-{i686,x86-64}-w64-mingw32.static-gd
```

Caveats:
- It was created for [Travis CI cross-builds](https://github.com/vszakats/harbour-core/blob/c582897d340d2e22b34114ee3ab0b9d9a4756ae5/.travis.yml#L64-L73) for a specific project, but the logic is largely generic.
- For better or worse, packages are installed into the `~/mxe` directory. Symlinks are retargeted for this location.
- Temporary files are stored in above directory.

Requirements:
- `ar` (BSD-specific), `gpg`, `curl`, `openssl`, `awk`, `sed`, `tar`, `gzip`

starius:
Looks interesting! It makes sense to announce it on [the mailing list](https://lists.nongnu.org/mailman/listinfo/mingw-cross-env-list) (maybe after some improvements, see below). Normally announces go there not to the issues.

* It's a shame we don't sign tar.xz files in the repo! Is it the only reason you use Debian packages instead of [tar.xz archieves](http://pkg.mxe.cc/repos/tar/)? I opened a bug about it: https://github.com/mxe/mxe-apt/issues/5

* Can you add `set -xue` in the beginning of the script? Currently it doesn't stop in case of a failure in one of its commands.

* Can you add the script to https://github.com/mxe/mxe/tree/master/tools to keep it in sync with build-pkg? If something changes in build-pkg, it will be reflected in your script in the same commit.

vszakats:
- Yes, I started out by using `.tar.xz` files, but it felt risky to use them without any option to verify them. Even the download goes in cleartext. Digital signatures would have been quite nice indeed. Possibly also for the control files in `http://pkg.mxe.cc/repos/deb-control/*`, which I was using for pulling the list of dependencies. Anyhow the current `.deb` method seems to work nicely and AFAICT without fundamental issues.
- `set -xue`: Agreed on `-ue`. `-x` would make the output very chatty so I wouldn't add it as default (though it could be enabled based on some envvar f.e.)
[...]

PR #1711, opened 2017-03-12
add bash script to install packages on non-Linux systems

vszakats:
*BSD systems are currently supported and macOS actually tested.

Ref: https://github.com/mxe/mxe/issues/1708

### DIFF
--- a/tools/mxe-get
+++ b/tools/mxe-get
@@ -15,10 +15,11 @@ Environment:
 
 MXE_HOME          Configure directory where packages will be installed.
                   Default: ~/mxe
-MXE_DONT_INSTALL  Comma/space separated list of package names to skip
+MXE_SKIP_INSTALL  Comma/space separated list of package names to skip
                   installing. Useful to exclude unnecessary dependencies.
                   *-linux-gnu-*, mxe-requirements, mxe-source packages
                   are excluded automatically.
+MXE_INST_LINUX    Install 'linux-gnu' packages. Default: no
 
 Required: ar (BSD), gpg, curl, openssl, awk, sed, tar, gzip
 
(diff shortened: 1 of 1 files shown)


starius:
Please add the following comment to the file:

```
```

tools/mxe-get
```diff
@@ -0,0 +1,129 @@
...
+    echo "! Installing mxe package '$1'"
+    mxe_get_pkg "$1"
+    shift
+  done
+
+  echo "! Installed:${done//|/}"
+
+  echo '! Retargeting symlinks...'
+  find . -type l -name '*' | while IFS= read -r f; do
+    # TOFIX: readlink may need to be adapted for non-macOS systems
```
starius: s/TOFIX/FIXME/

starius:
I tested it under Linux. It failed with the following error:

```
$ ./mxe-get 
! Downloading and verifying mxe APT package list...
gpg: new configuration file `/home/test/.gnupg/gpg.conf' created
gpg: WARNING: options in `/home/test/.gnupg/gpg.conf' are not yet active during this run
gpg: requesting key DB45C6AB from hkps server keyserver.ubuntu.com
gpgkeys: HTTP fetch error 1: unsupported protocol
gpg: no valid OpenPGP data found
```

Fixed by replacing `hkps://keyserver.ubuntu.com` with `x-hkp://keys.gnupg.net`

starius:
I tried to install Lua using the script:
```
$ ./mxe-get mxe-i686-w64-mingw32.static-lua
```

It installed only packages `mxe-i686-w64-mingw32.static-lua` and `mxe-i686-w64-mingw32.static-mxe-conf`. Compiler and other common packages were not installed.

vszakats:
(I've been busy with some updates just while you reviewed)

- Replace full MIT license with the reference to LICENSE.md
- TOFIX &rarr; FIXME done.
- `linux-gnu` packages are excluded. These have no use on non-Linux (f.e. *BSD) systems.
- `mxe-requirements` and `mxe-source` are also excluded. I didn't need them and their naming would be a special case, so it needs to be implemented if important. Also because `mxe-get` is part of `mxe-source`, it needs to present already to run the script.
- As for the GPG error. I use secure protocol to get the keys and this requires either GPG 2.x or the `gnupg-curl` packages to be installed on Linux. It also works fine on macOS with Homebrew `gpg` package installed. `hkp` and `x-hkp` are both cleartext and as you may have noticed :) I avoid it everywhere.

starius:
> are both cleartext and as you may have noticed :) I avoid it everywhere

I noticed this :)

I encourage encrypting everything in general, but in this particular case it introduces maintenance cost, which I'm not a big fan of. To get privacy from my ISP, I'd rather route all the traffic through Tor instead of fixing things like `hkps` one by one. Nevertheless it is up to you to decide which scheme to use in this case.

I installed `gnupg-curl` but it still fails:

```
gpg: requesting key DB45C6AB from hkps server keyserver.ubuntu.com
gpgkeys: HTTP fetch error 60: server certificate verification failed. CAfile: none CRLfile: none
gpg: no valid OpenPGP data found.
```

vszakats:
Went with replacing `gpg` with a `curl | gpg` combo. Let's hope this works more universally.

vszakats:
Added to rewrite hardcoded paths in some filetypes. Also added an option to install Linux packages. And some minor touchups.

starius:
Looks good to me!

starius:
Thanks!

PR #1695, opened 2017-03-05
add missing gstreamer dependencies

pavelvat:
fix for https://github.com/mxe/mxe/issues/1624

https://github.com/mxe/mxe/pull/1684 was splitted on two independent PR due to unresolved problem with high CPU usage with gstreamer 1.10.4.

PR #1724, opened 2017-03-19
enable native gettext and libiconv on Jessie

starius:
Fix for https://github.com/mxe/mxe/pull/1718
See https://github.com/mxe/mxe/pull/1695#issuecomment-287574293


tonytheodore:
Thanks!

PR #1712, opened 2017-03-13
Standardise GitHub downloads and updates

tonytheodore:

PR #1725, opened 2017-03-19
LZ4

manner82:
Hi,

This would add lz4, a fast compression library. All 4 target works (x86_64-w64-mingw32.shared x86_64-w64-mingw32.static i686-w64-mingw32.shared i686-w64-mingw32.static), little patching was only needed to avoid installation of the tools (*.exe).

Robert

### DIFF
--- a/src/lz4-test.c
+++ b/src/lz4-test.c
@@ -1,0 +1,36 @@
+#include <lz4.h>
+
+#include <string.h>
+#include <stdio.h>
+
+int
+main(int argc, char *argv[])
+{
+  const char *data;
+  int data_len;
+  char compressed[100];
+  int compressed_size;
+  char decompressed[100];
+
+  (void)argc;
+  (void)argv;
+
+  data = "Some data to compress";
+  data_len = strlen(data);
+
+  /* compress */
+  compressed_size = LZ4_compress_default(data, compressed, data_len, 100);
+  if (compressed_size <= 0) {
+    printf("Error compressing the data\n");
+    return 1;
+  }
+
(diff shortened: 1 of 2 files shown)


starius:
After #1712, you can replace $(PKG)_FILE, $(PKG)_SUBDIR, $(PKG)_URL, and $(PKG)_UPDATE with: $(PKG)_GH_CONF

starius:
The patch can be avoided by adding the following options:
```
-DCMAKE_INSTALL_BINDIR=$(BUILD_DIR)/null
-DCMAKE_INSTALL_MANDIR=$(BUILD_DIR)/null
```

starius:
Build tested on all 4 targets.

manner82:
Thanks for the ideas and the quick response. (And sorry for my delay, been a bit busy.)

I really like the github integration approach, but I have a little issue yet with using it. The lz4 package has a "v" version prefix, so versions look like: "v1.7.5". But when you extract the tar.gz, it contains a directory "lz4-1.7.5" which makes the build fail, because it would search for the sources including the version prefix "lz4-v1.7.5".

So I need to figure this out first (~weekend, I am quite new to mxe). Do you have an idea?

manner82:
Please do not merge yet, not working. I have just updated the request to show the problem.

tonytheodore:
#1729 fixes the `v` prefix issue. Can you move GH_CONF below CHECKSUM? Otherwise, this looks good to merge.

manner82:
Done. (Thanks for the prefix fix, seems to work fine.)

tonytheodore:
Thanks!

PR #1726, opened 2017-03-19
mxe-get: relocate qt related file-types

vszakats:


starius:
The change looks good.

Can you also use `sed -i` to change files in-place, please?

vszakats:
It's not supported in BSD sed.

tonytheodore:
(Free)BSD users should have [`coreutils/gmake/gsed` installed](http://mxe.cc/#requirements-freebsd) so you can use something like [`tools/compat-init.sh`](https://github.com/mxe/mxe/blob/master/tools/compat-init.sh) to get consistent behaviour.

vszakats:
I know, but in the current script that sort of workaround would have been heavier and would break without these optional tools. Anyhow, please feel free to customize the script for the MXE environment as it feels best. I'm closing this one.

vszakats:
(Forgot another important aspect: This script was specifically created with non-Linux, *BSD systems in mind and core functionality is implemented using `ar`, which is a BSD-specific tool. As previously mentioned, it was only tested on macOS. While it could also be useful in non-Debian Linuxes, I'm currently not using this in such environment to make decisions about it.)

PR #1729, opened 2017-03-20
github.mk: fix tag prefix with single `v` and pkg name <> project name

### LINKED ISSUE #1728: Github subdir incorrect on certain tags
from https://github.com/mxe/mxe/pull/1727#issuecomment-287655994, tags with a single `v` prefix seem to strip this when downloading. `qca` #1709 will have the same issue.

tonytheodore:
fixes #1728

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ MXE_LIB_TYPES      := static shared
 MXE_TARGET_LIST    := $(strip $(foreach TRIPLET,$(MXE_TRIPLETS),\
                           $(addprefix $(TRIPLET).,$(MXE_LIB_TYPES))))
 MXE_TARGETS        := i686-w64-mingw32.static
+.DEFAULT_GOAL      := all-filtered
 
 DEFAULT_MAX_JOBS   := 6
 SOURCEFORGE_MIRROR := downloads.sourceforge.net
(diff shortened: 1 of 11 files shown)


pavelvat:
Maybe this variant:
```
cd '$(BUILD_DIR)' && $(TARGET)-cmake '$(SOURCE_DIR)' \
```
more unified than:
```
cd '$(BUILD_DIR)' && $(TARGET)-cmake \
... \
... \
... \
'$(SOURCE_DIR)'
```

tonytheodore:
How do the [cmake rule examples](https://github.com/mxe/mxe/wiki/Customisable-Packages#rule-examples) look? We could add `$(MXE_CMAKE_OPTS)` and move options out of the toolchain/wrapper that [complicate downstream use](https://github.com/mxe/mxe/pull/1621#discussion_r106937505).

tonytheodore:
I'll take those `BUILD_DIR` changes out of this PR.

tonytheodore:
> I'll take those BUILD_DIR changes out of this PR.

Actually, I'll leave them in - there's no functional change and they're already tested.

This PR seems covers some more cases with GH_CONF, the odd single `v` prefix and projects with different package names. There's another api to investigate - `releases`, but it shouldn't hold this up.

PR #1732, opened 2017-03-22
Fix broken shared build of GTK3

ilhooq:
Static and shared build works with this patch

### DIFF
--- a/src/gtk3-1-fixes.patch
+++ b/src/gtk3-1-fixes.patch
@@ -7,33 +7,6 @@ From: Boris Nagaev <bnagaev@gmail.com>
 Date: Sun, 29 Jan 2017 21:47:39 +0100
 Subject: [PATCH] These were adapted from the gtk2 patchset.
 
-
-diff --git a/configure b/configure
-index 1111111..2222222 100755
---- a/configure
-+++ b/configure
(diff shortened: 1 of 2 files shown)


tonytheodore:
Are you sure the static build is actually static after this?

ilhooq:
The static build produce these files : 
usr/i686-w64-mingw32.static/lib/libgtk-3.a (35,6 MB)
usr/i686-w64-mingw32.static/bin/test-gtk3.exe (15,7MB)

And the test-gtk3.exe works fine as standalone binary file on windows (tested on windows 8)

tonytheodore:
I see, it simply causes `libtool` to complain and create static libs:

```
*** Warning: This system cannot link to static lib archive /Users/tonyt/dev/mxe/usr/i686-w64-mingw32.static/lib/../lib/libpangocairo-1.0.la.
*** I have the capability to make that library automatically link in when
*** you link to this library.  But I can only do this if you have a
*** shared version of the library, which you do not appear to have.
[...]

*** Since this library must not contain undefined symbols,
*** because either the platform does not support them or
*** it was explicitly requested with -no-undefined,
*** libtool will only create a static version of it.
```

The test also works for gtkmm3 so I've enabled that also.

Thanks!

PR #1735, opened 2017-03-26
Bump SDL2_image, SDL2_mixer, SDL2_tff and SDL2_gfx

Beuc:
Grab several bugfixes in SDL2 extensions :)


tonytheodore:
Tested on all four target - thanks!

P.S. in the future, could you say "update" or do separate commits with "foo: update old --> new"?

Beuc:
Sure.

PR #1737, opened 2017-03-31
binutils: Ensure that the timestamp in PE/COFF headers is always init…

Beuc:
Hi,

When using --no-insert-timestamp, instead of inserting a 0 timestamp, ld 2.25.1 does not initialized memory resulting in a random date timestamp. 

This pull requests applies the 2-lines fix from upstream, resulting in stable/reproducible output.

I noticed this issue when working on a reproducible build of my FreeDink game, you can see details at:
https://blog.beuc.net/posts/Practical_basics_of_reproducible_builds_2/
https://blog.beuc.net/posts/Practical_basics_of_reproducible_builds/

(For your info, Debian has a more extensive, upstream-pending patch related to the SOURCE_DATE_EPOCH specification, see specify-timestamp.patch in http://http.debian.net/debian/pool/main/b/binutils-mingw-w64/binutils-mingw-w64_7.4.tar.xz)

### DIFF
--- a/src/binutils-1-fixes.patch
+++ b/src/binutils-1-fixes.patch
@@ -1,0 +1,29 @@
+This file is part of MXE. See LICENSE.md for licensing information.
+
+Contains ad hoc patches for cross building.
+
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Sylvain Beucler <beuc@beuc.net>
+Date: Fri, 31 Mar 2017 11:59:35 +0200
+Subject: [PATCH] Ensure that the timestamp in PE/COFF headers is always
+ initialised.
+
+Source: http://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=commit;f=bfd/peXXigen.c;h=1c5f704fc035bc705dee887418f42cb8bca24b5d
+
+PR ld/20634
+* peXXigen.c (_bfd_XXi_only_swap_filehdr_out): Put 0 in the
+timestamp field if real time values are not being stored.
+
+diff --git a/bfd/peXXigen.c b/bfd/peXXigen.c
+index 1111111..2222222 100644
+--- a/bfd/peXXigen.c
++++ b/bfd/peXXigen.c
+@@ -875,6 +875,8 @@ _bfd_XXi_only_swap_filehdr_out (bfd * abfd, void * in, void * out)
+   /* Only use a real timestamp if the option was chosen.  */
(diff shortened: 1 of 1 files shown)


starius:
Thank you! I'm testing your patch.
I reformatted the patch file in the format produced by tools/patch-tool-mxe:
https://gist.github.com/265da3742f46d0bf055d00d96f269b4b
Can you replace the file in your pull request with it (and put it in src/binutils-1-fixes.patch), please?

Beuc:
Like this?

By the way, are there plans to upgrade binutils?
Stephen Kitt pointed out that a similar issue affects DLL files, his patch is here:
https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=commitdiff;h=030f4c7f08246bdc79d74c4c297729003773b67c

starius:
> Like this?

Exactly. Thanks!

> By the way, are there plans to upgrade binutils?

File src/binutils.mk has the following comment:

```
# see https://lists.nongnu.org/archive/html/mingw-cross-env-list/2016-01/msg00013.html
# 2.26 causes incorrect dlls to be built with sjlj exceptions
```

Are there any updates on that?

> Stephen Kitt pointed out that a similar issue affects DLL files, his patch is here

Can we add his patch as well?

Beuc:
> File src/binutils.mk has the following comment:
> 
> # see https://lists.nongnu.org/archive/html/mingw-cross-env-list/2016-01/msg00013.html
> # 2.26 causes incorrect dlls to be built with sjlj exceptions
> 
> Are there any updates on that?

I have no clue about this bug :/
However 2.28 is out now, and also 2.27 will be in the next Debian:
https://packages.debian.org/stretch/binutils-mingw-w64
I don't see any patch related to that issue in the package.
That doesn't mean anything per se but this is encouraging.

>     Stephen Kitt pointed out that a similar issue affects DLL files, his patch is here
> 
> Can we add his patch as well?

Given Stephen's credentials I would trust his patch, however I don't work much with DLLs these days hence I haven't tested the issue extensively as I did for this pull request.

Beuc:
Hi, any news on this? :)

tonytheodore:
@Beuc, interesting posts on reproducible builds - thanks.

I've done a new PR (#1822) to update binutils and add the `SOURCE_DATE_EPOCH` patch. The main holdup to updating was the 2.26 regression that's now fixed.

tonytheodore:
Replaced with #1822

PR #1739, opened 2017-04-01
update wt to 3.3.7

starius:
Tested on all 4 targets.


tonytheodore:
Confirmed on all 4 targets, thanks!

PR #1743, opened 2017-04-03
build-matrix: fix detection of native virtual packages

tonytheodore:


starius:
So it checks that build rules are set for a package in addition to existing checks.

I tested the patch locally and found that column "native" changed for the following packages from "supported" to "unsupported":

  * googlemock
  * googletest
  * lzma

Looks good to me!

starius:
Thanks!

Issue #1759, opened 2017-05-03
SFML 2.4.2 update


oomek:
Any chance to update sfml to 2.4.2 ?

starius:
Please take a look https://github.com/mxe/mxe/pull/1760

oomek:
Awesome, thanks. Will the precompiled package be updated soon?

oomek:
I've updated mxe with apt-get from pkg.mxe.cc but config.cppin the SFML filder still showing 2.3.2

starius:
The pull request https://github.com/mxe/mxe/pull/1760 hasn't been even merged yet.
Then the rebuild of binary distribution will take at least a day.

starius:
Travis fails to download new sfml for some reason:
https://travis-ci.org/mxe/mxe/builds/228516653#L735

oomek:
I see, the green checkmark mislead me. Thanks for explaining.

oomek:
Did you make any proggress with this issue by any chance?

oomek:
I desperately need this update. Have you managed to figure out why it's not compiling yet?

starius:
Sorry for the delay!

The issue was not with sfml but with the fact its site uses new CA which is not recognized by old version of wget used in the test. I updated the test environment in https://github.com/mxe/mxe/pull/1760/commits/3ad6743c7bfba5c16bb3fa8cf9c3e1eb08500d97 so it works now.

Regarding pkg.mxe.cc - I'll start the build soon.

oomek:
Thanks a lot!

oomek:
I don't want to hustle you or anything, but can you please be more specific what soon means in this case :)

starius:
The packages have been built. They are being packed to .deb and to .tar now.

starius:
The packages are available for testing. Can you test your workflow on the new version of packages, please? To try it, use official instructions (http://pkg.mxe.cc), replacing http://pkg.mxe.cc/repos/apt/debian with http://pkg.mxe.cc/releases/2017-05-19/apt/debian

I'll also do some tests on my side. If everything is OK, I'll announce the new version and switch default path http://pkg.mxe.cc/repos/apt/debian to the new version.

oomek:
Unfortunately on the new version compiling fails during linking.
Here is the log file:
http://paste.ubuntu.com/24614093/

oomek:
I see now all sfml package config files has been consolidated to one sfml.pc
[...]

PR #1760, opened 2017-05-03
sfml: update to 2.4.2

starius:
fix https://github.com/mxe/mxe/issues/1759


starius:
Travis fails because of Self-signed certificate:
https://travis-ci.org/mxe/mxe/builds/228715423#L787

PR #1765, opened 2017-05-08
let gnutls use libidn2

mabrand:
Gnutls is supposed to be able to detect and use libidn2, but it
tries to do this without pkg-config, in a way that doesn't
consider transitive dependencies. Using pkg-config with libidn2.pc
seems to work okay with MXE's version of libidn2.

### DIFF
--- a/src/gnutls.mk
+++ b/src/gnutls.mk
@@ -3,8 +3,8 @@
 PKG             := gnutls
 $(PKG)_WEBSITE  := https://www.gnu.org/software/gnutls/
 $(PKG)_DESCR    := GnuTLS
-$(PKG)_VERSION  := 3.5.11
-$(PKG)_CHECKSUM := 51765cc5579e250da77fbd7871507c517d01b15353cc40af7b67e9ec7b6fe28f
+$(PKG)_VERSION  := 3.5.12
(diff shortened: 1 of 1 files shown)


starius:
I removed the default PR description from the comment.

starius:
Can you use `./tools/patch-tool-mxe` to re-unify patch, please?
```
$ ./tools/patch-tool-mxe init gnutls
$ ./tools/patch-tool-mxe import gnutls
$ ./tools/patch-tool-mxe export gnutls
```

mabrand:
Hi Starius, Thanks for the review. Just curious, what are the benefits of "unifying" patch files with patch-tool-mxe? The output of "git format-patch -p" seems pretty unified to me. Even though I wrote the original version of patch-tool-mxe, I stopped using it after finding it pretty easy to work with git directly. The only thing that annoys me is that "git am" chokes on the extra MXE headers at the top of the patch files and the mangled email addresses.

starius:
Unifying makes patches reproducible (i.e. import - export result in the same file). It makes diff of patch less noisy, because the diff shows the things that really changed not just changed hashes of commits and similar stuff.

starius:
I'm building this PR with `build-pkg`. Everything quiet so far.

PR #1775, opened 2017-05-25
Updated cgal to 4.9.1.

rmarsa:


src/cgal.mk
```diff
@@ -1,14 +1,14 @@
 # This file is part of MXE. See LICENSE.md for licensing information.
 
 PKG             := cgal
-$(PKG)_WEBSITE  := https://www.cgal.org/
+$(PKG)_WEBSITE  := http://www.cgal.org/
```
starius: Why https URL was replaced with http?

starius:
Tested on all 4 targets.

src/cgal.mk
rmarsa: No reason.  I fixed and pushed the change.

starius:
Thanks!

PR #1776, opened 2017-05-26
wxwidgets: enable DLLs (close #1767)

### LINKED ISSUE #1767: Enable wxwidgets for *.shared targets
Can you please enable wxwidgets for the *.shared targets?
It's building without problems for me on `x86_64-w64-mingw32.shared` and `i686-w64-mingw32.shared`. The `test-wxwidgets.exe` are also working (tested only with wine). Only problem is that the build system installs the .dll files in `lib` instead of `bin`.

darealshinji:
Enabling i686-w64-mingw32.shared and x86_64-w64-mingw32.shared builds.


darealshinji:
Added a header to the patch so it's in the new preferred format.

tonytheodore:
Looks good with some stylistic fixups in https://github.com/darealshinji/mxe/pull/4. If you merge that and rebase against master to remove the merge commits, it's ready to go.

darealshinji:
Rebased it.

tonytheodore:
Thanks!

PR #1780, opened 2017-06-04
add new package: quazip

zgyarmati:
Continuing from #1176, cleaned up and added the requested fixes. If you spot anything further issue, please let me know. 
2 out of the 3 patches are sent to upstream, the one which adds explicit link to -lz is MXE specific, as the user free to link to other zip implementation as well (see https://sourceforge.net/p/quazip/patches/27/#6698 )


tonytheodore:
Thanks! The new PR is much easier to review. There are various fixups that were easier to do than write a review:

  - use standard header
  - add website
  - add explicit zlib dependency
  - build out of source
  - fix indentation
  - fix qmake handling of import lib
  - consolidate patches

See: https://github.com/mxe/mxe/commit/a4ebc1d7537626e8469631c90b8037a7611eb44f

zgyarmati:
Great, thanks! In the meantime the pkg-config file generating patch got accepted in the mainline quazip, so i'll look into it to sync them up, and at the next quazip release i'll also do the needed rework here.

Issue #938, opened 2015-10-15
qtbase log is broken


starius:
It can be viewed with `less` but can't be opened in a text editor. A text editor shows only beginning of the file, before first zero byte.

With `less`:
```
DEFAULT_INCDIRS="/home/mxe/mxe-toolchain-reorg/usr/lib/gcc/i686-w64-mingw32.static/5.2.0/include/c++
/home/mxe/mxe-toolchain-reorg/usr/lib/gcc/i686-w64-mingw32.static/5.2.0/include/c++/i686-w64-mingw32.static
/home/mxe/mxe-toolchain-reorg/usr/lib/gcc/i686-w64-mingw32.static/5.2.0/include/c++/backward
/home/mxe/mxe-toolchain-reorg/usr/lib/gcc/i686-w64-mingw32.static/5.2.0/include
/home/mxe/mxe-toolchain-reorg/usr/lib/gcc/i686-w64-mingw32.static/5.2.0/include-fixed
/home/mxe/mxe-toolchain-reorg/usr/i686-w64-mingw32.static/include
"
DEFAULT_LIBDIRS="/home/mxe/mxe-toolchain-reorg/usr/lib/gcc/i686-w64-mingw32.static/5.2.0
/home/mxe/mxe-toolchain-reorg/usr/i686-w64-mingw32.static/lib
"
^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@
^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@
^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@
^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@
^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@
^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@
^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@
[...]

PR #1785, opened 2017-06-14
gcc 7 build fix: strip  standard -I path from pkg-config output

mabrand:
MXE's standard include directory is '$(PREFIX)/$(TARGET)/include'.

Many of the .pc files installed by packages include the standard
include directory in Cflags. This can cause gcc to search paths in
the wrong order. This typically interferes with the intended effect of
the directive #include_next in gcc headers, with "not found" errors for
headers like stdio.h. Using gcc 7, such errors occur when building
qtbase in MXE.

    https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70129
    https://bugreports.qt.io/browse/QTBUG-53375

It is not obvious how to get pkg-config to strip the standard
include path when cross-compiling. According to the pkg-config manpage,
pkg-config normally strips "-I/usr/include" from Cflags.
PKG_CONFIG_SYSROOT_DIR could be set so that pkg-config would prefix
"/usr/include", but this doesn't help cross-compiling in MXE where the
standard include directory in the .pc files is the fully qualified path.
Addionally, MXE's standard include directory does not even end in
"/usr/include".

An alternative would be to patch all .pc files to remove the standard
include directory, but this requires considering each case individually
and creates a maintenance burden.

This solution offered by this patch modifies the wrapper script around
pkg-config, using sed to strip out '-I$(PREFIX)/$(TARGET)/include'.

Notes:
  - Variants of the path with extra preceding / are also handled.

  - Care is taken to get the exit code from pkg-config (not sed), so
    that "pkg-config --exists" can be used.

  - The exit code handling would be much harder without bash.

  - This change assumes the previously used "exec" in the wrappper
    script is not necessary, or at least no longer necessary.

This replaces 43214bf7e886bd310965f854dd3a37b64c685bfa, which fixed
the problem specifically for sqlite.

### DIFF
--- a/src/pkgconf.mk
+++ b/src/pkgconf.mk
@@ -20,8 +20,8 @@ endef
 
 define $(PKG)_BUILD
     # create pkg-config script
-    (echo '#!/bin/sh'; \
-     echo 'PKG_CONFIG_PATH="$(PREFIX)/$(TARGET)/qt5/lib/pkgconfig":"$$PKG_CONFIG_PATH_$(subst .,_,$(subst -,_,$(TARGET)))" PKG_CONFIG_LIBDIR='\''$(PREFIX)/$(TARGET)/lib/pkgconfig'\'' exec '$(PREFIX)/$(BUILD)/bin/pkgconf' $(if $(BUILD_STATIC),--static) "$$@"') \
+    (echo '#!/usr/bin/env bash'; \
+     echo 'PKG_CONFIG_PATH="$(PREFIX)/$(TARGET)/qt5/lib/pkgconfig":"$$PKG_CONFIG_PATH_$(subst .,_,$(subst -,_,$(TARGET)))" PKG_CONFIG_LIBDIR='\''$(PREFIX)/$(TARGET)/lib/pkgconfig'\'' '$(PREFIX)/$(BUILD)/bin/pkgconf' $(if $(BUILD_STATIC),--static) "$$@"' ' | ' '$(SED)' "'s,-I/*$(PREFIX)/$(TARGET)/include\($$\| \),,g;'" '; ' '( exit $${PIPESTATUS[0]} )')  \
              > '$(PREFIX)/bin/$(TARGET)-pkg-config'
     chmod 0755 '$(PREFIX)/bin/$(TARGET)-pkg-config'
 
--- a/src/sqlite.mk
+++ b/src/sqlite.mk
@@ -18,9 +18,6 @@ define $(PKG)_UPDATE
 endef
 
 define $(PKG)_BUILD
-    # The system include path doesn't belong in CFLAGS
-    # Causes problems like https://gitlab.kitware.com/cmake/cmake/issues/16919
-    $(SED) -i 's/^Cflags/#Cflags/;' '$(1)/sqlite3.pc.in'
     cd '$(1)' && ./configure \
         $(MXE_CONFIGURE_OPTS) \
         --disable-readline \


tonytheodore:
There's a compile option, `--with-system-includedir`, to specify the path to strip but that wouldn't help unless we want to compile `pkgconf` multiple times. There's also gcc spec files but they need a lot of investigation and have unintended consequences.

This approach seems good to me, I'll do some testing to see if the old `bash` on OSX (3.2 series) has any issues.

src/pkgconf.mk
```diff
@@ -20,8 +20,8 @@ endef
 
 define $(PKG)_BUILD
     # create pkg-config script
-    (echo '#!/bin/sh'; \
-     echo 'PKG_CONFIG_PATH="$(PREFIX)/$(TARGET)/qt5/lib/pkgconfig":"$$PKG_CONFIG_PATH_$(subst .,_,$(subst -,_,$(TARGET)))" PKG_CONFIG_LIBDIR='\''$(PREFIX)/$(TARGET)/lib/pkgconfig'\'' exec '$(PREFIX)/$(BUILD)/bin/pkgconf' $(if $(BUILD_STATIC),--static) "$$@"') \
+    (echo '#!/bin/bash'; \
```
tonytheodore: #!/usr/bin/env bash

mabrand:
If you think it makes more sense compile pkconfig for each target with the appropriate --with-system-includedir option and that works, I have no objection. It looks like we already build it for each target.

src/pkgconf.mk
mabrand: fixed

tonytheodore:
> I wonder what pkgconfig would do with the extra leading / that some of the .pc files have

Do you have an example? I can't find any with a simple grep.

> Do you know why that "exec" was there?

I'm guessing `exec` is a defensive call to ensure custom `PKG_CONFIG_PATH_i686_w64_mingw32_static` env vars are inherited and possibly cover [subtle differences in redirect behaviour](https://github.com/mxe/mxe/issues/938#issuecomment-149774654).

tonytheodore:
There's a [new (undocumented) env var in pkgconf](https://github.com/pkgconf/pkgconf/issues/86) that specifies the system dir at runtime `PKG_CONFIG_SYSTEM_INCLUDE_PATH` - this sounds like exactly what we need:
>Instead of hard-coding /usr/include, we now use the environment variable
PKG_CONFIG_SYSTEM_INCLUDE_PATH, defaulting to the argument of
./configure --with-system-include-path, which in turn defaults to
/usr/include.

If the [new `--pure` flag](https://github.com/pkgconf/pkgconf/commit/ec10dd8edbbf172cdf01000b5236f7acb0325cfa) works as expected, we can simply upgrade to the the latest `pkgconf`and stop using the pinned version.

mabrand:
That sounds good. I'll try it later.

>>I wonder what pkgconfig would do with the extra leading / that some of the .pc files have

>Do you have an example? I can't find any with a simple grep.

They are not hard coded in the .pc files, and at the moment I can't seem to reproduce this with my own pkg-config invocations. But you see the evidence of this in Qt .pri files containing the values that find their way into gcc invocations:

    grep 'QMAKE_INCDIR.* //'  `find usr/x86_64-w64-mingw32.static/qt5 -name '*.pri'`

I thought adding the '/*' to my filtering regex for sed was what solved this problem for me. Maybe there is something special about the way qmake invokes pkg-config. And I just noticed a longstanding error in src/qtbase.mk:

    PKG_CONFIG_LIBDIR="$(PREFIX)/$(TARGET)/lib/pc" 

That should end in "pkgconfig", not "pc".

mabrand:
The extra / comes from PKG_CONFIG_SYSROOT_DIR in qtbase.mk. It's there because the qtbase configure script wants to see PKG_CONFIG_SYSROOT_DIR.

mabrand:
The build of the new version of pkgconf wants to invoke a prefixed strip command, which doesn't exist for the host system. What do you think we should do about this?

tonytheodore:
> new version of pkgconf wants to invoke a prefixed strip command

I can't see that here, however the new version of pkgconf still doesn't have the desired (original) behaviour and we can't update to it yet. I've enabled the env var option in #1787 but there's not much to recommend that approach over `sed`.

mabrand:
>>new version of pkgconf wants to invoke a prefixed strip command
>I can't see that here,

What I saw was that the pkconf build appeared to succeed, but there was no usr/bin/x86_64-unknown-linux-gnu-pkg-config installed. In the log I saw the error about missing strip.

    x86_64-unknown-linux-gnu-strip '/home/brand/projects/mxe/mxe-wip/usr/x86_64-unknown-linux-gnu/bin/test-pkgconf.exe'
    bash: x86_64-unknown-linux-gnu-strip: command not found
    Makefile:654: recipe for target 'build-only-pkgconf_x86_64-unknown-linux-gnu' failed
    make[1]: [build-only-pkgconf_x86_64-unknown-linux-gnu] Error 127 (ignored)

tonytheodore:
> but there was no usr/bin/x86_64-unknown-linux-gnu-pkg-config installed

it should be in `usr/x86_64-unknown-linux-gnu/bin/pkgconf`

> In the log I saw the error about missing strip

That's just a generic strip in the Makefile (#1492) for any test programs.

mabrand:
Superseded by  #1788

vog:
I'm glad you finally found a simple fix, so there's nothing left to do for me - except for explaining the "exec thing".

That `exec` was originally introduced by myself.

The `pkgconf.mk` (formerly `pkg-config.mk`) creates a shell script that acts as a wrapper for pkg-config / pkgconf. I designed this wrapper to be as thin and as seamless as possible. Whenever a shell script is designed to wrap another program, calling that program with `exec` is generally considered good practice:

```
#!/bin/sh

# ... do stuff ...

exec wrapped_tool "$@"
```

Using `exec` means that the wrapper process is replaced by the `wrapped_tool` process. Otherwise, `wrapped_tool` is a sub process of the shell. Using `exec` has several benefits:

* The exit code of the wrapper will automatically become the exit code of `wrapped_tool`.  Of course, you can achieve the same effect if you use `set -e` in the shell script.
* While running, only the `wrapped_tool` is in memory, the shell is wiped out on `exec` and no longer uses any resources (memory, process scheduling, ...)
* The process name and path is `wrapped_tool`, not the shell script.
* If the wrapper is supervised and the caller decides to kill it, it will kill directly the `wrapped_tool`. Otherwise, it would only kill the shell and `wrapped_tool` might continue to run in the background.

Hope this helps to clarify your question.

PR #1787, opened 2017-06-15
pkgconf sys include

tonytheodore:
From #1785, enable stripping of system include path via env vars. This is probably less useful than the `sed` approach ;)

### DIFF
--- a/src/pkgconf-1-fixes.patch
+++ b/src/pkgconf-1-fixes.patch
@@ -1,0 +1,50 @@
+This file is part of MXE. See LICENSE.md for licensing information.
+
+Contains ad hoc patches for cross building.
+
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Tony Theodore <tonyt@logyst.com>
+Date: Thu, 15 Jun 2017 21:05:31 +1000
+Subject: [PATCH] main: support undocumented PKG_CONFIG_SYSTEM_INCLUDE_PATH and
+ PKG_CONFIG_SYSTEM_LIBRARY_PATH environment variables.
+
+Backported from:
+https://github.com/pkgconf/pkgconf/commit/7e6fa325eb668c3462981a16fb4c36270832e00f
+
+See:
+https://github.com/mxe/mxe/pull/1785
+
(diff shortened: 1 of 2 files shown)


mabrand:
If we use PKG_CONFIG_SYSTEM_INCLUDE_PATH, maybe we might as well also use PKG_CONFIG_SYSTEM_LIBDIR_PATH, even though I'm not aware of the unnecessary mentiones of the standard lib search directory causing chaos.

tonytheodore:
Library search paths aren't order-dependent for the linker and there's no equivalent of `#include_next` for libs, so it's unlikely to cause any issues.

mabrand:
>Library search paths aren't order-dependent for the linker and there's no equivalent of #include_next for libs, so it's unlikely to cause any issues.

Just to play devil's advocate:

In practice that's true at least, since there aren't any cases of different libraries of the same name in different search directories. If there were, I don't actually know which one gcc would pick, but if the behavior is deterministic, it's  hard to imagine that it doesn't have something to do with the order of search paths.

There might other good reasons for setting PKG_CONFIG_SYSTEM_LIBDIR_PATH, even if the case is not as strong as for PKG_CONFIG_SYSTEM_INCLUDE_PATH.

One is that pkg-config would normally strip out the system libdir, so this would help pkg-config behave according to normal expectations.

The second is aesthetic: to reduce the jumble of lib paths  in the linker lines.

tonytheodore:
>since there aren't any cases of different libraries of the same name in different search directories
>[...]
> it's hard to imagine that it doesn't have something to do with the order of search paths.

Apropos, I'm seeing an error in `gd` where an erroneous native lib path causes problems. I've never looked closely at these, but changing the search order does reduce the number of undefined references - very useful to consider when thinking about alternate implementations like lua/luabind or jpeg/jpeg-turbo.

> The second is aesthetic: to reduce the jumble of lib paths in the linker lines.

We use a pinned version of `pkgconf` for (almost) entirely aesthetic reasons! The version we use does a "minimal-ordered-dependency-set", basically reducing the jumble of libs. Later versions drop the "minimal" - this PR is somewhat an argument in favour of aesthetics.

PR #1788, opened 2017-06-16
Clean cflags v3

mabrand:
Better approach to cleaning up cflags proposed by Tony.

### DIFF
--- a/src/pkgconf-1-fixes.patch
+++ b/src/pkgconf-1-fixes.patch
@@ -1,0 +1,50 @@
+This file is part of MXE. See LICENSE.md for licensing information.
+
+Contains ad hoc patches for cross building.
+
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Tony Theodore <tonyt@logyst.com>
+Date: Thu, 15 Jun 2017 21:05:31 +1000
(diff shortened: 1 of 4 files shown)


mabrand:
Qtbase builds successfully with these changes.

tonytheodore:
All packages build with the exception of `gd`, which is somehow picking up a native lib path (most likely from `freetype` https://github.com/mxe/mxe/issues/1789#issuecomment-309369170).

Should `PKG_CONFIG_SYSROOT_DIR=/` be removed from `qtbase` and/or should `PKG_CONFIG_SYSROOT_DIR` be set correctly in the wrapper?

mabrand:
We cannot easily remove PKG_CONFIG_SYSROOT_DIR=/ from qtbase.mk, because the Qt build system assumes that PKG_CONFIG_SYSROOT_DIR must be set for cross-compiling. However, when we run pkg-config, we don't want PKG_CONFIG_SYSROOT_DIR to be set because pkg-config prepends it to the SYSTEM_INCLUDE_PATH and SYSTEM_LIBRARY_PATH. The extra leading '/' in turn prevents these paths from being stripped from --cflags and -libs output.

tonytheodore:
Okay, this looks good to me then - thanks!

PR #1793, opened 2017-06-21
Minor updates

tehnick:


src/ffmpeg.mk
```diff
@@ -4,11 +4,11 @@ PKG             := ffmpeg
...
 $(PKG)_IGNORE   :=
 $(PKG)_VERSION  := 3.3.2
-$(PKG)_CHECKSUM := 216900d5c0af9017bb8f76b0ad23f0ac53cf7fc618cf04b40d989bd99b088e6a
+$(PKG)_CHECKSUM := 1998de1ab32616cbf2ff86efc3f1f26e76805ec5dc51e24c041c79edd8262785
 $(PKG)_SUBDIR   := $(PKG)-$($(PKG)_VERSION)
-$(PKG)_FILE     := $(PKG)-$($(PKG)_VERSION).tar.bz2
-$(PKG)_URL      := https://ffmpeg.org/releases/$($(PKG)_FILE)
-$(PKG)_URL_2    := https://launchpad.net/ffmpeg/$(call SHORT_PKG_VERSION,$(PKG))/$($(PKG)_VERSION)/+download/$($(PKG)_FILE)
+$(PKG)_FILE     := $(PKG)-$($(PKG)_VERSION).tar.xz
+$(PKG)_URL      := http://ffmpeg.org/releases/$($(PKG)_FILE)
```
mabrand: Why http instead of https?

src/proj.mk
```diff
@@ -1,7 +1,7 @@
 # This file is part of MXE. See LICENSE.md for licensing information.
 
 PKG             := proj
-$(PKG)_WEBSITE  := https://trac.osgeo.org/proj/
+$(PKG)_WEBSITE  := http://proj4.org/
```
mabrand: https?

src/ffmpeg.mk
tehnick: Could you read a commit message?

src/proj.mk
tehnick: Invalid SSL certificate from DigiCert Inc.

tehnick:
> I think the URLs were switched to https on purpose not too long ago. Is there a good reason to change these?

Yes. HTTPS access should be used only for web-sites with valid SSL certificates received from trusted certification authorities. Also self-signed certificates should not be accepted by default.

tehnick:
Nowadays there is Let's Encrypt certification authority, so everyone interested may get valid SSL certificate absolutely free.

mabrand:
What does Tony think about using plain http for these cases?
The other commits look fine to me. If they were in a separate branch I would have merged them already.

tonytheodore:
For *_URL fields, I think the discussions in #1694 and #1702 are a solid policy - use https and certificate checking for initial seeding of files, allow fallback for downloads from our mirrors.

For *_WEBSITE fields, I'm a little ambivalent. I think public sites should be http, but use Let's Encrypt for every server I set up and force redirect to https (and don't really have many public sites). At least, public sites should not "need" to be concerned about privacy but there are so many bad actors that I don't think we'll go back to a more carefree time.

The `https://trac.osgeo.org/proj/` url redirects to `http://proj4.org` (in Chrome and Safari on macOS) but for MXE we can simply point to the project page `https://github.com/OSGeo/proj.4/`. Our users can find their way to the public site from any given code-hosting project page and are probably more likely to want the project page than the website.

@starius and @vog have a much better understanding of these issues than me. The main problem for me is conflating personal practices with what the project should do (and bike-shedding, of course!)

tehnick:
I have updated PR. Any additional comments?

tonytheodore:
@tehnick *URL and *UPDATE should be a primary source site with valid and well-known certs (if possible - ffmpeg has these).

*URL_2 can be removed since we have our mirrors.

*WEBSITE is orthogonal and should be moved to a new issue.

tehnick:
@tonytheodore Official [ffmpeg web-site](https://ffmpeg.org/releases/) with released tarballs uses SSL certificate from StartCom Ltd. And this is the root of problem.

tehnick:
https://blog.mozilla.org/security/2016/10/24/distrusting-new-wosign-and-startcom-certificates/

tehnick:
Of course we may use *.tar.gz tarballs from their mirror at GitHub:
https://github.com/FFmpeg/FFmpeg
But this is not a "primary source site" too.

tehnick:
Any practical recommends?

tonytheodore:
> Nowadays there is Let's Encrypt certification authority, so everyone interested may get valid SSL certificate absolutely free...

>Official ffmpeg web-site with released tarballs uses SSL certificate from StartCom Ltd. And this is the root of problem.

Not a practical recommendation but I'd be open to moving ffmpeg to a plugin or removing it. This is starting to look like an old style licence issue;)

tehnick:
@starius Comment this issue please.

tonytheodore:
> Of course we may use *.tar.gz tarballs from their mirror at GitHub

This sounds reasonable to me, git's content checksum and wire protocols should prevent the types of attacks we're trying to mitigate with using https for the initial seeding of tarballs. Again, @vog and @starius have more experience.

@tehnick are you seeing actual errors for ffmpeg? I've tried installing Firefox and purging all certificate caches etc. and can't seem to trigger any issues (on macOS Sierra). Reading the [link above](https://blog.mozilla.org/security/2016/10/24/distrusting-new-wosign-and-startcom-certificates/), it seems only certs issued after a point in time are being untrusted?

tehnick:
> are you seeing actual errors for ffmpeg?

```
boris@Tehnick:~/Opt$ lsb_release -d
Description:    Debian GNU/Linux unstable (sid)
boris@Tehnick:~/Opt$ LC_ALL=C wget -v https://ffmpeg.org/releases/ffmpeg-3.3.2.tar.xz
--2017-07-03 16:54:21--  https://ffmpeg.org/releases/ffmpeg-3.3.2.tar.xz
Resolving ffmpeg.org (ffmpeg.org)... 79.124.17.100
Connecting to ffmpeg.org (ffmpeg.org)|79.124.17.100|:443... connected.
ERROR: The certificate of 'ffmpeg.org' is not trusted.
ERROR: The certificate of 'ffmpeg.org' hasn't got a known issuer.
ERROR: The certificate of 'ffmpeg.org' was signed using an insecure algorithm. 
```

tehnick:
Latest news:
https://groups.google.com/a/chromium.org/forum/#!topic/net-dev/FKXe-76GO8Y
Older news:
https://security.googleblog.com/2016/10/distrusting-wosign-and-startcom.html
https://chromium.googlesource.com/chromium/src/+/e719fc626a3b9a528bf226b704785bcb24d07868
https://bugzilla.mozilla.org/show_bug.cgi?id=1309707

tonytheodore:
I've asked about this on the FFmpeg list:

http://ffmpeg.org/pipermail/ffmpeg-user/2017-July/036587.html

but a general solution/policy is still needed.

tehnick:
I have sent a message to [ffmpeg-devel](https://ffmpeg.org/pipermail/ffmpeg-devel/2017-July/date.html) few days ago, but this mailing list is premoderated for unregistered users, and my message still hasn't been published...

tehnick:
https://ffmpeg.org/pipermail/ffmpeg-devel/2017-July/thread.html#213715

vog:
@tehnick Thanks for making upstream (ffmpeg) aware of this issue, although it seems there won't be a quick solution.

Nevertheless, I disagree with degrading URLs in `WEBSITE` and `$(PKG)_UPDATE` from HTTPS to HTTP.

Regarding `WEBSITE`: We do have a working HTTPS link for `proj4`. Okay, they currently redirect that to a plain HTTP URL, but that is their choice and I hope they will revise this decision in the future. Then, the redirect will hopefully point to an HTTPS URL. No need to hard-code their currently bad decision into our URLs.

Regarding UPDATE scripts: I'd rather upgrade ffmpeg "by hand" until they fix their certificate, than going back to HTTP (or HTTPS without cert validation). The initial download is always critical, as the checksum is calculated from that and our mirrors are filled from that. It is dangerous to change that to HTTP just because of a temporary failure on upstream side.

tehnick:
@vog Ok. I have updated this PR in according to your requirements.

tonytheodore:
Somewhat related, note that every custom domain that's hosted by github will have an host name mismatch since it uses the `*.github.com` certificate (see https://mxe.cc). In these cases, I think we should just point to the github project instead of the website.

Thanks for you patience @tehnick !

tonytheodore:
ffmpeg uploaded to mirrors without intervention:
```
$ wget -q -O- http://d1yihgixbnrglp.cloudfront.net/ffmpeg-3.3.2.tar.bz2 | openssl sha256
(stdin)= 216900d5c0af9017bb8f76b0ad23f0ac53cf7fc618cf04b40d989bd99b088e6a
```

PR #1799, opened 2017-07-01
djvulibre: enable static library builds

darealshinji:


darealshinji:
I'm currently checking the suggestion from upstream to build with `-DDDJVUAPI= -DMINILISPAPI=` set (requires a bit of patching in libdjvu/DjVuGlobal.h). I will update and rebase the PR if it works as intended.

darealshinji:
Rebased commits. Now it only builds the library and not the unneeded tools.

tonytheodore:
Thanks!

PR #1800, opened 2017-07-01
homepage: s/MXE_TARGET/MXE_TARGETS/

starius:
See http://lists.nongnu.org/archive/html/mingw-cross-env-list/2017-06/msg00001.html


tonytheodore:
Thanks!

PR #1804, opened 2017-07-03
Adding graphviz package

hankhank:

  * install .pc file,

Thanks!

### DIFF
--- a/docs/packages.json
+++ b/docs/packages.json
@@ -89,6 +89,7 @@
     "googlemock": {"version": "1.7.0", "website": "https://github.com/google/googlemock", "description": "Google Mock"},
     "googletest": {"version": "1.7.0", "website": "https://github.com/google/googletest", "description": "Google Test"},
     "graphicsmagick": {"version": "1.3.21", "website": "http://www.graphicsmagick.org/", "description": "GraphicsMagick"},
+    "graphviz": {"version": "2.38.0", "website": "http://www.graphviz.org/", "description": "GraphViz"},
     "gsl": {"version": "2.3", "website": "https://www.gnu.org/software/gsl/", "description": "GSL"},
     "gsoap": {"version": "2.8.22", "website": "https://www.genivia.com/dev.html", "description": "gSOAP"},
     "gst-libav": {"version": "1.6.2", "website": "https://gstreamer.freedesktop.org/modules/gst-libav.html", "description": ""},
--- a/src/graphviz-1-remove-decl.patch
+++ b/src/graphviz-1-remove-decl.patch
@@ -1,0 +1,398 @@
+diff -ur a/ast_common.h b/ast_common.h
+--- a/ast_common.h	2016-08-11 20:45:52.985309265 +1000
++++ b/ast_common.h	2016-08-11 20:52:48.433612672 +1000
+@@ -134,10 +134,10 @@
+ /* dynamic linked library external scope handling */
+ #undef extern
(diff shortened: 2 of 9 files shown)


docs/packages.json
```diff
@@ -89,6 +89,7 @@
     "googlemock": {"version": "1.7.0", "website": "https://github.com/google/googlemock", "description": "Google Mock"},
     "googletest": {"version": "1.7.0", "website": "https://github.com/google/googletest", "description": "Google Test"},
     "graphicsmagick": {"version": "1.3.21", "website": "http://www.graphicsmagick.org/", "description": "GraphicsMagick"},
+    "graphviz": {"version": "2.38.0", "website": "http://www.graphviz.org/", "description": "GraphViz"},
```
tonytheodore: This file is generated and should be ignored by git

src/graphviz-1-remove-decl.patch
```diff
@@ -0,0 +1,398 @@
+diff -ur a/ast_common.h b/ast_common.h
+--- a/ast_common.h	2016-08-11 20:45:52.985309265 +1000
++++ b/ast_common.h	2016-08-11 20:52:48.433612672 +1000
+@@ -134,10 +134,10 @@
+ /* dynamic linked library external scope handling */
+ #undef extern
+ #if _dll_import && !defined(__EXPORT__) && _DLL_BLD
+-#define __EXPORT__      __declspec(dllexport)
```
tonytheodore: All these `__declspec` changes should be wrapped in macros in a single top-level header and then used throughout. 

See https://github.com/mxe/mxe/blob/c92ea065fb3773f90cf91983835b7b47a335d591/src/libepoxy-1-fixes.patch#L20-L31 for an example.

src/graphviz.mk
```diff
@@ -0,0 +1,23 @@
...
+$(PKG)_URL      := http://www.graphviz.org/pub/graphviz/stable/SOURCES/$($(PKG)_FILE)
+$(PKG)_DEPS     := gcc
+
+define $(PKG)_BUILD
+    cd '$(1)/cmd/lefty' && $(TARGET)-windres -l 0x409 -Iws/mswin32 -DNDEBUG -Ocoff -o lefty.res ws/mswin32/lefty.rc
+    cd '$(1)/lib/gvpr' && make mkdefs
+    cd '$(1)' && ./configure \
+	--host='$(TARGET)' \
+	--prefix='$(PREFIX)/$(TARGET)' \
+	--without-gdk-pixbuf --with-mylibgd --disable-static --disable-shared --disable-swig --without-x  --disable-tcl \
```
tonytheodore: Why both `--disable-static --disable-shared`?
Please use `$(MXE_CONFIGURE_OPTS)` for standard host, prefix, static etc. options and put the other options on separate lines.
Try to build out-of-source e.g:
```
    cd '$(BUILD_DIR)' && $(SOURCE_DIR)/configure \
        $(MXE_CONFIGURE_OPTS) \
        --without-foo \
        --disable-bar
    $(MAKE) -C '$(BUILD_DIR)' -j '$(JOBS)'
    $(MAKE) -C '$(BUILD_DIR)' -j 1 install
```

src/graphviz-1-remove-decl.patch
```diff
@@ -0,0 +1,398 @@
+diff -ur a/ast_common.h b/ast_common.h
```
tonytheodore: Patches should be in git format-patch style (like `graphviz-2-remove-regex.patch` below) instead of diff style

PR #1808, opened 2017-07-07
Updated cgal to 4.10.

rmarsa:

  * install .pc file,

Thanks!

### DIFF
--- a/src/cgal-TryRunResults.cmake
+++ b/src/cgal-TryRunResults.cmake
@@ -223,11 +223,20 @@ SET( OPENGL_RUN_RES__TRYRUN_OUTPUT
      CACHE STRING "Output from TRY_RUN" FORCE)
 
 
-SET( QT4_RUN_RES
+SET( QT_RUN_RES
      "0"
      CACHE STRING "Result from TRY_RUN" FORCE)
 
-SET( QT4_RUN_RES__TRYRUN_OUTPUT
-     "version=4.8.6"
+SET( QT_RUN_RES__TRYRUN_OUTPUT
+     "version=5.9.0"
+     CACHE STRING "Output from TRY_RUN" FORCE)
+
+
+SET( CGAL_test_cpp_version_RUN_RES
+     "0"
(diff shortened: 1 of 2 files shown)


tonytheodore:
It's missing a `TRY_RUN` entry:
```diff
diff --git a/src/cgal-TryRunResults.cmake b/src/cgal-TryRunResults.cmake
index 9e3acf4..1e9a6dd 100644
--- a/src/cgal-TryRunResults.cmake
+++ b/src/cgal-TryRunResults.cmake
@@ -231,3 +231,12 @@ SET( QT4_RUN_RES__TRYRUN_OUTPUT
      "version=4.8.6"
      CACHE STRING "Output from TRY_RUN" FORCE)

+set( CGAL_test_cpp_version_RUN_RES
+     "0"
+     CACHE STRING "Result from TRY_RUN" FORCE)
+
+set( CGAL_test_cpp_version_RUN_RES__TRYRUN_OUTPUT
+     "201103L"
+     CACHE STRING "Output from TRY_RUN" FORCE)
+
+
```
otherwise appears to build and install correctly, but the test program fails with boost references https://gist.github.com/tonytheodore/5ae4a30b801f3496326669c5009245fe.

tonytheodore:
Removing the tests simply moves any issues with the build/config downstream - we need these to confirm the package is actually usable.

rmarsa:
Yes.  So I know the package is useable because I am using it in production code.  The reason I removed those tests as part of the build is because for them to work, they need to link with boost_thread and boost_system.  This will involve changing CGAL since I would have to modify the build files for the CGAL examples.  As far as I can tell, these examples simply don't build correctly and I will have to submit a patch to CGAL to get them sorted.

The alternative is to make the mxe build patch them before building them and I can look into that.

tonytheodore:
Thanks, [patching to restore the tests](https://github.com/tonytheodore/mxe/blob/52195ec78243d4dd38d384cca191df04478342ed/src/cgal-1-fixes.patch) was fairly straightforward and this is merged in #1859 (along with some style changes)

PR #1813, opened 2017-07-09
openmp-validation: add current directory to @INC for newer Perl

mbunkus:
Perl v5.26.0 has removed the current directory from the include
path. This breaks the `run_test.pl` script which tries to load modules
from the build directory and fails if the system Perl is 5.26.0 or
newer.

Error message: `Can't locate ompts_parserFunctions.pm in @INC`


tonytheodore:
Thanks, patches consolidated https://github.com/mxe/mxe/commit/fda212c27372f4da7b46bc65f80fcd8d0ff1e00f

PR #1814, opened 2017-07-09
libxml++: update from 2.37.2 to 2.40.1

mbunkus:
This fixes the following error with gcc >= 6:

```
libxml++/io/istreamparserinputbuffer.cc: In member function 'virtual bool xmlpp::IStreamParserInputBuffer::do_close()':
libxml++/io/istreamparserinputbuffer.cc:42:12: error: cannot convert 'std::istream {aka std::basic_istream<char>}' to 'bool' in return
     return input_;
```


tonytheodore:
Thanks!

PR #1817, opened 2017-07-09
log4cxx: fix invalid narrowing conversion tripping up g++ >= 6

mbunkus:
g++ 6 is stricter regarding narrowing conversions in initializer
lists. This causes errors similar to the following during compilation
with the gcc6 or gcc7 overlays:

    locationinfo.cpp:163:21: error: narrowing conversion of '237'
    from 'int' to 'char' inside { } [-Wnarrowing]

This has already been fixed upstream in several consecutive patches,
but there's no release including those fixes yet. This patch consists
of the essence of said fixes.


tonytheodore:
Thanks, I added a patch header https://github.com/mxe/mxe/commit/eb34b4ea4250c36d951b14d7efce56de948be41f instead of consolidating these as they should all be removed next update.

PR #1822, opened 2017-07-14
binutils: update 2.25.1 --> 2.28

### LINKED ISSUE #1758: Cannot compile gcc on macOS: get "cannot compute suffix of object files"
https://gist.github.com/ronaaron/7a6fc224c0f13214dc7880574267a659

I was able, in the past, to compile on macOS.

tonytheodore:
* fixes original 2.26 regression with iostreams dll
    ```
    $ make gcc STRIP_TOOLCHAIN= MXE_TARGETS=i686-w64-mingw32.shared
    [...]
    $ i686-w64-mingw32.shared-nm  usr/i686-w64-mingw32.shared/bin/libstdc++-6.dll | grep InitC1
    6fefa2d0 T __ZNSt8ios_base4InitC1Ev
    ```
* replaces #1737
* fixes #1758
* tested on all four targets

WIP pending SOURCE_DATE_EPOCH patch from #1737.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -856,16 +856,6 @@ USE_OSGPLUGIN(&lt;plugin2&gt;)
     <h5>Certain packages have open issues on OS X</h5>
     </p>
     <p>
-    For Xcode ≥ 8.3 install an alternate compiler to build `gcc`.
-    For example, if you use Macports run:
-    <pre>sudo port install gcc5
-make cmake && \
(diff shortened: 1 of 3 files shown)


tonytheodore:
WIP removed, all four targets build with `SOURCE_DATE_EPOCH` patch.

I think this is good to merge, binutils historically doesn't introduce issues the way compiler updates do (anecdotally only, of course).

PR #1828, opened 2017-07-19
add qt5 freeze example plugin (e.g. winxp support)

tonytheodore:


tonytheodore:
The test runs on Win XP (with mysql removed) so this should be a starting point at least for people who need to do such things.

PR #1831, opened 2017-07-21
qtbase: fix CMake static linking dependencies

muellni:

As @mabrand said, it would be better to do this by adding a patch to src/qtbase-1-fixes.patch instead of a sed hack.
But I couldn't figure out where to put the dependencies. 
The `lib/cmake/Qt5Module/Qt5ModuleConfig.cmake` file is generated by the `mkspecs/features/data/cmake/Qt5BasicConfig.cmake.in` template file.
This template uses the `CMAKE_INTERFACE_QT5_MODULE_DEPS` variable which is fed by the `mkspecs/features/create_cmake.prf` script.
This script is called from `mkspecs/features/qt_module.prf`
I extracted the dependencies from the pkgconfig-file, which is still generated by qmake not by the prf scripts.

### DIFF
--- a/src/qtbase.mk
+++ b/src/qtbase.mk
@@ -102,6 +102,11 @@ define $(PKG)_BUILD
      printf 'test-qt5.exe\r\n'; \
      printf 'test-qtbase-pkgconfig.exe\r\n';) \
      > '$(PREFIX)/$(TARGET)/bin/test-qt5.bat'
+
+    # add libs to CMake config of Qt5Core to fix static linking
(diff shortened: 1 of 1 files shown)


tonytheodore:
I had some WIP on this (#1645), but it's very [hard to do neatly or keep updated](https://github.com/mxe/mxe/issues/1642#issuecomment-276876935). I'd say `sed` is acceptable in this case as it should only be temporary and a patch requires deep knowledge of the qt build system.

Thanks!

PR #1833, opened 2017-07-21
qt: set CMake QT_QMAKE_EXECUTABLE variable to cache

muellni:

IIRC this is needed to successfully run `find_package(Qt4)` multiple times


tonytheodore:
> needed to successfully run find_package(Qt4) multiple times

That rings a bell, thanks!

Issue #1772, opened 2017-05-19
Debian Packages errors


oomek:
```
(Reading database ... 387400 files and directories currently installed.)
Preparing to unpack .../mxe-x86-64-w64-mingw32.static-gtk3_3.22.7-20170208_amd64.deb ...
Unpacking mxe-x86-64-w64-mingw32.static-gtk3 (3.22.7-20170208) ...
dpkg: error processing archive /var/cache/apt/archives/mxe-x86-64-w64-mingw32.static-gtk3_3.22.7-20170208_amd64.deb (--unpack):
 trying to overwrite '/usr/lib/mxe/usr/x86_64-w64-mingw32.static/lib/gailutil.def', which is also in package mxe-x86-64-w64-mingw32.static-gtk2 2.24.29-20170208
Preparing to unpack .../mxe-x86-64-w64-mingw32.static-hdf-eos5_1.15-20170208_amd64.deb ...
Unpacking mxe-x86-64-w64-mingw32.static-hdf-eos5 (1.15-20170208) ...
dpkg: error processing archive /var/cache/apt/archives/mxe-x86-64-w64-mingw32.static-hdf-eos5_1.15-20170208_amd64.deb (--unpack):
 trying to overwrite '/usr/lib/mxe/usr/x86_64-w64-mingw32.static/include/bcea.h', which is also in package mxe-x86-64-w64-mingw32.static-hdf-eos2 19v1.00-20170208
dpkg-deb: error: subprocess paste was killed by signal (Broken pipe)
Preparing to unpack .../mxe-x86-64-w64-mingw32.static-id3lib_3.8.3-20170208_amd64.deb ...
Unpacking mxe-x86-64-w64-mingw32.static-id3lib (3.8.3-20170208) ...
dpkg: error processing archive /var/cache/apt/archives/mxe-x86-64-w64-mingw32.static-id3lib_3.8.3-20170208_amd64.deb (--unpack):
 trying to overwrite '/usr/lib/mxe/usr/x86_64-w64-mingw32.static/lib/libz.a', which is also in package mxe-x86-64-w64-mingw32.static-zlib 1.2.11-20170208
Preparing to unpack .../mxe-x86-64-w64-mingw32.static-pdcurses_3.4-20170208_amd64.deb ...
Unpacking mxe-x86-64-w64-mingw32.static-pdcurses (3.4-20170208) ...
[...]

Issue #1840, opened 2017-07-22
File conflicts between packages


davidebeatrici:
```
Unpacking mxe-x86-64-w64-mingw32.static-aubio (0.4.2-20170208) ...
dpkg: error processing archive /tmp/apt-dpkg-install-6a3ASn/615-mxe-x86-64-w64-mingw32.static-aubio_0.4.2-20170208_amd64.deb (--unpack):
trying to overwrite '/usr/lib/mxe/usr/x86_64-unknown-linux-gnu/bin/.waf-1.8.17-daa67386b71d29ad7cda321d44eb497a/waflib/ConfigSet.pyc', which is also in package mxe-x86-64-w64-mingw32.shared-aubio 0.4.2-20170208
```
```
Unpacking mxe-x86-64-w64-mingw32.static-gtk3 (3.22.7-20170208) ...
dpkg: error processing archive /tmp/apt-dpkg-install-6a3ASn/730-mxe-x86-64-w64-mingw32.static-gtk3_3.22.7-20170208_amd64.deb (--unpack):
trying to overwrite '/usr/lib/mxe/usr/x86_64-w64-mingw32.static/lib/gailutil.def', which is also in package mxe-x86-64-w64-mingw32.static-gtk2 2.24.29-20170208
```
```
Unpacking mxe-x86-64-w64-mingw32.static-hdf-eos5 (1.15-20170208) ...
dpkg: error processing archive /tmp/apt-dpkg-install-6a3ASn/743-mxe-x86-64-w64-mingw32.static-hdf-eos5_1.15-20170208_amd64.deb (--unpack):
trying to overwrite '/usr/lib/mxe/usr/x86_64-w64-mingw32.static/include/bcea.h', which is also in package mxe-x86-64-w64-mingw32.static-hdf-eos2 19v1.00-20170208
```
```
Unpacking mxe-x86-64-w64-mingw32.static-id3lib (3.8.3-20170208) ...
dpkg: error processing archive /tmp/apt-dpkg-install-6a3ASn/746-mxe-x86-64-w64-mingw32.static-id3lib_3.8.3-20170208_amd64.deb (--unpack):
trying to overwrite '/usr/lib/mxe/usr/x86_64-w64-mingw32.static/lib/libz.a', which is also in package mxe-x86-64-w64-mingw32.static-zlib 1.2.11-20170208
```
```
Unpacking mxe-x86-64-w64-mingw32.static-pdcurses (3.4-20170208) ...
[...]

PR #1844, opened 2017-07-24
fix various package conflicts

tonytheodore:
When `waf` executes, it creates a cache directory in:
`./usr/$(BUILD)/bin/.waf-<version>-<checksum>`

This causes conflicting files in `build-pkg` (see #1840) and should
be blacklisted.

### DIFF
--- a/src/gtk2.mk
+++ b/src/gtk2.mk
@@ -21,7 +21,7 @@ define $(PKG)_UPDATE
 endef
 
 define $(PKG)_BUILD
-    cd '$(1)' && ./configure \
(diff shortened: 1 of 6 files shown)


tonytheodore:
@starius, I've consolidated the other four PRs into this one. I think this covers all the conflicts from your [last report](https://gist.github.com/starius/05435e5095ca41ee062fc1338e6e736d) in #1759 and the ones in #1840.

tonytheodore:
Also fixes #1772.

Rebased onto current master - @starius, can you review please?

PR #1845, opened 2017-07-24
id3lib: add missing zlib dependency

tonytheodore:
If `zlib` isn't found an internal version is built that causes
conflicts with the actual `zlib` package (see #1772)


tonytheodore:
Consolidated to #1844

PR #1846, opened 2017-07-24
vtk: update to version 8 and removal of package vtk6

muellni:


tonytheodore:
Target `i686-w64-mingw32.static` on darwin succeeds, confirming `pcl` and other targets now.

PR #1847, opened 2017-07-24
gtk2/3: use new-style options and remove gailutil.def

tonytheodore:
`gailutil.def` causes package conflicts and `*.def` files aren't
really relevant for MXE (see #1772).


tonytheodore:
Consolidated to #1844

PR #1848, opened 2017-07-24
pdcurses: use new-style options and don't rename libs

tonytheodore:
Renaming to `libpanel` causes conflicts with `ncurses` (see #1772).

There are no packages in MXE that depend on `pdcurses` any longer,
other downstream users should be able to remove any workarounds they
have in place for the current renaming.


tonytheodore:
Consolidated to #1844

PR #1850, opened 2017-07-24
hdf-eos2: install to sub-directories

tonytheodore:
the `gctp` external project and some headers conflict with hdf-eos5
so install to sub-directories (see #1772).


tonytheodore:
Consolidated to #1844

PR #1854, opened 2017-07-25
cmake: add option to disable -isystem

tonytheodore:
Alternative to #1811, this seems like it should be fixed upstream
in gcc/cmake rather than individual packages


tonytheodore:
Builds with gcc5/6 on i686-w64-mingw32.static.

PR #1858, opened 2017-07-26
devil: use $(SED) instead of patch for EOL issues

### LINKED ISSUE #1856: devil build fails - different line endings
recent patch https://github.com/mxe/mxe/pull/1830/commits/4e3f9c4e4f253a2839a5bd8e817526d64c0a1a74#diff-62568a62f61fa47d59a30f0995c53661 fails to apply.

tonytheodore:
Can't figure out why the other two *.rc files patch correctly and
this one doesn't. Patch was created with `make export-patch-` with git
config `autocrlf = input`. Devil has it's own `.gitattributes` file
but nothing seems to work.

fixes #1856

Anyone have some pointers?


tonytheodore:
Still can't figure this out - I'll revisit it when we sweep up the use of `sed`.

PR #1860, opened 2017-07-26
[WIP] gcc6 fixes

tonytheodore:


tonytheodore:
I was never happy with this, a better approach may be to set `CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES` in the toolchain file - seems to work for cgal and hyperscan so far.

PR #1861, opened 2017-07-26
fix launch clementine_qt4 with clean settings

pavelvat:
As it turned out the clementine_qt4 have the same problem that Qt5 version of clementine (launch with clean settings) https://github.com/clementine-player/Clementine/issues/5297, https://github.com/clementine-player/Clementine/pull/5669, #1630 :
```1)build is successful, but crash on launch.```


tonytheodore:
Thanks!

PR #1862, opened 2017-07-29
updating to hamlib 3.1 + adding lusb-1.0 to pkgconfig

dl5rcw:
updating the package
adjusted lusb-1.0 to make the example compile with all 4 compilers (64bit required the lusb-1.0)

compiled and tested running debian 8 + ubuntu


tonytheodore:
`libusb` installs it's own *.pc file so adding a `Requires` section to the `hamlib.pc` would be better than additional libs.

dl5rcw:
@tonytheodore - added the Requires section as proposed. Thanks for the hint!

tonytheodore:
Thanks!

PR #1866, opened 2017-08-01
freetype: Don't autodetect harfbuzz.

TobiX:
This makes the harfbuzz support explicit: freetype-bootstrap does not
use harfbuzz, but freetype does.

This is inspired by a build error where I had a leftover pkgconfig file from harfbuzz, so freetype-bootstrap target was configured WITH harfbuzz, but failed the compile. Normally, this should not happen, but this makes the build a bit more reproducible - freetype-bootstrap always builds without harfbuzz, freetype always with.


tonytheodore:
Thanks!

PR #1738, opened 2017-04-01
[WIP] mxml: enable shared using cmake

tonytheodore:
From https://github.com/mxe/mxe/pull/1706#issuecomment-290213319 and https://github.com/michaelrsweet/mxml/pull/192.

@tehnick can you check if this works for you?

PR #1871, opened 2017-08-05
mxml and libomemo: enable shared

tonytheodore:
from #1738

@tehnick, can you confirm this is still correct?

### DIFF
--- a/src/libomemo-test.c
+++ b/src/libomemo-test.c
@@ -23,11 +23,11 @@ int main(int argc, char *argv[])
     uint32_t sid = 9178;
 
     omemo_message * msg_out_p;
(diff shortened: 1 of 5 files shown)


tehnick:
Hi. Sorry for a long delay. I have just checked this PR and it works fine for me. Thanks for your work on this.

tonytheodore:
Unfortunately, the approach of using cmake will only be viable if accepted by `mxml` upstream https://github.com/michaelrsweet/mxml/pull/192.

For MXE, it's probably easier to use `$(MAKE_SHARED_FROM_STATIC)` (as in `libomemo`).

tonytheodore:
Replaced with #1952

PR #1872, opened 2017-08-07
Add package file for Gumbo, an HTML5 parsing library

yipdw:
This PR adds a package file for [Gumbo](https://github.com/google/gumbo-parser), an HTML5 parsing library.  I have been able to compile Gumbo using this package file with the configuration

```
MXE_TARGETS := i686-w64-mingw32.shared.posix x86_64-w64-mingw32.shared.posix i686-w64-mingw32.static x86_64-w64-mingw32.static
MXE_PLUGIN_DIRS := plugins/gcc6
mingw-w64-headers_CONFIGURE_OPTS =: --enable-secure-api
```

Build checklist:

✔ install .pc file
✔ install bin/test-pkg.exe compiled with flags by pkg-config
✔ install .dll to bin/ and .a, .dll.a to lib
(N/A) use $(TARGET)-cmake instead of cmake
 ✔ build in `$(BUILD_DIR)` instead of `$(SOURCE_DIR)`
 ✔ do not run target executables with Wine
 (? [1]) do not download anything while building
  ✔ do not install documentation
  ✔ do not install .exe files except test and build systems

[1]: The package file doesn't download anything in the build phase, but I haven't checked whether the Gumbo build process proper downloads anything during build.

### DIFF
--- a/src/gumbo-test.c
+++ b/src/gumbo-test.c
@@ -1,0 +1,13 @@
+/*
+ * This file is part of MXE. See LICENSE.md for licensing information.
+ */
+
+#include <gumbo.h>
+
+int main()
+{
+    GumboOutput *output = gumbo_parse("<blink>Hello world!</blink>");
+    gumbo_destroy_output(&kGumboDefaultOptions, output);
+
+    return 0;
+}
--- a/src/gumbo.mk
+++ b/src/gumbo.mk
@@ -19,14 +19,6 @@ define $(PKG)_BUILD
         sbin_PROGRAMS= \
         noinst_PROGRAMS=
 
-    # create pkg-config files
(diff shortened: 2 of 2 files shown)


tonytheodore:
Nice, looks good. The package installs it's own *.pc file, so that section can be removed - otherwise this is ready to merge.

yipdw:
Cool, thanks for the tip.  I've removed the pkgconfig generation step from `gumbo.mk` and am currently rebuilding with that change to double-check things.

yipdw:
Rebuilding from a32c13a1c1805551156f39dd27ea486d71d9f3b6 seems to be okay on my test install.

tonytheodore:
All four default targets on macOS with gcc5 work also - thanks!

PR #1877, opened 2017-08-12
Fix #1876 by disabling the crypto code path in libarchive

### LINKED ISSUE #1876: libarchive uses undefined bcrypt symbol BCryptDeriveKeyPBKDF2
The package compiles fine, but trying to link against libarchive fails if one uses the affected code paths:

**test.c**:
```C
#include <archive.h>
#include <archive_entry.h>

int main() {
	struct archive *a = archive_read_new();
	archive_read_support_format_zip(a);

	return 0;
}
```

**Trying to compile**:
```
$ i686-w64-mingw32.static-gcc test.c $(i686-w64-mingw32.static-pkg-config --libs --cflags libarchive)
[...]/usr/lib/gcc/i686-w64-mingw32.static/5.4.0/../../../../i686-w64-mingw32.static/lib/../lib/libarchive.a(archive_cryptor.o): In function `pbkdf2_sha1':
[...]/tmp-libarchive-i686-w64-mingw32.static/libarchive-3.3.2/libarchive/archive_cryptor.c:78: undefined reference to `BCryptDeriveKeyPBKDF2@40'
```

This hasn't been a problem with old versions of libarchive, the update from July 24 introduced this bug.

phillipberndt:
This fixes #1876 by disabling libarchive's code path calling
BCryptDeriveKeyPBKDF2(). The code already has multiple code paths, with a stub
implementation as a last resort fallback, so this should not have any negative
consequences. (Other than, of course, that encrypted ZIP support is - at least
partially - disabled in mxe builds.)

### DIFF
--- a/src/libarchive-1-fixes.patch
+++ b/src/libarchive-1-fixes.patch
@@ -9,7 +9,7 @@ Subject: [PATCH] fix pkg-config
 
 
 diff --git a/build/pkgconfig/libarchive.pc.in b/build/pkgconfig/libarchive.pc.in
-index 1111111..2222222 100644
+index 95d7159..8288f31 100644
 --- a/build/pkgconfig/libarchive.pc.in
 +++ b/build/pkgconfig/libarchive.pc.in
 @@ -8,4 +8,4 @@ Description: library that can create and read several streaming archive formats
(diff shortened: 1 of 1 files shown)


tonytheodore:
Thanks, see #1878 to use `nettle` instead of `bcrypt` for these functions.

PR #1885, opened 2017-08-29
PCL: enabled 32-bit shared library build

hurzelchen:
I basically just tried running the PCL build with `PCL_SHARED_LIBS=ON`.

Only issue I ran into, was the mix up of Rpcrt4.lib and librpcrt4.a, which could be solved by replacing the library name shortly before building.


tonytheodore:
Thanks, there's a simpler way using `-DPCL_SHARED_LIBS=$(CMAKE_SHARED_BOOL)` in #1933.

PR #1888, opened 2017-08-31
Update opencv to 3.3.0

Lord-Kamina:
Changed patches that I could, removed ones that I couldn't.

Result builds and installs correctly.


starius:
It fails to build: https://gist.github.com/b7110e3bcc6177481163be4d3736eeaf
Probably protobuf dependency is missing.

Lord-Kamina:
Damn. I tried the shared target and that built fine.

Lord-Kamina:
Mind giving it a try now, @starius?

src/opencv.mk
```diff
@@ -52,12 +54,12 @@ define $(PKG)_BUILD
...
     # opencv builds it's own libIlmImf.a
     $(if $(findstring x86_64-w64-mingw32,$(TARGET)),\
-        $(SED) -i 's/OpenEXR//' '$(BUILD_DIR)/unix-install/opencv.pc')
-    $(SED) -i 's,share/OpenCV/3rdparty/,,g' '$(BUILD_DIR)/unix-install/opencv.pc'
-    $(INSTALL) -m755 '$(BUILD_DIR)/unix-install/opencv.pc' '$(PREFIX)/$(TARGET)/lib/pkgconfig'
-
+    $(SED) -i 's/OpenEXR//' '$(1).build/unix-install/opencv.pc')
+    $(SED) -i 's,share/OpenCV/3rdparty/,,g' '$(1).build/unix-install/opencv.pc'
+    $(INSTALL) -m755 '$(1).build/unix-install/opencv.pc' '$(PREFIX)/$(TARGET)/lib/pkgconfig'
+
```
starius: Remove trailing spaces, please.

src/opencv.mk
```diff
@@ -52,12 +54,12 @@ define $(PKG)_BUILD
     # openexr isn't available on x86_64-w64-mingw32
     # opencv builds it's own libIlmImf.a
     $(if $(findstring x86_64-w64-mingw32,$(TARGET)),\
-        $(SED) -i 's/OpenEXR//' '$(BUILD_DIR)/unix-install/opencv.pc')
-    $(SED) -i 's,share/OpenCV/3rdparty/,,g' '$(BUILD_DIR)/unix-install/opencv.pc'
-    $(INSTALL) -m755 '$(BUILD_DIR)/unix-install/opencv.pc' '$(PREFIX)/$(TARGET)/lib/pkgconfig'
-
+    $(SED) -i 's/OpenEXR//' '$(1).build/unix-install/opencv.pc')
```
starius: Why did you replace `$(BUILD_DIR)` with `$(1).build`? The later is deprecated.

src/opencv.mk
```diff
@@ -52,12 +54,12 @@ define $(PKG)_BUILD
...
+    $(SED) -i 's,share/OpenCV/3rdparty/,,g' '$(1).build/unix-install/opencv.pc'
+    $(INSTALL) -m755 '$(1).build/unix-install/opencv.pc' '$(PREFIX)/$(TARGET)/lib/pkgconfig'
+    
     '$(TARGET)-g++' \
-        -W -Wall -Werror -ansi -pedantic \
-        '$(1)/samples/c/fback_c.c' -o '$(PREFIX)/$(TARGET)/bin/test-opencv.exe' \
-        `'$(TARGET)-pkg-config' opencv --cflags --libs`
+      -v -W -Wall -Werror -ansi \
+      '$(1)/samples/cpp/fback.cpp' -o '$(PREFIX)/$(TARGET)/bin/test-opencv.exe' \
+      `echo \`i686-w64-mingw32.static-pkg-config opencv --cflags --libs | $(SED) -r 's/-llib([0-9a-zA-Z]+)/-l\1/g'\``
```
starius: This looks unclear. The output of pkgconfig should be usable as-is. Could you patch the .pc file instead, please?

starius: You can use `$$()` syntax instead of `, because it can be nested without escaping:
```
$$($$())
```
(We replace `$` with `$$`, otherwise `make` would consider it its macro.)

Also, why do we need nested at all? Why not
```
`i686-w64-mingw32.static-pkg-config opencv --cflags --libs | $(SED) -r 's/-llib([0-9a-zA-Z]+)/-l\1/g'`
```
?

src/opencv.mk
Lord-Kamina: Didn't mean to, I've got a merge conflict gone wrong; will probably end-up opening a new PR shortly

src/opencv.mk
Lord-Kamina: The reason for the initial dependency issues were pretty dumb; some libs (for example libtiff and libprotobuf) had their full name output like that onto the command so the result was -llibprotobuf for example, instead of -lprotobuf. I patched a couple I initially patched a couple I managed to find, but some of them I couldn't find, so I decided to go with that sed command.

Lord-Kamina: That's a fair point, I guess.

Lord-Kamina:
Closed in favor of #1891.

PR #1889, opened 2017-09-01
GIS related packages - New and updated

TriptoKnocker:
The rationale behind this branch is that I was attempting to build QGIS (http://qgis.org/en/site/), but found that many of the necessary dependent libraries are out of date - some of the image formats, and necessary geospatial libraries. Ultimately, I ended up using the host-toolchain and cross-compiled libraries to build QGIS natively on Windows (which is why there's no QGIS package), but the level of patching required to make QGIS itself a viable package is enough  work to give anyone a head-ache for days as it is rather fiddly.
At the very least I would like to offer my updates, and additional packages which are all useful in their own right.

The commits run in order of dependency - I think I've followed all of the required style and pull request rules, but if something's amiss, please feel free to send instruction so that I can fix it.

I've tested this on all 4 configurations on Ubuntu 16.04.03 with default GCC, and x86_64-w64-mingw32.shared with GCC7.2. (It allows me to successfully build QGIS natively in Windows)

One thing I'd like to deliberately point out to anybody reading this pull request message, is that I've changed gdal so that it will now also depend on spatialite. I'm also not convinced that the old dependency on postgres was working before, but feel free to challenge me about it.


tonytheodore:
Thanks for sharing these updates, smaller PRs are easier to review, but there doesn't seem to be too many changes required.

src/freexl.mk
```diff
@@ -0,0 +1,39 @@
...
+$(PKG)_DEPS     := gcc libiconv
+
+define $(PKG)_UPDATE
+    $(WGET) -q -O- 'http://www.gaia-gis.it/gaia-sins/freexl-sources/' | \
+    $(SED) -n 's,.*freexl-\([0-9][^>]*\)\.tar.*,\1,p' | \
+    tail -1
+endef
+
+define $(PKG)_BUILD
+    cd '$(1)' && ./configure \
```
tonytheodore: Try to use out-of-source builds (like qtkeychain below) using `$(BUILD_DIR)` and `$(SOURCE_DIR)` instead of `$(1)`

src/freexl.mk
```diff
@@ -0,0 +1,39 @@
...
+        $(MXE_CONFIGURE_OPTS)
+    $(MAKE) -C '$(1)' -j '$(JOBS)' $(MXE_DISABLE_PROGRAMS)
+    $(MAKE) -C '$(1)' -j 1 $(INSTALL_STRIP_LIB)
+
+    # the test program comes from the freexl sources itself (test_xl.c)
+    '$(TARGET)-g++' \
+        -W -Wall -Werror -ansi -pedantic \
+        '$(1)/examples/test_xl.c' -o '$(PREFIX)/$(TARGET)/bin/test-freexl.exe' \
+        `'$(TARGET)-pkg-config' $(PKG) --cflags --libs` -liconv
```
tonytheodore: Ideally, extra libs (like `-liconv`) shouldn't be needed with a correct pkg-config file.

src/freexl.mk
```diff
@@ -0,0 +1,39 @@
...
+
+    # create a batch file to run the test program (as the test program requires arguments)
+    (printf 'REM Run the test program against the provided .xls file.\r\n'; \
```
tonytheodore: Does `\r\n` need extra escaping? i.e. `\\r\\n`

src/spatialite.mk
```diff
@@ -0,0 +1,44 @@
...
+    $(SED) -n 's,.*libspatialite-\([0-9][^>]*\)\.tar.*,\1,p' | \
+    tail -1
+endef
+
+define $(PKG)_BUILD
+    # freeXL support is only necessary if you want to be able to parse .xls files.
+    # If you disable freexl support, remove -lfreexl from the test program below.
+    cd '$(1)' && autoreconf -fi -I ./m4
+    cd '$(1)' && ./configure \
+    $(MXE_CONFIGURE_OPTS) \
```
tonytheodore: Indentation

src/spatialite.mk
```diff
@@ -0,0 +1,44 @@
...
+    --enable-freexl=yes \
+    --with-geosconfig='$(PREFIX)/$(TARGET)/bin/geos-config'
+    $(MAKE) -C '$(1)' -j '$(JOBS)' $(if $(BUILD_SHARED), LDFLAGS='-no-undefined')
+    $(MAKE) -C '$(1)' -j 1  $(INSTALL_STRIP_LIB)
+
+    # compile one of the demo programs (copied from the source package)
```
tonytheodore: Could build test directly from source instead of copying to MXE tree.

src/spatialite.mk
```diff
@@ -0,0 +1,44 @@
...
+    '$(TARGET)-g++' $(if $(BUILD_SHARED), -Wno-undefined) \
+       -W -Wall -Werror -ansi -pedantic \
+       '$(TOP_DIR)/src/spatialite-test.c' -o '$(PREFIX)/$(TARGET)/bin/test-spatialite.exe' \
+       `'$(TARGET)-pkg-config' $(PKG) --cflags --libs` \
+       `'$(TARGET)-pkg-config' sqlite3 --cflags --libs` \
+       $(if $(BUILD_STATIC), -lgeos_c -lgeos -lfreexl -lxml2 -liconv -llzma -lproj -lws2_32 -lz -lstdc++ -lm)
```
tonytheodore: Really shouldn't need the extra calls to `pkg-config` and manual libs.

tonytheodore:
> Ultimately, I ended up using the host-toolchain and cross-compiled libraries to build QGIS natively on Windows (which is why there's no QGIS package), but the level of patching required to make QGIS itself a viable package is enough work to give anyone a head-ache for days as it is rather fiddly.

Interesting, I wouldn't have thought the host-toolchain would be good for complex projects.

src/freexl.mk
TriptoKnocker: I must confess that I borrowed the idea from qtbase.mk . Double-escaping doesn't seem necessary as the produced batch file correctly line-feeds if I open it up in notepad.

src/spatialite.mk
TriptoKnocker: I had considered this to a degree - the test in the MXE tree is subtly edited so that it will compile with the -Werror flag set (as far as I recall, there are pointer casting warnings that prevent the test from compiling.)
Should I relax the test's compilation flags to be able to build the test directly from the source path?

TriptoKnocker:
OK, thanks for the feedback - I'll make the changes as I'm able to.
I've learned a lot about compilation in general thanks to MXE. 
> smaller PRs are easier to review, but there doesn't seem to be too many changes required.

I wasn't sure if one guy spamming many small PRs at once would go down well, so I took a chance.

> Interesting, I wouldn't have thought the host-toolchain would be good for complex projects.

The host-toolchain works very well - I'm using MXE to produce a host version of GCC7.2. Although the qt-host-tools plugin produces qmake.exe some of the other necessary executable are produced as ELF executables - it might just be a case that qt-tools should also be built.
QGIS itself's build-scripts support MinGW, but I don't think they've been updated for an age. There are several places where the CMake makefiles will stop looking for some very windows-specific headers or locations - but we're in contact with the QGIS devs so perhaps I'll be able to make a QGIS app package yet.

src/freexl.mk
tonytheodore: Indeed, you're right - it's only the windows paths that need escaping.

src/spatialite.mk
tonytheodore: Our tests are mostly to confirm that linking works and serve as an example of how one might use whatever configuration system the package provides. Strict flags are appropriate for hand-written tests; they can be relaxed when using existing sources.

The runtime side of the test can also be relaxed, creating batch files to set env vars (for example) is a good thing but trying to exercise the functionality of a package is mostly out of our scope.

Our time is better spent on improving the configuration for downstream users (via pkg-config, cmake, et al) so they can incorporate packages with minimal manual specification of libs, flags, etc.

tonytheodore:
> I wasn't sure if one guy spamming many small PRs at once would go down well, so I took a chance.

I think there's a doctoral thesis in that - it's reasonably clear to the reviewer, but opaque to the submitter ;)

> I'm using MXE to produce a host version of GCC7.2

This will likely break after #1915 and #1918 won't fix it.

>The host-toolchain works very well [...] some of the other necessary executable are produced as ELF executables - it might just be a case that qt-tools should also be built.

Could you raise a new issue for this? The host-toolchain wasn't intended to be used, but if it's useful, any improvements and usage notes are welcome.

TriptoKnocker:
I've given the issues a poke - and things seem to be for the better. While I was at it, I've updated gdal again, which allowed me to remove the patch that I'd made for linking against openjpeg-2.2, which gdal was not able to detect before.
Spatialite I'm starting to dislike - rather than linking directly a lengthy list of libraries for the test/usage, I've farmed off as much as I'm able to pkg-conf. If this isn't desired, I'm not sure what to do further - seemingly a static build of spatialite's test requires the lengthy list of libs.

> This will likely break after #1915 and #1918 won't fix it.
This is a bit disheartening to hear, but perhaps solvable before the next tool upgrade cycle begins. For context, I've used MSYS2 in the past to keep a toolchain & Qt5 and very much prefer the idea of 'simply' cross-compiling via MXE as the packages are far more up to date, it's simply faster to compile the packages, and allows me a means to make use of the latest GCC to keep up with my linux builds.

> Could you raise a new issue for this? The host-toolchain wasn't intended to be used
Sure. I'm somewhat surprised by this - I realise that the host-toolchain is an example of a plugin, but it has genuinely helped in giving me a viable upgrade path for the toolchain we use for windows (as we  need the posix threading.)

tonytheodore:
There are couple of other style changes here https://gist.github.com/tonytheodore/88dc95debef736a52d164f6357747806 (I'm finding github reviews hard to follow) Basically:
 - avoid extra whitespace/empty lines (gta.mk)
 - always list `gcc` in `_DEPS` for cross-packages (spatialite.mk)
 - separate $(BUILD_DIR) and $(SOURCE_DIR) (freexl.mk, libspatialindex.mk, spatialite.mk)

The last allows building from a local repo/tree with 
```
make foo foo_SOURCE_TREE=/path/to/foo/source
```

> I've farmed off as much as I'm able to pkg-conf. If this isn't desired

That is the best approach, manually specifying libs/flags is too error-prone and putting them in pkg-config helps downstream packages and other users. Ideally:
 - use `Libs.private` and `Cflags.private` for static
 - use `Requires` and `Requires.private` to minimise manual specification of deps
 - adjust `configure.ac` so those flags reflect the build-time options and submit upstream...

I know that is an increasing amount of work, it's not essential but new packages are the best chance to get such things done.

tonytheodore:
> separate $(BUILD_DIR) and $(SOURCE_DIR)

Ignore for `libspatialindex`, it doesn't seem to support out-of-source builds.

tonytheodore:
`geos` downstream package `ossim` fails:
https://gist.github.com/tonytheodore/973ab197e7c13b2eac8a387ce180d5c7#file-ossim_i686-w64-mingw32-static-L516

TriptoKnocker:
This should be the style updates complete.
I'll mark this as WIP, as maybe the solution for ossim is to try and update it? I see the old repo for it is not being maintained as they've moved to github.

tonytheodore:
See https://github.com/tonytheodore/mxe/commit/ed4074ac0c49a2b6f0f72459ce0d53b9679e96ec  for the core ossim update.

tonytheodore:
All looks good to me now - thanks! If you can add the extra ossim packages as new (smaller) PRs, that would be great.

I've replaced this with #1926 to get rid of the doubled commits.

PR #1890, opened 2017-09-01
add qrencode.mk

nohea:
I've created a new qrencode.mk for the libqrencode library. 

I've successfully used it with MXE on my Ubuntu 14.04 x64 box, with target i686-w64-mingw32.static


src/qrencode.mk
```diff
@@ -0,0 +1,40 @@
...
+    # create pkg-config files
+    $(INSTALL) -d '$(PREFIX)/$(TARGET)/lib/pkgconfig'
+    (echo 'Name: $(PKG)'; \
+     echo 'Version: $($(PKG)_VERSION)'; \
+     echo 'Description: libqrencode'; \
+     echo 'Libs: -lqrencode';) \
+     > '$(PREFIX)/$(TARGET)/lib/pkgconfig/$(PKG).pc'
+
+endef
+
```
starius: Remove this empty line, please.

src/qrencode.mk
```diff
@@ -0,0 +1,40 @@
...
+    $(MAKE) -C '$(BUILD_DIR)' -j 1 install
+
+    # create pkg-config files
+    $(INSTALL) -d '$(PREFIX)/$(TARGET)/lib/pkgconfig'
+    (echo 'Name: $(PKG)'; \
+     echo 'Version: $($(PKG)_VERSION)'; \
+     echo 'Description: libqrencode'; \
+     echo 'Libs: -lqrencode';) \
+     > '$(PREFIX)/$(TARGET)/lib/pkgconfig/$(PKG).pc'
+
```
starius: Add a test, please.

starius:
See also https://github.com/mxe/mxe/pull/1094
That pull request has some patches.

src/qrencode.mk
nohea: Done.

src/qrencode.mk
nohea: I did update the .mk build command to run --with-tests , which builds the test exes. This was enough to identify a linking problem with libpng, which i fixed by adding a dependency and -lpng. 

However, i'm not sure how to run the test exe binaries in the mxe environment on Linux. 

If i add a line like this:
`$(PREFIX)/$(TARGET)/bin/test_qrencode.exe`

... i get this:
```
'/home/raulg/cc/utils/mxe/tmp-qrencode-i686-w64-mingw32.static/qrencode-3.4.4.build_'/tests/test_qrencode.exe
bash: /home/raulg/cc/utils/mxe/tmp-qrencode-i686-w64-mingw32.static/qrencode-3.4.4.build_/tests/test_qrencode.exe: cannot execute binary file: Exec format error
make[1]: *** [build-only-qrencode_i686-w64-mingw32.static] Error 126
```

nohea:
FYI - i reviewed pull request #1094 and i do not see a reason to include those patch files, when the flags may be added directly to the qrencode.mk file.

tonytheodore:
Some comments:
  - build doesn't produce shared libs
  - libpng and sdl are only required for tools/tests
  - pkg-config file is installed by project

I've addressed these by updating, switching to cmake, and building test with supplied *.pc file: https://github.com/mxe/mxe/commit/a9982192ff679f20bfc8f0038eb252339d017f3a

Thanks!

nohea:
Thank you for getting this in!

PR #1898, opened 2017-09-13
compile qt4 with xmlpatterns module

yhfudev:
The xmlpatterns module is disabled by qt4 configure by default.


tonytheodore:
Thanks!

PR #1899, opened 2017-09-16
secure/update URLs

vszakats:
* change `libssh2`/`apache`/`xiph` URLs to canonical ones
* resubmit patch to remove unnecessary
  `--no-check-certificate` option. The URLs
  are non-HTTPS anyway
* restore URL for `blas`, it's working again

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -267,9 +267,9 @@ ESCAPE_PKG = \
 
 BACKUP_DOWNLOAD = \
     (echo "MXE Warning! Downloading $(1) from backup." >&2 && \
-    ($(WGET) --no-check-certificate -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' $(PKG_MIRROR)/`$(call ESCAPE_PKG,$(1))` || \
-    $(WGET) --no-check-certificate -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' $(PKG_CDN)/`$(call ESCAPE_PKG,$(1))` || \
-    $(WGET) --no-check-certificate -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' $(GITLAB_BACKUP)/`$(call ESCAPE_PKG,$(1))`_$($(1)_CHECKSUM)))
+    ($(WGET) -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' $(PKG_MIRROR)/`$(call ESCAPE_PKG,$(1))` || \
+    $(WGET) -O '$(PKG_DIR)/.tmp-$($(1)_FILE)' $(PKG_CDN)/`$(call ESCAPE_PKG,$(1))` || \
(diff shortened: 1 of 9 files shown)


starius:
http://www.netlib.org/blas/blas.tgz returns a different file now (sha256=c5a654351f0b046a502bf04b16740b9ab49c7d8512d6d57ad3a64184c8e575c3).

http://www.netlib.org/blas/blas-3.5.0.tgz returns the file we want (sha256=ef7d775d380f255d1902bce374ff7c8a594846454fcaeae552292168af1aca24).

Can you amend the commit, please?

Also please rebase your PR:
```
git pull --rebase https://github.com/mxe/mxe master
```

vszakats:
Fixed, squashed and rebased on master.

vszakats:
@starius: Slightly off-topic question: Do you have plans to setup HTTPS (f.e. using Let'S Encrypt) for the `http://pkg.mxe.cc/` server?

starius:
HTTPS using Let's Encrypt for pkg.mxe.cc sounds good for me.
I think we should also establish HTTPS for the main domain mxe.cc.
I've opened an issue to track this: https://github.com/mxe/mxe/issues/1909

starius:
The PR looks good. Thank you very much!

vszakats:
Thank you too! Looking forward for the HTTPS updates, and happy about mxe.cc, too, of course.

Another slight off: Regarding the three mirrors, I believe it'd be more ideal to turn the first two of them to HTTPS and only leave a single one as HTTP fallback for those users who need it. All do support HTTPS, so it's just a matter of decision. What do you think?

starius:
GITLAB_BACKUP is currently out-of-date (I am looking why).
We need at least one of PKG_MIRROR and PKG_CDN to stay http until the issue with GITLAB_BACKUP is fuxed. I think we can update one of them to https now.

vszakats:
Great! The first backup is S3, so it'd be best to enable it first. Once gitlab gets in line, we can do the CloudFront one.

tonytheodore:
GITLAB_BACKUP redirects to https.

vszakats:
Good news! Also `cloudfront.net` supports HTTPS — any reason why HTTP is used there?

tonytheodore:
> cloudfront.net supports HTTPS — any reason why HTTP is used there?

Mostly since the primary purpose of backup mirrors is to make things "accessible". Our content checksums are sufficient to ensure the integrity of files and http causes less operational noise than https (as ftp once did). Downloads must be the single largest number of questions we get.

It's hard to imagine clients having issues with AWS, Gitlab etc., and since #1756 it's easy to add options for stricter or looser protocols.

vszakats:
Does it sound reasonable to upgrade GitLab to HTTPS (if it redirects there anyway) and move CloudFront to the _last_ backup mirror, keeping it as HTTP?

tonytheodore:
+1 https://github.com/mxe/mxe/pull/1974/commits/b82080ec4d304e096d7bbd6a023838ac691d2b87

vszakats:
Perfect, thank you!

PR #1903, opened 2017-09-23
fix typos found by misspell tool

starius:
The tool: https://github.com/client9/gospell


starius:
UPD: the tool is https://github.com/client9/misspell

Issue #1886, opened 2017-08-30
build-pkg issues


hurzelchen:
I'm currently having to minor issues with build-pkg:

- It doesn't create a meta-package for `qt5` that would just simply reference all dependencies.
- the packages for `mxe-conf` and `cmake-conf` contain conflicting `.gitkeep` files.

hurzelchen:
the `x86-64-pc-linux-gnu` versions of `mxe-conf` and `cmake-conf` also both contain `nonetwork.so`.

hurzelchen:
and a last more cosmetic issue: the `Installed-Size` field of the debian packages are not set, so `apt-get` always reports no new disk space to be used.

starius:
Binary package for qt5 is not created because the package itself is empty.
Empty packages such that no non-empty packages depend on them are skipped after https://github.com/mxe/mxe/issues/968
Frankly speaking I don't remember what was the issue with empty packages, probably just a matter of taste. I'll try to change build-pkg to build such packages. Also it is likely to result in its simplification.

starius:
I've sent the fix in https://github.com/mxe/mxe/pull/1906
Can you verify that it builds the qt5 package, please?

starius:
With regard to the conflicting .gitkeep files, I identified all such files:
```
$ cat *.list | grep .gitkeep | sort | uniq -d
usr/i686-w64-mingw32.shared/installed/.gitkeep
usr/i686-w64-mingw32.static/installed/.gitkeep
usr/x86_64-unknown-linux-gnu/installed/.gitkeep
usr/x86_64-unknown-linux-gnu/lib/.gitkeep
usr/x86_64-w64-mingw32.shared/installed/.gitkeep
usr/x86_64-w64-mingw32.static/installed/.gitkeep
```

In all cases a package installs other files to the same library, so these `.gitkeep` files can be safely added to the blacklist. I'll send a PR.

starius:
The PR to ignore these .gitkeep files: https://github.com/mxe/mxe/pull/1907

starius:
https://github.com/mxe/mxe/pull/1908 was sent to address the Installed-Size issue.

Issue #968, opened 2015-11-04
build-pkg should not create empty packages


starius:
More than 20% of packages created by build-pkg are empty. Empty package is a package that installs only file `usr/<target>/installed/<package>`.

```
$ wc -l *.list | egrep '\s1\s' | wc -l
328

$ ls *.list | wc -l
1555
```

I think build-pkg should omit such packages and report a package depending on an empty package.

PR #1906, opened 2017-09-24
build-pkg: pack empty packages as well

starius:
See https://github.com/mxe/mxe/issues/1886#issue-253920956
See https://github.com/mxe/mxe/issues/968

### DIFF
--- a/tools/build-pkg.lua
+++ b/tools/build-pkg.lua
@@ -892,10 +892,6 @@ local function progressPrinter(items)
     return printer
 end
 
-local function isEmpty(files)
-    return #files == 1
-end
-
 -- build all packages, save filelist to list file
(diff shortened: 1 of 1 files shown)


tonytheodore:
@starius from https://github.com/mxe/mxe/issues/1886#issuecomment-331718504:

>Frankly speaking I don't remember what was the issue with empty packages, probably just a matter of taste.

I think the issue is that there's no way to distinguish a disabled package from a download/dependency package. Installing an empty disabled package is misleading while virtual packages are useful.

It may be best to add some new metadata (such as `$(PKG)_IS_VIRTUAL`) to the latter so they can be easily identified and handled by the `Makefile` and `print-deps-for-build-pkg` - this looks like the correct behaviour for `build-pkg`, thanks!

PR #1910, opened 2017-09-24
bump primary package mirror to HTTPS

vszakats:
Ref: https://github.com/mxe/mxe/pull/1899#issuecomment-331723251


starius:
Thanks!

PR #1922, opened 2017-10-06
Improve plugin patch handling and fix gcc7/gcc-host related issues

tonytheodore:
From https://github.com/mxe/mxe/pull/1889#issuecomment-333657206, keep host-toolchain usable.


tonytheodore:
@TriptoKnocker, do you use the host qt5-tools or cmake? They're both failing for me but I suspect the qt5 is mostly related to OSX.

TriptoKnocker:
It's a complicated procedure that I must still describe, but at the moment neither. Using the standard qt5 binaries produced by mxe, I use windows with a host toolchain to make a sort-of Frankenstein's monster version of qt5 to produce qmake, and the other qt-tools and then splice them together.

tonytheodore:
Okay, I'll revisit qmake down the track, I also recall it being fiddly to get right. This PR seems good for gcc7 and gcc-host at least.

PR #1931, opened 2017-10-10
Add package sqlcipher

### LINKED ISSUE #1911: Add support for sqlcipher
Hi, would it generally be possible to add the sqlcipher library? Its a sqlite variant with builtin encryption support. See https://github.com/sqlcipher/sqlcipher

tonytheodore:
Closes #1911.

Installs the executable since that falls into the (small) class of utilities that help developers. The `tcl` bindings can remain disabled until someone investigates what's needed.

@aquiles2k @starius, open issues:
 - requires `tcl` to build - should we do a native build or add it to requirements?
 - test program appears to be simply an `sqlite` test - I couldn't find any minimal examples of usage for a test case.

### DIFF
--- a/plugins/tcl.tk/tcl.mk
+++ b/plugins/tcl.tk/tcl.mk
@@ -1,30 +1,5 @@
 # This file is part of MXE. See LICENSE.md for licensing information.
 
 PKG             := tcl
-$(PKG)_WEBSITE  := https://tcl.tk/
-$(PKG)_OWNER    := https://github.com/highperformancecoder
-$(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 8.6.4
-$(PKG)_CHECKSUM := 9e6ed94c981c1d0c5f5fefb8112d06c6bf4d050a7327e95e71d417c416519c8d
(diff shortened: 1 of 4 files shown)


aquiles2k:
Thanks, looking good! 

But why do you want to build tcl when its disabled? 

The mptest as already stated seems to be just a sqlite multiprocess(mp?) test as for generally testing that the sqlcipher sqlite variant work in principal. 

Thats what homebrew does in its formula [(link)](https://github.com/Homebrew/homebrew-core/blob/master/Formula/sqlcipher.rb) with the simple sql create table test. For actually testing the encryption one would need to extend this by also setting the passphrase and testing against that.

tonytheodore:
> But why do you want to build tcl when its disabled?

Native `tclsh` is required to generate parts of the build system, I just added it https://github.com/mxe/mxe/pull/1931/commits/1a293c0fab6b48b42abdc8f0c8aef0b7c9dac8a3.

> as for generally testing that the sqlcipher sqlite variant work in principal

Yes, your right. This will test the linking as we require, I think I was just confused about what this package is.

I think this is good to go.

Issue #1901, opened 2017-09-16
Cannot build Qt 5.7.x for Win XP compatibility with qt5-freeze


uwehermann:
Hi, I'm aware that this is more or less just an example, but it would be nice if qt5-freeze could actually be used as is, without modifications, to build Qt 5.7.x on recent MXE (instead of the latest Qt5) in order to retain Windows XP compatibility.

I'm trying to build MXE from scratch (for building PulseView, see here: https://sigrok.org/gitweb/?p=sigrok-util.git;a=blob;f=cross-compile/mingw/sigrok-cross-mingw) as we usually do, with the only change being that I try to force building of Qt 5.7.x via the qt5-freeze plugin:

```
make MXE_TARGETS=i686-w64-mingw32.static.posix MXE_PLUGIN_DIRS=plugins/examples/qt5-freeze gcc glib libzip libusb1 libftdi1 glibmm boost check qt5 qtbase_CONFIGURE_OPTS='-no-sql-mysql'
```

It appears to work mostly OK, it does indeed build Qt 5.7.x, but then fails when it tries to download qtspeech-opensource-src-5.7.1.tar.xz, probably because that wasn't available yet in Qt 5.7.x.

I'm guessing the qt5.mk needs fixing to build its dependency list from qt5-freeze instead of src?

tonytheodore:
Indeed, the deps also need to be frozen - something like this should work (untested):
```
# parent of Qt 5.8 update
git checkout a0f9e61
export QT5_PKGS=`grep -l qtbase_VERSION src/*.mk | sed -n 's,src/\(.*\)\.mk.*,\1,p' | tr '\n' ' ' | sed -n 's,qt5,qtbase,p'`

git checkout master
echo "\$(PKG)_DEPS     := $QT5_PKGS" >> plugins/examples/qt5-freeze/qt5.mk
```
or at least be a starting point for manual edits to qt5.mk.

tonytheodore:
> I'm guessing the qt5.mk needs fixing to build its dependency list from qt5-freeze instead of src?

I didn't read to the end, that's a better approach!

 Replacing `$(TOP_DIR)/src` with `$(dir $(lastword $(MAKEFILE_LIST)))` will read from current directory of `qt5.mk`.

uwehermann:
Looks good, but there's one additional issue which pops up with the following command-line:

```
[...]

PR #1932, opened 2017-10-10
qttools: read test from current directory and freeze example 5.7 plugin

tonytheodore:
see: https://github.com/mxe/mxe/issues/1901#issuecomment-333551106

@uwehermann, this should be the last issue with the freeze example - can you confirm?


tonytheodore:
Tested with and without plugin, all seems good.

uwehermann:
Confirmed, seems to build fine. Thanks!

Issue #1518, opened 2016-09-10
gettext fails to build on Bash on Ubuntu on Windows (Ubuntu 14.04)


mshindal:
https://gist.github.com/mshindal/f4e02701f32dcb170bd4070c73f1a8e2

mshindal:
That didn't seem to work. Just to clarify, you're saying I should run `ln -s ../automake.mk plugins/native/trusty/automake.mk` inside of my cloned `mxe` directory?

darealshinji:
Yes.

mshindal:
I get this error: 
```make: *** No rule to make target `/mnt/d/Code/mxe/usr/x86_64-unknown-linux-gnu/installed/autoconf', needed by `/mnt/d/Code/mxe/usr/x86_64-unknown-linux-gnu/installed/automake'.  Stop.```

**Comment by darealshinji** [comment]
Ah, dependencies. Create links for `autoconf.mk` and `m4.mk` too and try it again.

**Comment by tonytheodore** [comment]
I'm not sure manual symlinks are the best way to go (you want `libtool` also), either of:

```
make OS_SHORT_NAME=wheezy ...
```
or
```
make libtool MXE_PLUGIN_DIRS=plugins/native MXE_TARGETS=`ext/config.guess`
```

will build a working autotools setup.

The core issue is why `aclocal` is called at all? We don't modify any of the base files so it shouldn't be needed for this package.

darealshinji:
My guess is that Ubuntu on Windows 10 might behave different than a real Ubuntu system. Last time I tried to compile something there it didn't work at all.

tonytheodore:
Good point @darealshinji, I haven't used it before - I wonder if there's some weirdness with timestamps on the files extracted from the tarball?

starius:
@mshindal 
What is the output of `lsb_release -sc` and `uname -s`?

mshindal:
`trusty` and `Linux`.

starius:
@tonytheodore
Why not to build these native packages on trusty to make it work out-of-box?

darealshinji:
It seems to be working now.

Saniq:
@darealshinji it might be fixed due to the newest version of _Bash on Ubuntu on Windows_ coming with Ubuntu 16.04 which in turn has automake-1.15 in its package source. But the underlying issue still seems to persist (at least for me on Windows 10 Creators Update (1703 / 15063.332)).

[...]

PR #1949, opened 2017-10-29
native plugin fixups

### LINKED ISSUE #1892: Problem with `gettext` (when basing on tutorial)
I follow the tutorial from http://mxe.cc/#tutorial and get stuck on simple `make`... It looks like `gettext` package fails to make itself.

Logs:
https://gist.github.com/PsychoXIVI/53c08dbf93d5f3fbc3652dce8503f69e

`tmp-gettext-x86_64-pc-linux-gnu`  folder of `mxe` after trying `make` again:
https://files.fm/u/57vfk4m2

In `settings.mk` file I uncommented the line with `MXE_TARGETS` (no. 15).

I am using Ubuntu 14.04, with gettext 0.18.3 (gnu runtime) . Tried `apt update`/`upgrade` - no better results.

By the way: I am not sure this should use `x86_64-pc-linux-gnu` instead of, as example, `i686-w64-mingw32.static` (from `settings.mk` file).

tonytheodore:
- gettext requires recent autotools
      - add for Ubuntu 14.04 (trusty) - fixes #1892
      - add deps for Debian Jessie and Wheezy
  - libtool is built for prior Debian/Ubuntu, remove from docs
  - tidy up darwin overrides


tonytheodore:
Also fixes #1518

PR #1952, opened 2017-10-30
mxml and libomemo: enable shared

tonytheodore:
Replaces #1871


tonytheodore:
@tehnick, let me know if any issues with this. In the case of all functions being exported to *.dll, it's easier for MXE to simply use `$(MAKE_SHARED_FROM_STATIC)`.

Downstream usage of `libomemo` should also be simplified with use of `pkg-config`. E.g `src/libomemo-test.cmake`:
```cmake
find_package(PkgConfig REQUIRED)
pkg_check_modules(LIBOMEMO libomemo)

include_directories(${LIBOMEMO_INCLUDE_DIRS})
target_link_libraries(${TGT} ${LIBOMEMO_LIBRARIES})
```

PR #1900, opened 2017-09-16
the new version uses nasm 2.13 +

Rinat84:
nasm.mk for x264

```
PKG             := nasm
$(PKG)_WEBSITE  := http://www.nasm.us/
$(PKG)_DESCR    := nasm
$(PKG)_IGNORE   := 
$(PKG)_VERSION  := 2.13.01
$(PKG)_CHECKSUM := aa0213008f0433ecbe07bb628506a5c4be8079be20fc3532a5031fd639db9a5e
$(PKG)_SUBDIR   := $(PKG)-$($(PKG)_VERSION)
$(PKG)_FILE     := $(PKG)-$($(PKG)_VERSION).tar.xz
$(PKG)_URL      := http://www.nasm.us/pub/$(PKG)/releasebuilds/$($(PKG)_VERSION)/$($(PKG)_FILE)
$(PKG)_DEPS     :=

define $(PKG)_UPDATE
    $(WGET) -q -O- 'http://www.nasm.us/pub/nasm/releasebuilds/?C=M;O=D' | \
    $(SED) -n 's,.*href="\([0-9\.]*[^a-z]\)/".*,\1,p' | \
    head -1
endef

define $(PKG)_BUILD
    cd '$(BUILD_DIR)' && '$(SOURCE_DIR)/configure' \
        --prefix='$(PREFIX)/$(BUILD)' 
    $(MAKE) -C '$(BUILD_DIR)' -j '$(JOBS)'
    $(MAKE) -C '$(BUILD_DIR)' -j 1 install
    ln -sf '$(PREFIX)/$(BUILD)/bin/nasm' '$(PREFIX)/bin/$(TARGET)-nasm'
endef
```

PR #1954, opened 2017-10-31
Update x264 and use nasm assembler

tonytheodore:
Replaces #1900


tonytheodore:
Direct downstream ffmpeg and gst-plugins-ugly build on all four targets.

tonytheodore:
All downstream packages (aubio chromaprint cimg ffmpeg gst-plugins-bad gst-plugins-ugly imagemagick opencv) also build successfully on all four targets.

Issue #1961, opened 2017-11-05
SDL2: SDL_opengl.h declares extension prototypes even without -DGL_GLEXT_PROTOTYPES


charlemagnelasse:
@richard42 and @rcgordon should be interested in the discussion.

I've tried to compile mupen64plus against MXE and found following problem with SDL2 in MXE:

 * https://travis-ci.org/charlemagnelasse/mupen64plus-video-glide64/jobs/297566071
 * https://travis-ci.org/charlemagnelasse/mupen64plus-video-glide64/jobs/297566072
 * https://travis-ci.org/charlemagnelasse/mupen64plus-video-rice/jobs/297566188
 * https://travis-ci.org/charlemagnelasse/mupen64plus-video-rice/jobs/297566189

Problem is that SDL_opengl.h is declaring prototypes even when GL_GLEXT_PROTOTYPES is not set. And when you set GL_GLEXT_PROTOTYPES then it declares it twice (one time in SDL_opengl.h and one time in SDL_opengl_glext.h):

    i686-w64-mingw32.shared-g++ -fvisibility-inlines-hidden -std=gnu++0x -O3 -flto -Wall -Wno-unused-function -ffast-math -fno-strict-aliasing -fvisibility=hidden -I../../src -I../../src/Glitch64/inc -DGCC -mmmx -msse -fno-PIC -I/usr/lib/mxe/usr/i686-w64-mingw32.shared/include   -I/usr/lib/mxe/usr/i686-w64-mingw32.shared/include/libpng16    -I/usr/lib/mxe/usr/i686-w64-mingw32.shared/include/SDL2 -Dmain=SDL_main -DGL_GLEXT_PROTOTYPES "-I/home/travis/build/charlemagnelasse/mupen64plus-video-glide64mk2/deps/mupen64plus-core/src/api/" -MD -MP -DNO_FILTER_THREAD -DTEXTURE_FILTER -DDUMP_CACHE  -c -o _obj/Glitch64/OGLgeometry.o ../../src/Glitch64/OGLgeometry.cpp
    In file included from ../../src/Glitch64/OGLcombiner.cpp:30:0:
    ../../src/Glitch64/glitchmain.h:77:38: error: 'void (__attribute__((__stdcall__)) * glActiveTextureARB)(GLenum)' redeclared as different kind of symbol
         extern PFNGLACTIVETEXTUREARBPROC glActiveTextureARB;
                                          ^
[...]

PR #1963, opened 2017-11-06
sdl2: disable conflicting GL extension prototypes

charlemagnelasse:
The issue for this PR is https://github.com/mxe/mxe/issues/1961

@rcgordon, @slouken would be good to hear from you guys because this should also be fixed upstream. The upstream ticket is https://bugzilla.libsdl.org/show_bug.cgi?id=3944.


charlemagnelasse:
Ah, misunderstood how patch should be created with the ./tools/patch-tool-mxe tool. I have therefore remade the patch and updated the branch with it.

tonytheodore:
Thanks! Patch consolidated with:
```
make init-git-sdl2
make import-all-patches-sdl2
make export-patch-sdl2
```

PR #1965, opened 2017-11-07
patch.mk: use numbered patch headers

tonytheodore:
Numbered patches can cause minor noise when patches change but they
can also help when reviewing a patch removal. If the first patch says:

```
Submitted upstream: https://....
Remove next update
```

then that could just as easily mean the whole file.


tonytheodore:
See https://github.com/mxe/mxe/pull/1016#issuecomment-161714267

tonytheodore:
Removed the numbered patch change to update the current patches first. I'll add it back as a final step using `make refresh-patches`.

tonytheodore:
The extra noise in diffs is mostly the initial conversion, after that the numbering changes are easy to glide past. There's another case of unnumbered patches causing potential confusion when the first patch is very long (i.e. [coin](https://github.com/mxe/mxe/commit/1118024f45c5e274bc1a82d0b14f25293302b809#diff-12d8d1129e4ddb5c43b158998206c88e)) and a small trailing patch. The extra context with numbering is well worth it.

Four packages fail with [weird line ending issues](https://github.com/mxe/mxe/pull/1965/commits/c01dc95ed113c96c27d979544ebcbc0dc2a2bd3b) but this is good to go. The remaining patches not named `*-1-fixes.patch` can be gradually converted in the future.

PR #1966, opened 2017-11-07
secure URLs

vszakats:

  * install .pc file,

Thanks!

### DIFF
--- a/src/itpp.mk
+++ b/src/itpp.mk
@@ -1,7 +1,7 @@
 # This file is part of MXE. See LICENSE.md for licensing information.
 
 PKG             := itpp
-$(PKG)_WEBSITE  := http://itpp.sourceforge.net
+$(PKG)_WEBSITE  := https://itpp.sourceforge.io/
 $(PKG)_IGNORE   :=
 $(PKG)_VERSION  := 4.3.1
 $(PKG)_CHECKSUM := 50717621c5dfb5ed22f8492f8af32b17776e6e06641dfe3a3a8f82c8d353b877
--- a/src/sdl2_gfx.mk
+++ b/src/sdl2_gfx.mk
@@ -1,18 +1,18 @@
 # This file is part of MXE. See LICENSE.md for licensing information.
 
 PKG             := sdl2_gfx
-$(PKG)_WEBSITE  := http://www.ferzkopp.net/joomla/content/view/19/14/
+$(PKG)_WEBSITE  := https://www.ferzkopp.net/joomla/content/view/19/14/
 $(PKG)_DESCR    := SDL2_gfx
 $(PKG)_IGNORE   :=
 $(PKG)_VERSION  := 1.0.3
 $(PKG)_CHECKSUM := a4066bd467c96469935a4b1fe472893393e7d74e45f95d59f69726784befd8f8
 $(PKG)_SUBDIR   := SDL2_gfx-$($(PKG)_VERSION)
 $(PKG)_FILE     := SDL2_gfx-$($(PKG)_VERSION).tar.gz
-$(PKG)_URL      := http://www.ferzkopp.net/Software/SDL2_gfx/$($(PKG)_FILE)
+$(PKG)_URL      := https://www.ferzkopp.net/Software/SDL2_gfx/$($(PKG)_FILE)
 $(PKG)_DEPS     := gcc sdl2
 
 define $(PKG)_UPDATE
-    $(WGET) -q -O- 'http://www.ferzkopp.net/joomla/content/view/19/14/' | \
+    $(WGET) -q -O- 'https://www.ferzkopp.net/joomla/content/view/19/14/' | \
     grep 'href.*tar\.' | \
     $(SED) -n 's,.*SDL2_gfx-\([0-9][^>]*\)\.tar.*,\1,p' | \
     head -1
--- a/plugins/native/glib2-macports.mk
+++ b/plugins/native/glib2-macports.mk
@@ -1,7 +1,7 @@
 # This file is part of MXE. See LICENSE.md for licensing information.
 
 PKG             := glib2-macports
-$(PKG)_WEBSITE  := packages.macports.org
+$(PKG)_WEBSITE  := https://packages.macports.org/
 $(PKG)_DESCR    := glib2 pre-built macports package
 $(PKG)_IGNORE   :=
 $(PKG)_VERSION  := 2.50.3
--- a/src/quazip.mk
+++ b/src/quazip.mk
@@ -7,11 +7,11 @@ $(PKG)_VERSION  := 0.7.3
 $(PKG)_CHECKSUM := 2ad4f354746e8260d46036cde1496c223ec79765041ea28eb920ced015e269b5
 $(PKG)_SUBDIR   := $(PKG)-$($(PKG)_VERSION)
 $(PKG)_FILE     := $(PKG)-$($(PKG)_VERSION).tar.gz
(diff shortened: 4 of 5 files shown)


src/freexl.mk
```diff
@@ -8,11 +8,11 @@ $(PKG)_VERSION  := 1.0.3
 $(PKG)_CHECKSUM := f8ed29e03a6155454e538fce621e53991a270fcee31120ded339cff2523650a8
 $(PKG)_SUBDIR   := freexl-$($(PKG)_VERSION)
 $(PKG)_FILE     := freexl-$($(PKG)_VERSION).tar.gz
-$(PKG)_URL      := http://www.gaia-gis.it/gaia-sins/freexl-sources/$($(PKG)_FILE)
+$(PKG)_URL      := https://www.gaia-gis.it/gaia-sins/freexl-sources/$($(PKG)_FILE)
 $(PKG)_DEPS     := gcc libiconv
 
 define $(PKG)_UPDATE
-    $(WGET) -q -O- 'http://www.gaia-gis.it/gaia-sins/freexl-sources/' | \
+    $(WGET) -q -O- 'https://www.gaia-gis.it/gaia-sins/freexl-sources/' | \
```
tonytheodore:
```
Forbidden

You don't have permission to access /gaia-sins/freexl-sources/ on this server.

Apache/2.2.15 (CentOS) Server at www.gaia-gis.it Port 443
```

src/itpp.mk
```diff
@@ -1,7 +1,7 @@
 # This file is part of MXE. See LICENSE.md for licensing information.
 
 PKG             := itpp
-$(PKG)_WEBSITE  := http://itpp.sourceforge.net
+$(PKG)_WEBSITE  := https://itpp.sourceforge.io/
```
tonytheodore: Just redirects the http://itpp.sourceforge.net site

src/sdl2_gfx.mk
```diff
@@ -1,18 +1,18 @@
...
 $(PKG)_CHECKSUM := a4066bd467c96469935a4b1fe472893393e7d74e45f95d59f69726784befd8f8
 $(PKG)_SUBDIR   := SDL2_gfx-$($(PKG)_VERSION)
 $(PKG)_FILE     := SDL2_gfx-$($(PKG)_VERSION).tar.gz
-$(PKG)_URL      := http://www.ferzkopp.net/Software/SDL2_gfx/$($(PKG)_FILE)
+$(PKG)_URL      := https://www.ferzkopp.net/Software/SDL2_gfx/$($(PKG)_FILE)
 $(PKG)_DEPS     := gcc sdl2
 
 define $(PKG)_UPDATE
-    $(WGET) -q -O- 'http://www.ferzkopp.net/joomla/content/view/19/14/' | \
+    $(WGET) -q -O- 'https://www.ferzkopp.net/joomla/content/view/19/14/' | \
```
tonytheodore: Redirects to http.

src/spatialite.mk
```diff
@@ -8,11 +8,11 @@ $(PKG)_VERSION  := 4.4.0-RC1
 $(PKG)_CHECKSUM := 80f7fff0a147044c5eb197e565f598ac1f137d86d0a548cbc8f52fb7ff7cac68
 $(PKG)_SUBDIR   := libspatialite-$($(PKG)_VERSION)
 $(PKG)_FILE     := libspatialite-$($(PKG)_VERSION).tar.gz
-$(PKG)_URL      := http://www.gaia-gis.it/gaia-sins/libspatialite-sources/$($(PKG)_FILE)
+$(PKG)_URL      := https://www.gaia-gis.it/gaia-sins/libspatialite-sources/$($(PKG)_FILE)
 $(PKG)_DEPS     := gcc dlfcn-win32 freexl geos libiconv libxml2 proj sqlite zlib
 
 define $(PKG)_UPDATE
-    $(WGET) -q -O- 'http://www.gaia-gis.it/gaia-sins/libspatialite-sources/' | \
+    $(WGET) -q -O- 'https://www.gaia-gis.it/gaia-sins/libspatialite-sources/' | \
```
tonytheodore:
```
Forbidden

You don't have permission to access /gaia-sins/libspatialite-sources/ on this server.

Apache/2.2.15 (CentOS) Server at www.gaia-gis.it Port 443
```

src/itpp.mk
vszakats: I know, it needs to be fixed by the project by enabling HTTPS on the SourceForge admin panel. I've made this change for all other sf homepages, so I think this should pass too.

vszakats:
I didn't know there is a new policy that doesn't allow insecure redirects. Should I close?

src/spatialite.mk
vszakats: My mistake. It's a weird server config though!

vszakats:
`gaia-gis.it` removed.

tonytheodore:
> I didn't know there is a new policy that doesn't allow insecure redirects.

More a question than a policy - is an insecure redirect giving people a false sense of security and is that worse that just linking to the plain http site?

vszakats:
For URLs used in a browser (i.e. webpage), redirections are clearly visible, so it doesn't give a false sense of security. As for URLs used inside tools, it may be a problem, if such redirection is not made visible (e.g. as an info/warning message) when those tools are used. Homebrew has also an option to disable such redirection by users, if they don't wish to allow them.

With above in mind, I believe it's better to use landing URLs that are secure, even if they are redirected to insecure, because in these cases it makes it visible that the server isn't secured correctly and maintainer be informed about it. It also means that the connection will be automatically secured (fully) as soon as the maintainer fixes its server config.

Not sure if MXE has an option to fully disable insecure transfers (no just redirects) or if it displays redirects, but they would nice to implement. With `curl`, insecure redirects for https:// URLs can be disabled by this option: `-L --proto-redir =https`.

vszakats:
(Accidental) case in point: All the URLs updated in this patch belong to misconfigured servers.

vszakats:
One more thought: I generally believe that if a server offers HTTPS even in a slightly "off" way like these, the _intent_ is usually good — with the though of security, but the execution is either incomplete or incorrect. Such issues can expected be fixed if the intent is there.

tonytheodore:
> Not sure if MXE has an option to fully disable insecure transfers (no just redirects) or if it displays redirects

These wget options don't work:
```
--https-only
--accept-regex 'https://*'
```
these do:
```
--secure-protocol=[SSlv2, SSLv3, TLSv1, TLSv1_1, TLSv1_2 or PFS]
http_proxy=foo.com
```
MXE whitelists *PROXY and *proxy env vars so you could just set those.

> (Accidental) case in point: All the URLs updated in this patch belong to misconfigured servers.

See also #1967, an outdated client may accept a certificate and allow a package to be updated but it won't subsequently make it to our backup servers so other clients will fail.

tonytheodore:
Try:
```
http_proxy=localhost:0 make download-only-a52dec REMOVE_DOWNLOAD=1 MXE_NO_BACKUP_DL=1
```
 at a quick glance, SourceForge and Gnome projects can redirect to http.

tonytheodore:
> With above in mind, I believe it's better to use landing URLs that are secure

`s/secure/encrypted` and I agree;)

vszakats:
Yeah, I stuck with secure for brevity, other better alternatives are TLS, TLS/SSL or just HTTPS. Encrypted is fine (and readable), though it’s not just encryption.

tonytheodore:
> http_proxy=localhost:0 make download-only-a52dec REMOVE_DOWNLOAD=1 MXE_NO_BACKUP_DL=1

See #1756, you can make it MXE-specific with:
```
echo 'WGET_TOOL = http_proxy=localhost:0 wget' >> settings.mk
```

and add any other options of course.

PR #1969, opened 2017-11-09
Add support for GitHub releases api

tonytheodore:
`GH_CONF` is now specified as:
```
Track branch - Tarball API
  GH_CONF := owner/repo/branch
  updates will use the last commit from the specified branch as
  a version string and bypass `sort -V`

Track releases - Release API
  GH_CONF := owner/repo/releases[/latest], tag prefix, tag suffix, tag filter-out, version separator
  updates can optionally use the latest non-prerelease with /latest
  or manually specify version numbering based on:
  <tag prefix><s/<version sep>/./version><tag suffix>

Track tags - Archive API
  GH_CONF := owner/repo/tags, tag prefix, tag suffix, tag filter-out, version separator
  updates will construct a version number based on:
  <tag prefix><s/<version sep>/./version><tag suffix>
```


tonytheodore:
All packages successful with:
```
make check-gh-conf-dl
```

tonytheodore:
Most of the remaining packages can actually just use the tags api, I've added `libass` since it specifically uses releases with `configure`, I think this is good to go.

PR #1972, opened 2017-11-09
gtk3: restore built-in icons from 3.12

### LINKED ISSUE #1968: gtk3 patch
[gtk3-2-builtinicons.patch.txt](https://github.com/mxe/mxe/files/1455140/gtk3-2-builtinicons.patch.txt)
[gtk3-test-icons.c.txt](https://github.com/mxe/mxe/files/1455142/gtk3-test-icons.c.txt)
[Attached ".txt" to file names because Github doesn't let me attach them otherwise?!?]

Hello,
please add the attached patch file to MXE.

It solves the problem of GTK+ standard icons not being available if an icon theme isn't installed on the target machine. This also means that some standard widgets (GtkSpinButton, GtkComboBoxText, possibly others) which rely on the icons aren't displayed correctly. The attached test file demonstrates the problem.

The patch restores the builtin icons that used to be provided by GTK+ until version 3.12, but were removed in the 3.14 release. It's large because it adds back a file which contains the icon pixmaps packed in a 400+ kbytes character array.

With this patch, a static executable built with MXE and copied to a Windows machine will "just work" without requiring to install also an icon theme.

I don't know git, so can you please build the pull request.

Thank you
Gerardo

tonytheodore:
http://lists.nongnu.org/archive/html/mingw-cross-env-list/2017-11/msg00000.html
closes #1968


tonytheodore:
Thanks @gballabio, I'm testing this now.

Do you have notes on how this patch was created? They will be helpful for future updates.

tonytheodore:
Looks good!

screen shot 2017-11-10 at 18 10 24

screen shot 2017-11-10 at 18 11 37

gballabio:
Trying to lay out what I did:

After trying to understand how Gtk+ draws icons and getting quite lost (I still haven't figured out the complete process), I got finally started on the right path when I found a reference in the gtk mailing list archives that there used to be a file gtkbuiltinicons.h where icons were hardcoded. So I added back the file from version 3.12, then searched where the builtin_icons array was referenced, that was in a few places in gtkicontheme.c, so I rather blindly added those lines back.
That wasn't quite enough, it did stop spewing lots of warnings, but the "missing-icon" icon was being displayed everywhere instead of the expected ones.
After some digging I found out that until 3.12, when Gtk+ was given an icon name to search, it searched also for a few variations of the name, for example if "list-add-symbolic" wasn't found it could fall back to "list-add". And indeed the builtin icons set contained only the non-symbolic variant. That logic had been dropped in 3.14, so I added that back too (in function choose_icon). And that was it!
The final piece was to add the new icon "pan-down-symbolic" that is now required by the GtkComboBoxText widget (previously it used a different mechanism to draw the arrow). That required some reverse-engineering of the builtin_icons array, I haven't fully decoded its format, but enough to be able to insert a new icon at the end of the array (I copied the icon pixmap from my Debian box). If you think that's useful I can post more details also on this.

tonytheodore:
That should be enough of an outline though it looks like it could be added back in upstream with a configure option to leave it disabled by default.

Thanks!

PR #1974, opened 2017-11-12
Makefile: try next download url on failed checksum

tonytheodore:
There are various cases where a download succeeds but the checksum
is wrong:
  - upstream re-releases tarball without new version
  - sourceforge returns 200 with a text message about outage
  - other strange glitches

In these cases, we currently fail and stop the build. This PR allows
for falling through the next URLs until it will eventually get to
the checksum-suffixed version.

This ensures builds are always (potentially) repeatable as the checksum
associated with the version is reachable. Next step would be to change
all the backups to use suffixed tarballs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,9 +17,12 @@ MXE_TARGETS        := i686-w64-mingw32.static
 
 DEFAULT_MAX_JOBS   := 6
 SOURCEFORGE_MIRROR := downloads.sourceforge.net
+MXE_MIRROR         := https://mirror.mxe.cc/pkg
 PKG_MIRROR         := https://s3.amazonaws.com/mxe-pkg
 PKG_CDN            := http://d1yihgixbnrglp.cloudfront.net
 GITLAB_BACKUP      := https://gitlab.com/starius/mxe-backup2/raw/master
+# reorder as required, ensuring final one is a http fallback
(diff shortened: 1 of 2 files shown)


tonytheodore:
@starius wrote https://github.com/mxe/mxe/pull/1681#issuecomment-280887497
> Can we change tools/s3-fetch-and-sync to do so as well (providing current filenames as well for existing files for backwards compatibility)?

Done in https://github.com/mxe/mxe/pull/1974/commits/ba358185fa71930403a55c00262bec7d82c4fc37

tonytheodore:
S3 updated with:
```
make download CREATE_SUFFIXED_ARCHIVE=true REMOVE_DOWNLOAD=true
```
this has the benefit of fetching the archives from gitlab for any packages currently failing.

tonytheodore:
See: #1899 (comment)
  - GitLab redirects to https anyway
  - final fallthrough to Cloudfront will be http
  - GitLab is (currently) the most complete so try it first

tonytheodore:
Added `mirror.mxe.cc/pkg` since the files are sitting there anyway.

tonytheodore:
> GitLab is (currently) the most complete so try it first

Just realised this doesn't actually matter, nothing prior to this PR will use the logic and GitLab was previously failing due to trailing slash.

It's easy to change order down the track with `MIRROR_SITES` variable.

PR #1960, opened 2017-11-04
[WIP] New MXE packages with focus on security

tonytheodore:
Replaces #1560

 - [ ] src/abiword.mk
 - [ ] src/asn2quickder.mk
 - [ ] src/enchant.mk
 - [x] src/ldns.mk
 - [x] src/libtasn1.mk
 - [x] src/p11-kit.mk
 - [ ] src/quick-der.mk
 - [ ] src/tlspool-gui.mk
 - [ ] src/tlspool.mk
 - [x] src/unbound.mk
 - [ ] src/wv.mk

Some will go in plugins/apps

Issue #1310, opened 2016-04-19
separate grabbing of available versions from updating to the latest version


starius:
Sometimes version tags are disordered by prefixes. In this case the list of versions is correct, but the order is wrong. I want to have the following command refactored from current updaters:

```
$ make show-versions-<package>
1.2.3
1.2.2
1.1.10
v1.3.4
```

tonytheodore:
I have some WIP that does that for github tags, there's no reason it can't be generalised:

```makefile
# common tag filtering is applied with `grep -v`:
GITHUB_TAG_FILTER := 'alpha\|beta\|rc'

# called with owner/repo, tag prefix, tag suffix, filter
define MXE_GET_GITHUB_TAGS
    $(WGET) -q -O- 'https://api.github.com/repos/$(strip $(1))/git/refs/tags/' \
    | $(SED) -n 's#.*"ref": "refs/tags/\([^"]*\).*#\1#p' \
    | $(and $(4),grep -v '$(strip 4)') \
    | $(SED) 's,^$(strip $(2)),,g' \
    | $(SED) 's,$(strip $(3))$$,,g' \
    | $(SORT) -V
endef
```

PR #1975, opened 2017-11-13
Makefile: add generic updater function

tonytheodore:
Initially used in #1960 but should be expanded to #1310


tonytheodore:
#1310 is a separate requirement, there's not much that can be done for existing updaters and the filtering in the generic one should remove the need to inspect versions manually.

PR #1977, opened 2017-11-13
Makefile: speedup LOOKUP_PKG_RULE

tonytheodore:
LOOKUP_PKG_RULE is called many times and spends a lot of time searching
for rules when the default rule is the most common. This simply
avoids exhaustive searching since we can determine beforehand if there
are multiple rules.

Also drops some of the gmsl functions if favour of normal make
variables.

First commit adds sort/strip so print-deps-for-build-pkg output can be
easily compared.

```
$ git checkout HEAD~1
$ time make print-deps-for-build-pkg \
      MXE_TARGETS="`echo {i686-w64-mingw32,x86_64-w64-mingw32}.{static,shared}`"> orig
real	0m17.564s
user	0m16.410s
sys	0m0.783s

$ git checkout speedup
$ time make print-deps-for-build-pkg \
      MXE_TARGETS="`echo {i686-w64-mingw32,x86_64-w64-mingw32}.{static,shared}`"> new
real	0m4.036s
user	0m3.624s
sys	0m0.326s

$ diff -u orig new
$
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -736,13 +736,13 @@ show-upstream-deps-%:
 # print first level pkg deps for use in build-pkg.lua
 .PHONY: print-deps-for-build-pkg
 print-deps-for-build-pkg:
-	$(foreach TARGET,$(MXE_TARGETS), \
+	$(foreach TARGET,$(sort $(MXE_TARGETS)), \
 	    $(foreach PKG,$(sort $($(TARGET)_PKGS)), \
-	        $(info for-build-pkg $(TARGET)~$(PKG) \
+	        $(info $(strip for-build-pkg $(TARGET)~$(PKG) \
 	        $(subst $(space),-,$($(PKG)_VERSION)) \
 	        $(addprefix $(TARGET)~,$(value $(call LOOKUP_PKG_RULE,$(PKG),DEPS,$(TARGET)))) \
 	        $(addprefix $(TARGET)~,$(if $(call set_is_not_member,$(PKG),$(MXE_CONF_PKGS)),$(MXE_CONF_PKGS))) \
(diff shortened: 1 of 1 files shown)


tonytheodore:
Don't really need the `print-deps-for-build-pkg` test, though the changes are harmless - we can directly compare the full debug output with:
```
make TIMESTAMP=test GIT_HEAD=test --debug=all -n --always-make
```
(don't know why this didn't occur to me earlier)
```
$ wc orig
  363419 1940070 27497348 orig
$ wc new
  363419 1940070 27497348 new
$ diff -u orig new | wc
      66     368    5428
```
Those differences are all process related:
```diff
-Putting child 0x7f80a1dd2c60 (/Users/tonyt/dev/mxe/usr/installed/.gitkeep) PID 22784 on the chain.
-Live child 0x7f80a1dd2c60 (/Users/tonyt/dev/mxe/usr/installed/.gitkeep) PID 22784 
-Reaping winning child 0x7f80a1dd2c60 PID 22784 
+Putting child 0x7fb1c5f24c30 (/Users/tonyt/dev/mxe/usr/installed/.gitkeep) PID 22967 on the chain.
+Live child 0x7fb1c5f24c30 (/Users/tonyt/dev/mxe/usr/installed/.gitkeep) PID 22967 
+Reaping winning child 0x7fb1c5f24c30 PID 22967 
```

I think it's good to go.

tonytheodore:
Oops, build-matrix doesn't use MXE_TARGETS https://github.com/mxe/mxe/commit/13a1345a161337d64cd3cd95761f2933e1b0c3a5

PR #1849, opened 2017-07-24
qt4 overlay plugin (for vtk and qwt)

muellni:
This PR is part of splitting #1527
The purpose of the qt4-overlay plugin is to create a linking variant against qt4 for packages which support either qt5 or qt4. 
This makes package `qwt_qt4` obsolete.

PR #1988, opened 2017-11-23
Added a qwt5_qt4 project that builds version 5 of qwt.

highperformancecoder:

  * install .pc file,

Thanks!

### DIFF
--- a/src/qwt5_qt4.mk
+++ b/src/qwt5_qt4.mk
@@ -1,0 +1,31 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := qwt5_qt4
+$(PKG)_WEBSITE  := https://qwt.sourceforge.io/
+$(PKG)_DESCR    := Qwt
+$(PKG)_VERSION  := 5.2.3
+$(PKG)_CHECKSUM := a23e583ae66a70f21bf2acb87624eb142a5c310088687f7fa51fa3fb8222b5e5
+$(PKG)_SUBDIR   := qwt-$($(PKG)_VERSION)
+$(PKG)_FILE     := $($(PKG)_SUBDIR).zip
+$(PKG)_URL      := https://$(SOURCEFORGE_MIRROR)/project/qwt/qwt/$($(PKG)_VERSION)/$($(PKG)_FILE)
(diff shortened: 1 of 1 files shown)


tonytheodore:
See https://github.com/mxe/mxe/blob/master/plugins/qt4/qt4-overlay.mk

You can use that or just create a similar one in a local directory that also pins the version.

tonytheodore:
See also https://github.com/mxe/mxe/tree/master/plugins/examples/qt5-freeze for a more involved example of freezing package versions.

highperformancecoder:
Hi Tony, I'm not sure what you are suggesting I do with this.

Currently, for one of my projects, I need qwt5 and qt4 libraries. It is a long term goal to port the software to newer versions of the library, but that is likely to take years, if it ever happens.

I can continue with qwt5_qt4 in my own fork, of course, I just offered the package as one of potential use to others. This is PR currently the only difference between my fork's develop branch and MXE's head.

tonytheodore:
Package `qwt_qt4` was removed in #1849 as we can't maintain many 
combinations of pinned versions. Plugins are a much better way for
people to do that locally, you could add this to the example plugins
as a way to both pin versions and alter the build slightly.

Something like (untested) `plugins/examples/qwt5_qt4/qwt_overlay.mk`:

```
PKG               := qwt
$(PKG)_VERSION    := 5.2.3
$(PKG)_CHECKSUM   := a23e583ae66a70f21bf2acb87624eb142a5c310088687f7fa51fa3fb8222b5e5
$(PKG)_DEPS       := gcc qt
$(PKG)_QT_DIR     := qt

define $(PKG)_BUILD
    $(SED) -i -e 's/target.path.*$$/target.path = $$$$PREFIX\/lib\/qwt5/' '$(SOURCE_DIR)/qwtconfig.pri'
    $(SED) -i -e 's/headers.path.*$$/headers.path = $$$$PREFIX\/include\/qwt5/' '$(SOURCE_DIR)/qwtconfig.pri'
    $(qwt_BUILD)
endef
```

tonytheodore:
Note you probably don't need to re-define the build rule as users of this plugin won't need a side-by-side install (I'm guessing that's what it does).

PR #1989, opened 2017-11-25
Improve handling of virtual packages and build messages

tonytheodore:
- add metadata to distinguish virtual package types
  - improve build order and status messages for empty packages
        print pkg type or `disabled` instead of `no-build`
  - add BUILD_DRY_RUN option and fixup make -t
  - update build-matrix rule to identify virtual pkgs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -416,21 +416,24 @@ $(PREFIX)/installed/print-git-oneline-$(GIT_HEAD): | $(PREFIX)/installed/.gitkee
 	@rm -f '$(PREFIX)/installed/print-git-oneline-'*
 	@touch '$@'
 
+# distinguish between deliberately empty rules and disabled ones
+VIRTUAL_PKG_TYPES := archive meta
+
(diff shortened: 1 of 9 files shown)


pavelvat:
Seems this packages also must be changed:
geoip-database 
nasm
waf

tonytheodore:
I'm working on cross-target deps (#516) and that will address those packages - I just wanted to get a preview as it's a fairly large change.

Does it look okay to you with the bullets to signify the package is relevant to the target but neutral about support status? There's also adding  `[{type}-pkg]` to the description and pkg-message to the tooltip for deprecated packages (e.g.blas).

pavelvat:
Yes it's look okay.

tonytheodore:
See https://htmlpreview.github.io/?https://raw.githubusercontent.com/tonytheodore/mxe/a39c938e6ece00/docs/build-matrix.html

tonytheodore:
Covered in #1990

PR #1992, opened 2017-11-26
sdl pkgs: enable shared

tonytheodore:


tonytheodore:
Tested on all four targets.

PR #1995, opened 2017-11-28
Add meson-wrapper plugin

Akaricchi:
This plugin adds a "meson-wrapper" package, which installs Meson cross files and wrapper scripts for each target, similar to "cmake-conf". This allows Meson projects to seamlessly integrate with MXE.

The wrapper scripts require Meson to be installed on the host system.

Example usage:

```sh
x86_64-w64-mingw32.static-meson sourcedir destdir
# you can now use 'meson configure', 'ninja', etc. in destdir as normal
```

### DIFF
--- a/plugins/meson-wrapper/conf/mxe-crossfile.meson.in
+++ b/plugins/meson-wrapper/conf/mxe-crossfile.meson.in
@@ -1,0 +1,23 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+[binaries]
+c = '@PREFIX@/bin/@TARGET@-gcc'
+cpp = '@PREFIX@/bin/@TARGET@-g++'
+ar = '@PREFIX@/bin/@TARGET@-ar'
+ranlib = '@PREFIX@/bin/@TARGET@-ranlib'
+ld = '@PREFIX@/bin/@TARGET@-ld'
+strip = '@PREFIX@/bin/@TARGET@-strip'
+windres = '@PREFIX@/bin/@TARGET@-windres'
+windmc = '@PREFIX@/bin/@TARGET@-windmc'
+pkgconfig = '@PREFIX@/bin/@TARGET@-pkg-config'
+# MXE forbids this
+# exe_wrapper = 'wine'
+
+[properties]
+needs_exe_wrapper = true
+
+[host_machine]
+system = 'windows'
+cpu_family = '@CPU_FAMILY@'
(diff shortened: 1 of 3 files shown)


Akaricchi:
it says on your page that you're curious about what people use MXE for, so I'll use this opportunity to shill :)

We at [Taisei Project](https://taisei-project.org/) use MXE to build the official Windows releases. We're in process of switching from CMake to Meson currently, and this is pretty much why I added this feature. Maybe you'd like to add us to [that list of yours](http://mxe.cc/#used-by). I'll be happy to link back.

tonytheodore:
Nice!

Following @starius comments on the mailing list, a plugin seems right since no packages depend on it. You probably also want to control the version rather than relying on an installed version. I'd name the plugin simply `meson` and 
  - add `python3-conf` wrapper (similar to https://github.com/tonytheodore/mxe/commit/3ee5c108c8fb206f6e66ae29fa4d24d2171769fd)
  - add `meson` package (like https://github.com/tonytheodore/mxe/pull/1960/commits/573f8d28d504068fe93cf9de98d6be19b00aa92c)

tonytheodore:
Then `meson` can install it's python scripts during native build and it's wrapper during cross build (like [`pkgconf`](https://github.com/mxe/mxe/blob/master/src/pkgconf.mk)). The main reason `cmake-conf` is separate is that `cmake` takes a long time to rebuild for simple conf changes.

Akaricchi:
I could do that, but then a dependency on the host's `ninja` (the default backend) still remains. Is that acceptable, or should I add a `ninja` package as well?

Akaricchi:
In fact, `ninja` is a hard dependency of `meson`, if I recall correctly.

GreedyMe:
I’d probably build ninja as well, it looks pretty lightweight.

Akaricchi:
Thanks for the input. I've done as you suggested and submitted another pull request (#1996). Didn't want to rebase this one, since I've changed the plugin name to `meson` too.

tonytheodore:
Thanks!

tonytheodore:
> Maybe you'd like to add us to that list of yours.

added in https://github.com/mxe/mxe/commit/cb213793042b3c60d0c9b6ce8821854f570721ea

Akaricchi:
@tonytheodore thank you, I've also linked back in https://github.com/taisei-project/taisei-website/commit/ce2534e67e00b21ba694181796a7319d591599ee

PR #1996, opened 2017-11-29
[plugin] Meson build system

Akaricchi:
This plugin adds support for the Meson build system (as well as Ninja, Meson's default backend). Cross files and wrappers are installed for each target, similar to `cmake-conf`. This allows Meson projects to seamlessly integrate with MXE.

Example usage:

```sh
make MXE_PLUGIN_DIRS=plugins/meson MXE_TARGETS=x86_64-w64-mingw32.static meson
cd $destdir
x86_64-w64-mingw32.static-meson $sourcedir
x86_64-w64-mingw32.static-meson configure -Dfoo=bar
x86_64-w64-mingw32.static-ninja
```

The host's `meson` and `ninja` executables can also be used after the initial configuration step, given their versions are compatible.

Python 3.4 or later is required to be installed on the host system. A `python3-conf` package, based on @tonytheodore's [`python-conf`](https://github.com/tonytheodore/mxe/commit/3ee5c108c8fb206f6e66ae29fa4d24d2171769fd), has been added to `src` to make this work.

This PR supersedes #1995.

### DIFF
--- a/plugins/meson/meson.mk
+++ b/plugins/meson/meson.mk
@@ -46,7 +46,26 @@ endef
 
 define $(PKG)_BUILD_$(BUILD)
     cd '$(SOURCE_DIR)' && python3 setup.py install \
-        --prefix='$(PREFIX)/$(TARGET)' && \
-    $(SED) -i 's,^#!/usr/bin/python3,#!/usr/bin/env python3,' \
-        '$(PREFIX)/$(TARGET)/bin/meson'
+        --prefix='$(PREFIX)/$(TARGET)'
+
+    # Awful hacks: we must hijack the python entry points here to install our
+    # site-packages path. This is because Meson is going to put the path to the
+    # real python interpreter and script it's been invoked with into the build
+    # rules file, bypassing any of our shell wrappers. This causes automatic
+    # reconfiguration to fail.
+
+    for prog in meson{,conf,introspect,test} wraptool; do \
+        $(SED) '1d' '$(PREFIX)/$(TARGET)'/bin/$${prog} > '$(1)'/$${prog}.tail; \
+        echo "#!/usr/bin/env python3" > '$(PREFIX)/$(TARGET)'/bin/$${prog}; \
+        echo "__mxe_python_path = r'''" >> '$(PREFIX)/$(TARGET)'/bin/$${prog}; \
+        echo '$(PREFIX)/$(TARGET)/lib/python$(PY3_XY_VER)/site-packages' >> '$(PREFIX)/$(TARGET)'/bin/$${prog}; \
+        echo "'''[1:-1]" >> '$(PREFIX)/$(TARGET)'/bin/$${prog}; \
+        echo 'import sys; sys.path.insert(1, __mxe_python_path)' >> '$(PREFIX)/$(TARGET)'/bin/$${prog}; \
+        echo "__mxe_path = r'''" >> '$(PREFIX)/$(TARGET)'/bin/$${prog}; \
+        echo '$(PREFIX)/$(BUILD)/bin:$(PREFIX)/bin' >> '$(PREFIX)/$(TARGET)'/bin/$${prog}; \
+        echo "'''[1:-1]" >> '$(PREFIX)/$(TARGET)'/bin/$${prog}; \
+        echo 'import os; os.environ["PATH"] = "{0}:{1}".format(__mxe_path, os.environ["PATH"])' >> '$(PREFIX)/$(TARGET)'/bin/$${prog}; \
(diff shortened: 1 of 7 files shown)


Akaricchi:
Ran into some unexpected problems, this currently doesn't work as intended. I'll reopen the PR once I've sorted everything out, sorry.

Akaricchi:
Looks like I managed to fix it but, well, sacrifices were made.

plugins/meson/meson.mk
```diff
@@ -46,7 +46,26 @@ endef
...
-        --prefix='$(PREFIX)/$(TARGET)' && \
-    $(SED) -i 's,^#!/usr/bin/python3,#!/usr/bin/env python3,' \
-        '$(PREFIX)/$(TARGET)/bin/meson'
+        --prefix='$(PREFIX)/$(TARGET)'
+
+    # Awful hacks: we must hijack the python entry points here to install our
+    # site-packages path. This is because Meson is going to put the path to the
+    # real python interpreter and script it's been invoked with into the build
+    # rules file, bypassing any of our shell wrappers. This causes automatic
+    # reconfiguration to fail.
```
tonytheodore: What happens if you use`#!/$(PREFIX)/$(BUILD)/bin/python3` in the meson wrapper? And/or copy the python executable instead of the wrapper setting `PYTHONPATH`?

Akaricchi: The shebang line doesn't matter, Meson always puts the path to the real python executable into `build.ninja`. Copying the python3 executable also does not accomplish anything (and it's an even nastier hack in my opinion). While it would cause it to put "our" python3 into the generated build rules, `sys.path` would still not be set up correctly, and import of `mesonbuild` would fail unless it's available on the host system... which defeats the whole point of packaging meson in the first place. And yes, I have tried copying it. `sys.path` doesn't change.

The only way I see to fix this properly without patching Meson is to also package python3 properly. But even then we might still need some PATH hack to ensure that Meson finds the correct `ninja` executable.

Alternatively, we can just give up on this and go with my original thin wrapper for the native Meson.

This kind of set up with native python interpreted + bundled python packages is kind of volatile, anyway. Everything will probably fall apart if the user upgrades their python version, for example (without also rebuilding the relevant MXE packages).

Akaricchi: @tonytheodore so, what do you think about this?

tonytheodore: You're right that anyone using this plugin will already have `meson` and `ninja` installed, so a thin wrapper is probably the best option for now.

I had a quick look at building the `meson` branch of `taisei` and it looks like a very nice build system.

P.S. the [venv](https://docs.python.org/3/library/venv.html#module-venv) python module works better than setting PYTHONPATH for this, but doesn't help with the `meson` paths.

Akaricchi: @tonytheodore Alright, I'm going to reopen #1995 then. Please tell me if there are any issues remaining with it.

I also thought of trying to make use of venv, but unfortunately, the way it works requires setting up some environment variables in its `bin/activate` shell script, so that's not very useful here either. Maybe I should try contributing a way to override the `meson` invocation command upstream, so that we (and whoever else) can have a clean wrapper without patches.

tonytheodore: venv works fine without activate, [from docs](https://docs.python.org/3/library/venv.html#venv-def):
> There should be no need in other circumstances to activate a virtual environment

Meson's [invocation of itself](https://github.com/mesonbuild/meson/blob/master/mesonbuild/environment.py#L327-L331) seems to change a lot since the last release so a solution upstream would be great.

plugins/meson/meson.mk
```diff
@@ -0,0 +1,71 @@
...
+$(PKG)_DESCR     := An open source build system meant to be extremely fast and as user friendly as possible.
+$(PKG)_IGNORE    :=
+$(PKG)_VERSION   := 0.43.0
+$(PKG)_CHECKSUM  := 324894427dcd29f6156fe06b046c6ad1b998470714debd7c5705902f21aaaa73
+$(PKG)_GH_CONF   := mesonbuild/meson/releases/latest
+$(PKG)_SUBDIR    := $(PKG)-$($(PKG)_VERSION)
+$(PKG)_FILE      := $($(PKG)_SUBDIR).tar.gz
+$(PKG)_URL       := https://github.com/mesonbuild/meson/archive/$($(PKG)_VERSION).tar.gz
+$(PKG)_FILE_DEPS := $(wildcard $(PWD)/plugins/meson/conf/*)
+$(PKG)_DEPS      := python3-conf ninja
```
jonaski: Missing cmake-conf

PR #2000, opened 2017-12-06
hyperscan: fix build where native SSSE3 detection fails

### LINKED ISSUE #1997: Hyperscan build is failing when making qt5base-dev
The hyperscan make fails in a recently installed MXE, for both Windows build targets.
I see that there is a recent patch for something in this area, but have no idea how to install it, or if it is relevant to this issue.

My underlying issue is the failure of Cmake to  understand qt5_wrap_ui, and I have the impression that installing qt5base-dev might be the solution to this.

tonytheodore:
closes #1997

### DIFF
--- a/src/hyperscan.mk
+++ b/src/hyperscan.mk
@@ -10,12 +10,22 @@ $(PKG)_GH_CONF  := 01org/hyperscan/tags, v
 $(PKG)_DEPS     := gcc boost
 # $(PKG)_NATIVE_DEPS := ragel
 
+# Add the following options to run on (virtual) machine without AVX2 or
+# build on machine where native detection of SSSE3 may fail
+# -DCMAKE_C_FLAGS="-march=core2" -DCMAKE_CXX_FLAGS="-march=core2"
+
+$(PKG)_ARCH_FLAGS = \
+    $(TARGET)-gcc -xc /dev/null -o- -S -fverbose-asm \
+        -march=native | \
+        grep mssse3 >/dev/null 2>&1 || \
+        echo -march=core2
+
 define $(PKG)_BUILD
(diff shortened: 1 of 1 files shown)


Cenkron:
It seems to have had no effect.  Let me verify I did the right thing for this test:
In my mxe root directory, I did a "git pull" followed by a "make", and retrieved the log which was properly dated, so I know it retried the build.

[hyperscan_i686-w64-mingw32.static.posix.txt](https://github.com/mxe/mxe/files/1535438/hyperscan_i686-w64-mingw32.static.posix.txt)

tonytheodore:
It isn't merged yet so you need to do something like:
```
git checkout -b tonytheodore-hyperscan-arch master
git pull https://github.com/tonytheodore/mxe.git hyperscan-arch
```
and then `make` again. The commands above are usually copy-and-paste ready in the `command line instructions` on every PR.

Cenkron:
I tried to do this, but it is asking for your username and password.

Cenkron:
I was revisiting this, and I realize I have no idea where the command line instructions you mentioned are.

tonytheodore:
Sorry, the instructions are near the merge button which is only visible to committers.

You haven't set up [any ssh keys](https://github.com/Cenkron.keys) so the login details are just your normal credentials you use for github.

You could try:
```
curl -L https://github.com/mxe/mxe/pull/2000.patch | git am
#or
curl -L https://github.com/mxe/mxe/pull/2000.diff | patch -p1
```
in your new branch.

Cenkron:
I finally got back to where I could test your patch.  It was successful; hyperscan is now building properly.
Thanks.

PR #2010, opened 2017-12-16
Add libao.

toots:


tonytheodore:
Merged with some changes to fix shared build, add test, use GH_CONF, and build out-of-source https://github.com/mxe/mxe/commit/a75d79da1e3c522f5faa74e01e24219f60d7d3a1.

Thanks!

PR #2012, opened 2017-12-18
qtbase: add Qt's angle backend

manner82:
Hi,

As a hackathon project I have tried to crosscompile Qt with angle support.

The biggest disadventage I have stumbled across is that it seems to need fxc.exe from microsoft directx sdk (now windows sdk). This tool is used to generate some header files containing the compiled shaders as bytes. I did not found an opensource alternative (do you know any?), so I guess we could either
- run fxc.exe through wine... This has the disadventage that mxe users needs to install it manually, I guess it is not legal to redistribute it in tar.gz format...
- I thought it would be a better option to provide some simple script to generate the headers it would create, and add them as patch, so during the build no fxc.exe would be needed.

Additionally it seems to need opengl32.dll from mesa to work with software rasterizer, so I have added a target for that. It would be much better if it could use d3d, but that still needs some magic.

I have also changed "-opengl desktop" to "-opengl dynamic" which results in that it tries to run the opengl backend and if fails, tries the angle one, and finally falls back to software mode.

Do you think this could/should be part of mxe in some way? Any feedback is welcome, I am no expert on this.

### DIFF
--- a/src/mesa-1-fixes.patch
+++ b/src/mesa-1-fixes.patch
@@ -1,0 +1,25 @@
+This file is part of MXE. See LICENSE.md for licensing information.
+
+Contains ad hoc patches for cross building.
+
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Robert Manner <robert.manner@balabit.com>
+Date: Fri, 15 Dec 2017 16:40:18 +0100
+Subject: [PATCH] scons/crossmingw.py: compiler prefix can be customized
+ through environment
+
+
+diff --git a/scons/crossmingw.py b/scons/crossmingw.py
+index 1111111..2222222 100644
+--- a/scons/crossmingw.py
++++ b/scons/crossmingw.py
+@@ -51,6 +51,9 @@ prefixes64 = SCons.Util.Split("""
+ """)
+ 
+ def find(env):
++    if os.environ['MINGW_PREFIX']:
++        return os.environ['MINGW_PREFIX']
++
+     if env['machine'] == 'x86_64':
+         prefixes = prefixes64
+     else:
--- a/src/mesa.mk
+++ b/src/mesa.mk
@@ -1,0 +1,13 @@
+PKG             := mesa
+$(PKG)_VERSION  := 17.3.0
+$(PKG)_CHECKSUM := 29a0a3a6c39990d491a1a58ed5c692e596b3bfc6c01d0b45e0b787116c50c6d9
+$(PKG)_SUBDIR   := mesa-$($(PKG)_VERSION)
+$(PKG)_FILE     := mesa-$($(PKG)_VERSION).tar.xz
+$(PKG)_URL      := ftp://ftp.freedesktop.org/pub/mesa/$($(PKG)_FILE)
+$(PKG)_DEPS     := gcc
+
+define $(PKG)_BUILD
+    cd '$(1)' && \
+    MINGW_PREFIX='$(TARGET)-' scons platform=windows toolchain=crossmingw machine=x86_64 verbose=1 build=release libgl-gdi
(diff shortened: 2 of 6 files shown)


mardy:
Hi Róbert, do you have an update on this? I'm trying your branch on a newer version of MXE (with Qt 5.14.2) and the patches to Qt's configure script don't apply anymore. I can work on it, but first I wanted to know if you happen to have them ready :-)

manner82:
> Hi Róbert, do you have an update on this? I'm trying your branch on a newer version of MXE (with Qt 5.14.2) and the patches to Qt's configure script don't apply anymore. I can work on it, but first I wanted to know if you happen to have them ready :-)

Yes and no :) We have abandoned this way, it turned out not quite maintainable, at least the header generation and its script.
We chose instead to call fxc during the normal Qt build. I can upload my patches, but it is based on Qt 5.13, I might try to rebase however.

How it works and why didn't I share:

I have copied fxc.exe from a Windows (it is part of winSDK) to the build "machine". You only need the exe (and no .dlls), it is around ~150KB, so would be easy for mxe to fetch or include, nice. Then I have created a wrapper script named "fxc" which calls 'wine /the/real/fxc.exe "$@"' or something similar, so the Qt build can use it. Unfortunately, I do not think that redistributing fxc.exe is legally ok.

All hope is not lost however: there is a project dxc[1], which is also from microsoft and is the new fxc and is opensource and is able to compile on linux (that part is by google).

I have tried to use it instead, needs some argument adjustments related to shader formats(?). But I was not able to do a proper compile (yet). Also Qt 5.14 had some changes that might affect this[2], they have replaced the opengl es backend with some more general, so might be "angle" is not needed any more? I am not sure, and also do not really have time in foreseeable future to go on with this, sorry.

Hope these points help a bit anyway, and I would like a lot if you'd keep me updated if you progress.

[1] Here is the repo: https://github.com/microsoft/DirectXShaderCompiler, or my packaging for archlinux: https://aur.archlinux.org/packages/?O=0&SeB=nd&K=directx-shader-compiler&outdated=&SB=n&SO=a&PP=50&do_Search=Go
[2] https://www.qt.io/blog/qt-quick-on-vulkan-metal-direct3d

manner82:
The patches seemed to have no conflict, so I have uploaded them on qt514-angle branch[1] (they are no secret). But to repeat: I do not know if they still work/compile for 5.14, and more importantly, if they are still needed at all, because 5.14 might have a better solution for the problem.

[1] https://github.com/manner82/mxe/tree/qt514-angle

mardy:
Thanks Róbert for the extensive reply!

> Also Qt 5.14 had some changes that might affect this[2], they have replaced the opengl es backend with some more general, so might be "angle" is not needed any more?

I think I made some progress, thanks to your hint here. I simply changed `--opengl=desktop` to `--opengl=software` in the `qtbase/mk` file (without also enabling `--angle`), and the build failed because `EGL/egl.h` was not found. So I added the mesa package (by mostly reusing your recipe) and make sure it installed all the headers needed by Qt into the MXE include directory, and that was enough to get qtbase to build. I'm a bit surprised that no `.lib` files were needed.

Then I rebuilt my application, and it worked exactly as before: it crashes because the opengl32.dll provided by microsoft only provides OpenGL 1.1, and it worked if I added the `--device=software` command-line option (this is equivalent to setting `QT_OPENGL=software` in the environment). So, no improvement and no regressions.

But! :-) Copying the binary files `libEGL.dll` + `libGLESv2.dll` (these two work as a pair) or opengl32sw.dll from a Qt Windows installation into my application's folder made my application work without the need to specify any parameter. By running it with `QT_LOGGING_RULES=qt.qpa.gl=true` I can see that it first tries to use the system's openGL libraries, then falls back to EGL, and finally (if that fails too) to opengl32sw.dll.

I'll now clean up my changes a bit and push them in a branch, then I guess we can discuss over the MR.

mardy:
Here it is: https://github.com/mxe/mxe/pull/2480

PR #2016, opened 2017-12-23
Enable high-quality resampling for SDL2

Akaricchi:
This patch enables `libsamplerate` support in SDL2. SDL's internal resampling is known to be bad and cause a lot of distortion. I've received many complaints about it from my users.

`sdl2_mixer` had to be updated to version `2.0.2`. The old version does not use the new `SDL_AudioStream` API, which is required to make use of high-quality resampling.


tonytheodore:
Thanks!

PR #2023, opened 2017-12-30
Update source files fot qt5, and update qt4 overlays.

saiarcot895:
This patch makes openscenegraph build with qt5 (to match what some other libraries such as poppler are building with), and updates the qt4 overlay to include poppler and openscenegraph in that list. The qt5 build has been tested on the x86_64-w64-mingw32.static target.


tonytheodore:
Thanks!

PR #2028, opened 2018-01-01
add package cmark

mbunkus:


mbunkus:
Please note that I have not followed all the steps. I did start out by cloning a script that should be good, though: `src/libical.mk`.

  * ✓ install .pc file,
  * ✗ install bin/test-pkg.exe compiled with flags by pkg-config, — I don't have a meaningful test case to compile and install (but see below)
  * ✓ install .dll to bin/ and .a, .dll.a to lib/,
  * ✓ use $(TARGET)-cmake instead of cmake,
  * ✓ build in `$(BUILD_DIR)` instead of `$(SOURCE_DIR)`,
  * ✓ install .dll to bin/ and .a, .dll.a to lib/,
  * ✓ use $(TARGET)-cmake instead of cmake,
  * ✓ build in `$(BUILD_DIR)` instead of `$(SOURCE_DIR)`,
  * ✓ do not run target executables with Wine,
  * ✓ do not download anything while building,
  * ✓ do not install documentation,
  * ✗ do not install .exe files except test and build systems, — installs `cmark.exe`. Note that some people may even use that `cmark.exe`, and it might fill in as a replacement for the aforementioned test case.
  * ✗ and .patch files are generated by tools/patch-tool-mxe. — The first patch was generated that way, but for the second one I've just used `git show` for the relevant upstream commit and saved that as a patch file. That works nicely.

mabrand:
Pushed as 3a073ff7e3dfd6d5a6ed3d528d33597cbe6f331d

PR #2033, opened 2018-01-09
Update lame to v3.100

darealshinji:

  * install .pc file,

Thanks!


tonytheodore:
Is the `lame-1-fix-xmmintrin-errors.patch` still required? It touches `configure.in` so a `autoreconf -fi` will be needed.

darealshinji:
It didn't want to build because of GTK issues, `lame-1-fix-xmmintrin-errors.patch` fixed that. I will re-add the `autoreconf -i`, dunno why it worked without it.

tonytheodore:
> I will re-add the autoreconf -i, dunno why it worked without it.

It probably detected the files had changed and re-generated what it 
needed automatically. We need the explicit invocation to determine 
which packages use autotools for the systems we build them on (jessie, 
trusty, wheezy). `grep` is currently used for this, though we'll
probably add new metadata at some stage to determine the build system.

I added a fix for [out-of-source autoreconf](https://github.com/mxe/mxe/commit/4cdef2162b1fe0661f0ee5ae511ee32a4b09dc60) and [consolidated the
patches](https://github.com/mxe/mxe/commit/639238144bd369b5a4793cae40494a40621a3593).

Thanks!

PR #2036, opened 2018-01-13
sdl2: remove obsolete patch

Akaricchi:
The patch in question is supposed to disable dinput support, but in fact, it doesn't seem to do anything anymore. According to the logs, SDL2 successfully builds with dinput support, with or without the patch, and it works correctly too.


tonytheodore:
Makes sense, we don't use `*-pc-mingw32` anymore - thanks!

PR #2038, opened 2018-01-13
Fix poppler Qt5 bindings

rgcjonas:
Make poppler's Qt5 bindings compile by adding the Qt5 tools to $PATH, and use C++11 since the Qt5 header files require it now.


tonytheodore:
Thanks!

PR #2040, opened 2018-01-21
Fix devil shared lib compilation

ashdnazg:
Straightforward case fix.

Also sent upstream here: https://github.com/DentonW/DevIL/pull/71 but it doesn't seem the most active, so I've figured a fix here can help too :)


tonytheodore:
Thanks!

Added patch link so it's easier to tell if it's still required in future updates: https://github.com/mxe/mxe/commit/2218926bee1744d4a93e02a34870d585f8533b99

PR #2043, opened 2018-01-29
update boost to 1.66.0

fiesh:
This updates boost to 1.66.0.  Due to API changes in `boost::context`, the test had to be changed completely.  I also cleaned it up marginally.

### DIFF
--- a/src/qtbase-1-fixes.patch
+++ b/src/qtbase-1-fixes.patch
@@ -221,3 +221,240 @@ index 4dd9e9603b..c94d64d9aa 100644
 -- 
 2.30.2
 
+commit 696ff656d0ec521653a0095db09fd87c859212bb
+Author: Martchus <martchus@gmx.net>
+Date:   Sat Jun 26 22:24:12 2021 +0200
+
+    Fix crashes in rasterization code using setjmp
+    
+    * Use C++ exceptions instead of setjmp to workaround crashes
+    * The setjmp/longjmp code crashes when compiling for x86_64-w64-mingw32
+      with GCC 11 and optimizations enabled¹. It crashes when jumping back
+      to handle the case of insufficient memory. This change uses C++
+      exceptions instead (turning the compile unit into a C++ unit instead
+      of just using C) which should behave identical but don't seem to
+      crash.
+    * Fix rendering certain SVGs and possibly other graphics
+    
+    Change-Id: I01937d13569dd01ab4cb1f608020544c93bc343c
+    
+    ---
+    
+    ¹ See https://bugreports.qt.io/browse/QTBUG-94692 for details.
+
+diff --git a/src/gui/painting/painting.pri b/src/gui/painting/painting.pri
+index c68e8d6e9a..a144ce9b33 100644
+--- a/src/gui/painting/painting.pri
++++ b/src/gui/painting/painting.pri
+@@ -1,5 +1,7 @@
+ # Qt gui library, paint module
+ 
++CONFIG += exceptions
++
+ HEADERS += \
+         painting/qbackingstore.h \
+         painting/qbezier_p.h \
+@@ -80,7 +82,7 @@ SOURCES += \
+         painting/qcosmeticstroker.cpp \
+         painting/qdrawhelper.cpp \
+         painting/qemulationpaintengine.cpp \
+-        painting/qgrayraster.c \
++        painting/qgrayraster.cpp \
+         painting/qicc.cpp \
+         painting/qimagescale.cpp \
+         painting/qmatrix.cpp \
+diff --git a/src/gui/painting/qgrayraster.c b/src/gui/painting/qgrayraster.cpp
+similarity index 98%
+rename from src/gui/painting/qgrayraster.c
+rename to src/gui/painting/qgrayraster.cpp
+index 0143e9b602..803a676e1a 100644
+--- a/src/gui/painting/qgrayraster.c
++++ b/src/gui/painting/qgrayraster.cpp
+@@ -39,7 +39,7 @@
+ 
+ /***************************************************************************/
+ /*                                                                         */
+-/*  qgrayraster.c, derived from ftgrays.c                                  */
++/*  qgrayraster.cpp, derived from ftgrays.c                                */
+ /*                                                                         */
+ /*    A new `perfect' anti-aliasing renderer (body).                       */
+ /*                                                                         */
+@@ -157,17 +157,12 @@
+ #  include <vxWorksCommon.h>    /* needed for setjmp.h */
+ #endif
+ #include <string.h>             /* for qt_ft_memcpy() */
+-#include <setjmp.h>
+ #include <limits.h>
+ 
+ #define QT_FT_UINT_MAX  UINT_MAX
+ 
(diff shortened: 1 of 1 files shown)


WSoptics:
Also I realized there was a bug in `boost::thread` that kept our software from compiling.  Filed a pull request there and added the patch to the commit above (and rebased).

tonytheodore:
Thanks, I'm testing this alongside the openssl1.1 update in #2057 and there are a few packages that don't build with boost 1.66 - `libtorrent-rasterbar`, `pcl`, and `wt`. Do you know much about these packages? Updating to the latest versions doesn't fix any of them, downgrading boost to 1.65.1 works for static targets but not shared.

src/boost.mk
```diff
@@ -67,11 +67,11 @@ define $(PKG)_BUILD
     echo 'set(Boost_THREADAPI "win32")' > '$(CMAKE_TOOLCHAIN_DIR)/$(PKG).cmake'
 
     '$(TARGET)-g++' \
-        -W -Wall -Werror -ansi -U__STRICT_ANSI__ -pedantic \
+        -W -Wall -Werror -ansi -std=c++11 -U__STRICT_ANSI__ -pedantic \
         '$(PWD)/src/$(PKG)-test.cpp' -o '$(PREFIX)/$(TARGET)/bin/test-boost.exe' \
```
tonytheodore: What version of gcc are you using? I had to also add `-DCMAKE_CXX_STANDARD=11` to the cmake test below with gcc5.

fiesh: Oh that makes sense.  I was using gcc7, and I think it is fine with gcc6 as well since from gcc6 on gcc defaults to C++14.  But for gcc5, it would have to be explicitly specified.

fiesh:
Alas I have no idea about `libtorrent-rasterbar`, `pcl`, or `wt`.  Do they not compile with boost 1.66 in general, or is it MXE specific?

tonytheodore:
I think it's MXE specific, or more a matter of gcc version and `posix` threads. I'll test some different toolchains.

tonytheodore:
`libtorrent-rasterbar` succeeds with gcc7

tonytheodore:
`pcl` and `wt` remain broken with updates and different toolchains. I've merged a minor update to fix openssl with the current boost (#2057), the patch can simply be deleted for this PR.

PsychoXIVI:
Since there are no still merge, I tried to make use of it just for me. 

I tried to use 1.68.0 version. To make update available, I replaced package download URL with
```make
$(PKG)_URL      := https://dl.bintray.com/boostorg/release/$($(PKG)_VERSION)/source/$($(PKG)_FILE)
```
and update rule macro with
```make
define $(PKG)_UPDATE
	$(WGET) -q -O- 'https://dl.bintray.com/boostorg/release/' | \
	grep -o '[0-9./]*' | \
	grep -o '[0-9.]*' | \
	$(SORT) --version-sort | \
    tail -1
endef
```

Also, I replaced 
```cpp
#include <boost/context/all.hpp> 
```
with 
```cpp
#include <boost/context/continuation.hpp> 
```
in the test since in newer version of boost there are no aggregate header in context.

After that, there are sill weird problem while compiling the test. I don't know what I am missing ;_;
```
[ 50%] Building CXX object CMakeFiles/test-boost-cmake.dir/mnt/e/Libraries/MXE/mxe/src/boost-test.cpp.obj
[100%] Linking CXX executable test-boost-cmake.exe
/mnt/e/Libraries/MXE/mxe/usr/bin/x86_64-w64-mingw32.static.posix-ld: CMakeFiles/test-boost-cmake.dir/objects.a(boost-test.cpp.obj):boost-test.cpp:(.rdata$.refptr._ZN5boost6system6detail25generic_category_instanceE[.refptr._ZN5boost6system6detail25generic_category_instanceE]+0x0): undefined reference to `boost::system::detail::generic_category_instance'
/mnt/e/Libraries/MXE/mxe/usr/bin/x86_64-w64-mingw32.static.posix-ld: /mnt/e/Libraries/MXE/mxe/usr/lib/gcc/x86_64-w64-mingw32.static.posix/7.3.0/../../../../x86_64-w64-mingw32.static.posix/lib/../lib/libboost_thread-mt.a(thread.o):thread.cpp:(.text+0x2f46): undefined reference to `boost::system::detail::generic_category_ncx()'
/mnt/e/Libraries/MXE/mxe/usr/bin/x86_64-w64-mingw32.static.posix-ld: /mnt/e/Libraries/MXE/mxe/usr/lib/gcc/x86_64-w64-mingw32.static.posix/7.3.0/../../../../x86_64-w64-mingw32.static.posix/lib/../lib/libboost_thread-mt.a(thread.o):thread.cpp:(.text$_ZN5boost21thread_resource_errorC1Ev[_ZN5boost21thread_resource_errorC1Ev]+0xa): undefined reference to `boost::system::detail::generic_category_ncx()'
CMakeFiles/test-boost-cmake.dir/build.make:92: recipe for target 'test-boost-cmake.exe' failed
make[4]: *** [test-boost-cmake.exe] Error 1
```
Its look like there are "undefined reference to `boost::system::detail::generic_category_" things, but... The library needed is included in build command (`-lboost_system-t`), file with that name exist; there are the symbols for them in strings from `.a` files; 

I checked it multiply times, no progress. Any idea what might be wrong?

tonytheodore:
CMake's FindBoost has a lot of [version specific code tests](https://github.com/Kitware/CMake/blob/v3.10.0/Modules/FindBoost.cmake#L783-L803) and a sense of [known versions](https://github.com/Kitware/CMake/blob/v3.10.0/Modules/FindBoost.cmake#L783-L803). I'd start with updating cmake since those boost versions stop at 1.65.1 for our current cmake version (3.10.0).

PsychoXIVI:
Well, after day of trying to do that well and playing games because of desperation, I again read Google carefully, and its looks like the libraries linking order is worng. ( https://stackoverflow.com/questions/15280882/why-undefined-reference-to-boostsystemgeneric-category-even-if-i-do-link). 
Moving `boost_system` still not resolved all problems.

Also, found: https://github.com/boostorg/system/issues/26 and I realized that there can be similar issue - I use g++7 plugin, where C++14 is default C++ standard, I believe. 

So finally, I managed to update this pull request to work with Boost! <3

How about adding to MXE ability to specify C++ standard to use for building libraries that can vary on it? 

Should I post here changes or create another pull request? @fiesh ?

tonytheodore:
> How about adding to MXE ability to specify C++ standard

It's an [ongoing work](https://github.com/mxe/mxe/labels/package%20customization), check the [wiki](https://github.com/mxe/mxe/wiki/Customisable-Packages) in the coming weeks as I have some wip that is groundwork for some of it.

> Should I post here changes or create another pull request?

These can't be merged till all downstream packages and various toolchains are confirmed working. I think we'll end up using plugins for boost along the lines of:
```
plugins/
  boost/
    boost166/
      boost166.mk
      boost166-test.cpp
      boost166-1-fixes.patch
    boost168/
      boost168.mk
      boost168-test.cpp
      boost168-1-fixes.patch

```

a separate PR would be easier to incorporate that way.

orgads:
Can be closed.

PR #2044, opened 2018-01-29
update OpenSceneGraph to 3.5.9

WSoptics:
The 3.4 line of OpenSceneGraph is rather outdated, this updates OpenSceneGraph to 3.5.9.

### DIFF
--- a/src/openscenegraph-1-fixes.patch
+++ b/src/openscenegraph-1-fixes.patch
@@ -62,20 +62,6 @@ Date: Sun, 10 Jan 2016 14:19:57 +1100
 Subject: [PATCH 3/6] use pkg-config for gstreamer detection in mingw
 
 
-diff --git a/CMakeModules/FindGStreamer.cmake b/CMakeModules/FindGStreamer.cmake
-index 1111111..2222222 100644
---- a/CMakeModules/FindGStreamer.cmake
-+++ b/CMakeModules/FindGStreamer.cmake
-@@ -65,7 +65,7 @@
- #    )
- #endmacro()
- 
--if (WIN32)
-+if (WIN32 AND NOT MINGW)
-     macro(FIND_GSTREAMER_COMPONENT _component_prefix _pkgconfig_name _header _library)
-         find_path(${_component_prefix}_INCLUDE_DIRS
-             NAMES ${_header}
-
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Boris Nagaev <bnagaev@gmail.com>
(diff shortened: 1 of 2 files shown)


WSoptics:
Hold on, I just realized it omits `osgQt`...

WSoptics:
Actually, that seems to be because there is no `osgQt` any more.  I'll see why that is.

WSoptics:
Indeed it's been moved to a separate project:

https://github.com/openscenegraph/osgQt

I actually wonder what one needs it for, we use both `osg` and `Qt` rather heavily and have no dependency on it.

So I suppose we can remove the dependency on `Qt` from `osg` altogether?

tehnick:
Personally I do not think that updating to unstable version is a good idea.
Maintainers from Debian and Ubuntu also use latest stable release (3.4.1).

After 4617731acebcdc41ecd4c80cc0084445e0669d86 GH_CONF should be used in this package.

fiesh:
@tehnick The problem is that the official release schedule is lagging a lot now with OpenSceneGraph, 3.4 to 3.4.1 took two years, and 3.5 is a rather stable branch.  There are also no compatibility issues between the two as far as I can tell.

tehnick:
All end-user software which I saw is built with stable version of library.
(But I have not checked if they have changed API or not.)
Probably it is better to ask for comments from OSG developers.

tehnick:
> All end-user software which I saw is built with stable version of library.

For GNU/Linux and MS Windows.

tonytheodore:
A couple of threads from OSG:
  - They [don't use the `releases` API in github](https://github.com/openscenegraph/OpenSceneGraph/issues/419) so even numbers will have to be filtered in GH_CONF (third parameter)
  - [API changes](https://github.com/openscenegraph/OpenSceneGraph/issues/239) are summarised with the delta between tags. For example, [ostQt was removed in 3.5.5](https://abi-laboratory.pro/tracker/objects_report/openscenegraph/3.5.4/3.5.5/report.html)

For MXE it's more appropriate to stick to releases but we want to make it easy for people to use newer versions. If the patches were all sent upstream, that could be as simple as locally overriding version and checksum - that also has the benefit that MXE can be updated immediately when there's a new release.

WSoptics:
Two days ago, OSG 3.6.0 was release, a new stable version.  This makes this pull request obsolete, so I'll close it.

tonytheodore:
Thanks for the update, are you working on a PR for the new release? If not, I'll look at getting the patches upstream and updating.

fiesh:
That would be great, it would be a while before I can get around to doing this.

PR #2046, opened 2018-02-01
Add Psi+ to the list of projects which use MXE

tehnick:
Proof:
https://sourceforge.net/projects/psiplus/files/MS-Windows/Personal-Builds/tehnick/
https://github.com/psi-plus/maintenance/tree/master/scripts/win32/cross-compilation-examples

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -3051,6 +3051,9 @@ endef</pre>
         <a href="https://www.pokerth.net/">PokerTH</a>
     </li>
     <li>
+        <a href="https://psi-plus.com/">Psi+</a>
+    </li>
+    <li>
         <a href="https://pushover.sourceforge.io/">Pushover</a>
     </li>
     <li>


tehnick:
Rebased against master.

@mabrand Ping?

starius:
Thanks!

Few notes. The first link does not mention MXE. Scripts from the second link seem to use a fork of MXE (https://github.com/sibuserv/mxe). Can they be rewritten to use MXE master?

tehnick:
> The first link does not mention MXE.

Yep. This is just a repo with builds of program and with simple README oriented to end users.

> Scripts from the second link seem to use a fork of MXE

It uses `hobby` branch of that fork, which has very small difference with `master` branch. The most important and only necessary change is a file:
https://github.com/sibuserv/mxe/blob/hobby/settings.mk

It includes the complete list of libraries required for build of Psi+ and EiskaltDC++ projects.

Also it forces usage of GCC 6.x, because QtWebKit for i686-w64-mingw32.shared target causes  segfaults if built with GCC < 6.x or with GCC 7.x. Hope what QtWebKit 5.212.0 with a huge amount of improvements will be released in a foreseeable future...

> Can they be rewritten to use MXE master?

There is nothing to change in these scripts. MXE directory used for build is set in Sibuserv settings.

tehnick:
Tiny rectification:
> The `hobby` branch of that fork is used.

This is for MS Windows 7 and newer.

For MS Windows XP a separate branch `hobby-winxp` is used.

tehnick:
Ping?

tehnick:
Rebased (again).

Please merge or reject.

tonytheodore:
Looks good to me - thanks!

P.S. I'm not sure how you added `settings.mk` without changing `.gitignore`; but if it works, it works. This form of customisation (file-based) isn't going away, but command line options are preferable in scripts. I still haven't thought through all the cases of plugins (#1015 #1259), and if there's any changes there - it may break your scripts. Better to use declarative options specifying what you want, `settings.mk` is really just a shortcut for session/terminal usage.

tehnick:
@tonytheodore  Thanks for info!

> P.S. I'm not sure how you added `settings.mk` without changing `.gitignore`

git is very flexible and manageable so this is simple:
```
git add settings.mk -f
git commit -m "add settings.mk"
```

PR #1508, opened 2016-08-26
openssl: update

mabrand:
Please read http://mxe.cc/#creating-packages

In particular, make sure that your build rules:
- install .pc file,
- install bin/test-pkg.exe compiled with flags by pkg-config,
- install .dll to bin/ and .a, .dll.a to lib/,
- use $(TARGET)-cmake instead of cmake,
- build in `$(BUILD_DIR)` instead of `$(SOURCE_DIR)`,
- do not run target executables with Wine,
- do not download anything while building,
- do not install documentation,
- do not install .exe files except test and build systems,

and .patch files are generated by tools/patch-tool-mxe.

If you add a package, you can use tool tools/skeleton.py.

Thanks!

PR #2057, opened 2018-02-25
Update openssl & switch qt4 pkgs to qt5

tonytheodore:
#1508 openssl update (breaks qt4 so move to plugins)
#2043 boost update (1.66 breaks various packages so downgraded to 1.65.1)
replaces #1549

Tested on i686-w64-mingw32.static, running complete tests from clean now.


mabrand:
I just want to mention that vmime doesn't build with openssl 1.1. https://github.com/kisli/vmime/issues/146.

Since MXE builds vmime with gnutls support, it's not a blocker. It's just a problem for users who want to use vmime configured for openssl. I haven't seen any patches for this and wonder if someone is working on it.

tonytheodore:
I've added an [example of (re)enabling openssl v1.0.x support for dcmtk and vmime](https://github.com/mxe/mxe/pull/2057/commits/d0120d40594bb76129aca2b50a229ca93e215719) for users who need it.

I think this is good to go, all packages build on all targets on Ubuntu 16.04 and Darwin 17.4. Some notes:
  - Qt4 [was fixed on archlinux](https://aur.archlinux.org/cgit/aur.git/tree/qt4-openssl-1.1.patch?h=lib32-qt4&id=ae544c977343bbf0dfdd1f2901bf0e5fe27d5233) so I've added that while converting any remaining packages over to Qt5.
  - I wasn't able to update boost, version 1.62 has a fix for openssl, but versions 1.62 to 1.65 have broken shared builds. boost 1.66 fixes the shared build, but breaks wt and pcl (not fixed in upstream updates). I ended up just taking the net changes for boost.asio from 1.62 and applying them to current 1.60 as a minimal fix for openssl.

Rinat84:
>      $(WGET) -q -O- 'https://www.openssl.org/source/' | \
>      grep "openssl-1.0.2" | \
>      $(SED) -n 's,.*openssl-\([0-9][0-9a-z.]*\)\.tar.*,\1,p' | \
for Long Term Support (LTS) openssl

tonytheodore:
> for Long Term Support (LTS) openssl

Done https://github.com/mxe/mxe/pull/2057/commits/7e8515fc668d35e48ec71367df3e23fc77330fc8

tehnick:
> Qt4 was fixed on archlinux so I've added that while converting any remaining packages over to Qt5.

In Debian we have such patch too:
https://salsa.debian.org/qt-kde-team/qt/qt4-x11/blob/master/debian/patches/openssl_1.1.patch

tonytheodore:
The archlinux Qt4 patch is based on the same Debian bug report, but applies cleanly to the patched MXE sources. I also couldn't consolidate the patches with `git am` so it's done as a separate file.

mabrand:
Thanks for getting this done!

SmtpClient-for-Qt seems like a good alternative for vmime when wanting to use openssl 1.1. https://github.com/mxe/mxe/pull/2072/commits/6b6b3637969dfec28a9b7c2662dab8a8bdb91715

Issue #2064, opened 2018-03-03
build-pkg.lua fails creating debian packages


ooxi:
When trying to create debian packages using tools/build-pkg.lua, no `.deb` files are created. However, during the build I receive numerous messages like

```
dpkg-deb: error: package architecture is missing or empty
dpkg-deb: warning: parsing file './mxe-x86-64-w64-mingw32.static-binutils_2.28/DEBIAN/control' near line 16 package 'mxe-x86-64-w64-mingw32.static-binutils':
 empty value for architecture
```

How can we fix this issue?

ooxi:
I most likely found the reason, build-pkg.lua is using dpkg-architecture which I did not have installed. I'm currently testing whether this fixes the problem and will prepare a pull request for the documentation, if that's the case.

tonytheodore:
Ouch! That must happen after the entire MXE build as I can build with `MXE_BUILD_PKG_NO_DEBS` on darwin and still create the deb-control files (with missing arch).

ooxi:
I would have but thankfully you guys already provided `MXE_BUILD_PKG_MAX_ITEMS` which allowed my to experiment with much shorter iteration times :-)

Anyway for anybody looking at this in the future, make sure you have the commands `dpkg-deb` as well as `dpkg-architecture` available, e.g. by installing the Debian package `dpkg` as well as `dpkg-dev`

tonytheodore:
Also note the `MXE_BUILD_PKG_NO_SECOND_PASS` option if you just want to build deb pkgs. It reveals [missing deps](https://github.com/mxe/mxe/commit/b5b527f782f8526a2094f724b9c5ecfd0bbe8eec) but more than doubles the runtime.

ooxi:
I did not realize that, thank you!

PR #2065, opened 2018-03-03
Update build-pkg.lua dependencies

ooxi:
build-pkg.lua uses `dpkg-architecture` to determine the architecture of a package in function `getArch`, however the program was not declared as dependency.

Fixes issue #2064

### DIFF
--- a/tools/build-pkg.lua
+++ b/tools/build-pkg.lua
@@ -6,7 +6,7 @@ This file is part of MXE. See LICENSE.md for licensing information.
 build-pkg, Build binary packages from MXE packages
 Instructions: http://pkg.mxe.cc
 
-Requirements: MXE, lua, fakeroot, dpkg-deb.
+Requirements: MXE, lua, fakeroot, dpkg-deb, dpkg-architecture.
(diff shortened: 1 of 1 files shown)


tonytheodore:
`dpkg-deb` seems to be a mispelling of [`dpkg-dev`](https://packages.debian.org/stretch/all/dpkg-dev/filelist).

tonytheodore:
Wouldn't the requirements be better as package names rather than individual tools?

ooxi:
> Wouldn't the requirements be better as package names rather than individual tools?

I sure wouldn't disagree :-) it's useful either way for noobies like me to have the Debian packages you need to install or the commands to be executed in the documentation.

Feel free to close this pull request and add `dpgk, dpkg-dev` to the documentation :-)

tonytheodore:
I added some copy-paste notes for Debian and ones I use myself on darwin https://github.com/mxe/mxe/commit/b7d071aaca27077acac7248e877675109ee22d46.

ooxi:
Thank you, that's even better!

PR #2076, opened 2018-03-10
Fix OGG not working in SDL_mixer 1.2.12

SupSuper:
Due to a autoconf error, SDL_mixer 1.2.12 does not correctly detect the OGG Vorbis library when using BUILD_STATIC. This was probably missed because normally SDL_Mixer is built using shared libraries instead, and wasn't fixed until SDL_mixer 2.0.

However, since static building is the default for mxe, I thought it was important to carry this configuration fix over to the old version.


tonytheodore:
Thanks! I [consolidated the patches and removed the patch to `configure.in`](https://github.com/mxe/mxe/commit/2573d98c7ed462ca38563dd52168860f45cf27de) - this isn't necessary and could trigger autoreconf.

Issue #2069, opened 2018-03-05
How to build only certain packages using build-pkg.lua


ooxi:
Is it possible to use build-pkg.lua to build the Debian packages only for a subset of MXE packages and their dependencies?

I am aware of `MXE_BUILD_PKG_MAX_ITEMS`, however I don't know how to influence which packages should be build exactly.

If it's not possible yet (which I think it is not), it should be fairly easy to implement by restricting [`items` here](https://github.com/mxe/mxe/blob/master/tools/build-pkg.lua#L1092).

Would you be open for a patch? This feature would be really handy for me since I want up to date Debian packages but only for a subset of components.

tonytheodore:
Good idea! We should also add the ability to specify plugins so people can build their own PPAs for gcc7 etc.

> it should be fairly easy to implement by restricting items here.

It's probably better to filter the list before [sending it to `build-pkg`](https://github.com/mxe/mxe/blob/master/Makefile#L875) as the upstream pkgs need to be included also and build-pkg doesn't know about source-only packages. I've been meaning to tidy up the `show-deps` functions since #1990 so will have a look at it.

As an aside, there's a sample repo at https://github.com/mxe/mxe-apt

ooxi:
Thanks for taking a look at it! I'll also try to hack a PoC for the `build-pkg` changes.

tonytheodore:
See #2078 for new env var `MXE_BUILD_PKG_PKGS`. `MXE_PLUGIN_DIRS` already works.

tonytheodore:
#2078 is good to go.

PR #2078, opened 2018-03-11
Add option to build subset of pkgs with build-pkg

tonytheodore:
See #2069


tools/build-pkg.lua
```diff
@@ -32,6 +36,13 @@ set environment variable MXE_BUILD_PKG_TARGETS to
 the list of targets separated by space.
 By default, all 4 major targets are built.
 
+To set list of MXE packages to build,
+set environment variable MXE_BUILD_PKG_PKGS to
+the list of packages separated by space. This is similar
+to a normal `make` invocation in that all dependencies
+will be built, so list just the packages you require.
+By default, all packages are built.
+
```
ooxi: This is epic, thank you so much!

Issue #2073, opened 2018-03-08
GH_CONF with package name as prefix fails


aquiles2k:
Hi,
im trying to create a package for xmlsec, a xml signing lib.

See: https://www.aleksey.com/xmlsec/

But the tags look something like `xmlsec-1_2_25.tar.gz`
See: https://github.com/lsh123/xmlsec/tags

So when using `$(PKG)_GH_CONF := lsh123/xmlsec/tags,xmlsec-,,,_` it seems mxe is throwing away the xmlsec part of the filename.

When doing make update-checksum-xmlsec or just make xmlsec after manually filling in the checksum the url he is aiming for is always just lsh123/xmlsec/archive/1_2_25.tar.gz. If i change xmlsec to any other string in GH_CONF it works as intended.

aquiles2k:
For completeness:
xmlsec.mk
```
# This file is part of MXE. See LICENSE.md for licensing information.

PKG             := xmlsec
$(PKG)_WEBSITE  := https://www.aleksey.com/xmlsec/
$(PKG)_IGNORE   :=
$(PKG)_VERSION  := 1.2.25
$(PKG)_CHECKSUM := 
$(PKG)_GH_CONF  := lsh123/xmlsec/tags,xmlsec-,,,_
$(PKG)_DEPS     := cc libxml2 libxslt openssl gnutls

define $(PKG)_BUILD
    cd '$(BUILD_DIR)' && $(TARGET)-cmake '$(SOURCE_DIR)'
    $(MAKE) -C '$(BUILD_DIR)' -j '$(JOBS)' VERBOSE=1
    $(MAKE) -C '$(BUILD_DIR)' -j 1 install VERBOSE=1
endef

```
checksum update
```
make update-checksum-xmlsec
[...]

Issue #1982, opened 2017-11-18
Standardise GH_CONF naming


tonytheodore:
#1969 introduced tags and release naming:
```
  GH_CONF := owner/repo/branch
  GH_CONF := owner/repo/releases[/latest], tag prefix, tag suffix, tag filter-out, version separator
  GH_CONF := owner/repo/tags, tag prefix, tag suffix, tag filter-out, version separator
```
and comment:
```
# Using the third segment for api means you can't track a branch with these names:
GH_APIS := releases tags
```
Branch tracking should really be:
```
  GH_CONF := owner/repo/branches/branch
```
for consistency.

PR #2080, opened 2018-03-13
GH_CONF changes

tonytheodore:


tonytheodore:
Tested on all targets on Ubuntu 16.04 and Darwin 17.4.0.

There's a few more packages that can be changed but I'll do them separately - the main thing is that `GH_CONF` is standardised (#1982) and works as expected (#2073).

PR #2082, opened 2018-03-15
[WIP] Updates

tonytheodore:


tonytheodore:
Tested all pkgs on all target on Ubuntu 16.04 and Darwin 17.4.0

PR #2084, opened 2018-03-16
Add package osgearth

tehnick:
Build is tested for:
```
MXE_TARGETS :=  i686-w64-mingw32.static i686-w64-mingw32.shared \
                x86_64-w64-mingw32.static x86_64-w64-mingw32.shared
```

### DIFF
--- a/src/osgearth-1-fixes.patch
+++ b/src/osgearth-1-fixes.patch
@@ -1,0 +1,71 @@
+This file is part of MXE. See LICENSE.md for licensing information.
+
+Contains ad hoc patches for cross building.
+
+From 0000000000000000000000000000000000000000
+From: Boris Pek <tehnick-8@yandex.ru>
+Date: Fri, 22 Dec 2017 23:09:25 +0300
+Subject: [PATCH 1/3] fix osgEarthSplat static library
+
+
+diff --git a/src/osgEarthSplat/Export b/src/osgEarthSplat/Export
+index 1111111..2222222 100644
+--- a/src/osgEarthSplat/Export
++++ b/src/osgEarthSplat/Export
+@@ -42,7 +42,7 @@
+ 
+ #if defined(_MSC_VER) || defined(__CYGWIN__) || defined(__MINGW32__) || defined( __BCPLUSPLUS__)  || defined( __MWERKS__)
+     #  if defined( OSGEARTHSPLAT_LIBRARY_STATIC )
+-    #    define OSGEARTH_SPLAT_EXPORT
++    #    define OSGEARTHSPLAT_EXPORT
+     #  elif defined( OSGEARTHSPLAT_LIBRARY )
+     #    define OSGEARTHSPLAT_EXPORT   __declspec(dllexport)
+     #  else
+
+From 0000000000000000000000000000000000000000
+From: Boris Pek <tehnick-8@yandex.ru>
+Date: Fri, 22 Dec 2017 23:15:21 +0300
+Subject: [PATCH 2/3] disable usage of extra libraries
+
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 1111111..2222222 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -126,10 +126,10 @@ FIND_PACKAGE(OpenGL)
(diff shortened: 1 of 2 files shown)


src/osgearth.mk
```diff
@@ -0,0 +1,27 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := osgearth
+$(PKG)_WEBSITE  := http://osgearth.org/
+$(PKG)_DESCR    := Geospatial SDK for OpenSceneGraph
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 2.9
+$(PKG)_CHECKSUM := 22aeef42bb700c1e669d9ba57ce4155c0668caf86cc738750a6b9d34c1eaf2a4
+$(PKG)_GH_CONF  := gwaldron/osgearth/tags, osgearth-
```
tonytheodore: Use 
```
$(PKG)_GH_CONF  := gwaldron/osgearth/releases/latest, osgearth-
```
to ignore the `simdis-sdk`

src/osgearth.mk
```diff
@@ -0,0 +1,27 @@
...
+$(PKG)_DEPS     := curl gcc gdal openscenegraph sqlite tinyxml2 zlib
```
tonytheodore: It seems to actually use tinyxml and the build fails for me on shared targets. Can you make a gist of one of the shared logs?

tonytheodore: There's not much in the log but I see what's likely happened - you started from clean (good practice) and didn't have tinyxml present so it used its internal one. I'd just use the internal one as our package will be removed eventually.

src/osgearth-1-fixes.patch
```diff
@@ -0,0 +1,59 @@
...
+@@ -117,10 +117,10 @@ ENDIF (OSGEARTH_USE_GLES)
+ 
+ FIND_PACKAGE(CURL)
+ FIND_PACKAGE(GDAL)
+-FIND_PACKAGE(GEOS)
++#FIND_PACKAGE(GEOS)
+ FIND_PACKAGE(Sqlite3)
+ FIND_PACKAGE(ZLIB)
+-FIND_PACKAGE(Poco)
++#FIND_PACKAGE(Poco)
```
tonytheodore: Can these be enabled? It's preferable to enable as many features as possible.

If not, can they be made options and sent upstream? It will make future updates easier (also for apps and tests below)

src/osgearth.mk
tonytheodore: Also, use `cc` instead of `gcc` an place it first in deps. `gcc` is only really for people what want to build a barebones compiler.

src/osgearth.mk
tehnick: It works. Thanks.

src/osgearth.mk
tehnick: > There's not much in the log but I see what's likely happened - you started from clean (good practice) and didn't have tinyxml present

Yes, that is the case.

> so it used its internal one

I'll check it.

> Also, use `cc` instead of `gcc` an place it first in deps.

Just forgot to update this dependency. This package was prepared in 2017 and only now I have found time to propose it to upstream...

src/osgearth-1-fixes.patch
tehnick: > Can these be enabled? It's preferable to enable as many features as possible.

Yes, sure. I just forgot to drop this part of patch during preparing of PR. (I do not use this part of osgearth in my MXE fork.)

> If not, can they be made options and sent upstream? It will make future updates easier (also for apps and tests below)

Ok. I'll do this for apps and tests.

tehnick:
I have updated this PR. Patch is proposed to upstream: https://github.com/gwaldron/osgearth/pull/1076

tehnick:
Hmm, I see that build for *.shared with enabled `geos` is broken. Will look at it later.

tehnick:
One more PR: https://github.com/gwaldron/osgearth/pull/1077

tehnick:
@tonytheodore Now this PR should be ready for merging. Please re-check it.

tonytheodore:
Builds on all targets and looks good - thanks!

PR #2085, opened 2018-03-19
libdvdread: force absolute path to accomplish library build

artynet:
```
$ ~/MPVCROSS/mxe-trunk/tmp-libdvdread-x86_64-w64-mingw32.static/libdvdread-6.0.0/src/dvd_input.c:56:133: 
fatal error: /home/user/MPVCROSS/mxe-trunk/tmp-libdvdread-x86_64-w64-mingw32.static/libdvdread-6.0.0/msvc/contrib/dlfcn.c: No such file or directory
compilation terminated.
```
The import  path needs to be forced otherwise the library won't build. This prevents the build of libdvdnav as well.

### DIFF
--- a/src/libdvdread.mk
+++ b/src/libdvdread.mk
@@ -22,6 +22,8 @@ define $(PKG)_UPDATE
 endef
 
 define $(PKG)_BUILD
+    # sed modification to force absolute path to the file
+    $(SED) -i "s~../msvc~$(SOURCE_DIR)/msvc/msvc~g" $(SOURCE_DIR)/src/dvd_input.c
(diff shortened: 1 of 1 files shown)


tonytheodore:
Looks like it always uses `dlfcn`, regardless of the comment:

```
# libdvdread supports libdvdcss either by dynamic loading (dlfcn-win32) or
# directly linking to libdvdcss. We directly links to the library here.
$(PKG)_DEPS     := cc libdvdcss
```

We should either add `dlfcn-win32` to deps or explicitly disable dynamic loading somehow.

PR #2090, opened 2018-03-22
add package xmlsec

aquiles2k:
A small lib implementing XML signing based on W3C Standard: XMLDSIG using libxml2 and changeable encyrption backends. We needed it for one of our projects and i thought it might be a good addition to mxe for others to use.

See:
-https://www.aleksey.com/xmlsec/
-https://www.w3.org/TR/xmldsig-core/
-https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-xmlsec/PKGBUILD


aquiles2k:
Should i care for the deploy preview?

aquiles2k:
@tonytheodore Do you know whats that with deploy/netlify?

tonytheodore:
I wouldn't worry about the deploy preview, it seems to still happen if I retry, but will probably fix itself.

A couple of points:
  - static builds are failing
  - please use $(SOURCE_DIR) and $(BUILD_DIR)
  - add a link test using one of the installed *.pc files - this may need patching to ensure [XMLSEC_STATIC](https://github.com/lsh123/xmlsec/blob/master/include/xmlsec/exports.h#L25) is defined.
  - disable apps

aquiles2k:
doh the rebase didnt work, ill close this pr and reopen a new one

PR #2094, opened 2018-03-27
add package xmlsec

aquiles2k:
A small lib implementing XML signing based on W3C Standard: XMLDSIG using libxml2 and changeable encyrption backends. We needed it for one of our projects and i thought it might be a good addition to mxe for others to use.

See:
-https://www.aleksey.com/xmlsec/
-https://www.w3.org/TR/xmldsig-core/
-https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-xmlsec/PKGBUILD

I tested with i686, x64, shared and static.

### DIFF
--- a/src/xmlsec.mk
+++ b/src/xmlsec.mk
@@ -1,0 +1,22 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := xmlsec
+$(PKG)_WEBSITE  := https://www.aleksey.com/xmlsec/
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 1.2.25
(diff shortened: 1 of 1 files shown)


aquiles2k:
@tonytheodore sorry for the noise with #2090 :(

I now also tested static, disabled apps as well as docs and used $(SOURCE_DIR) and $(BUILD_DIR).

I'm not sure about test linkage with generated .pc files as i've no experience with this and normally dont work with statical linkinage or pkg-config. Could you help with that or can we postpone this?

tonytheodore:
Thanks! Merged in https://github.com/mxe/mxe/compare/9afbe320...85e059ca with a *.pc fix sent upstream https://github.com/lsh123/xmlsec/pull/162

aquiles2k:
Thanks alot!

PR #2099, opened 2018-04-02
upgrade package and add LIBS

dl5rcw:
New version of hamlib available - 3.1 to 3.2
It now requires lusb-1.0 to compile, so do some modules when using the library.
Tested on two debian 9 servers.


src/hamlib.mk
```diff
@@ -26,11 +27,11 @@ define $(PKG)_BUILD
      echo 'Description: $(PKG)'; \
      echo 'Requires: libusb-1.0'; \
      echo 'Cflags: -DHAMLIB_LIB'; \
-     echo 'Libs: -lhamlib -lpthread -lws2_32';) \
+     echo 'Libs: -lhamlib -lpthread -lws2_32 -lusb-1.0';) \
```
tonytheodore: The `Requires` above should mean you don't have to manually specify the lib (likewise for test program below).

dl5rcw: @tonytheodore you are right for the test file, but the configure will fail w/o the LIBS.

Without the manual LIBS='-lusb-1.0' \, hamlib will fail to build
`checking for libusb_init in -lusb-1.0... no`
here is a [gist](https://gist.github.com/dl5rcw/9485029267effe16da92386496658f17) with the complete output

If I place `LIBS='-lusb-1.0' \`, hamlib compiles fine.
If I place the `echo '' > *.pc` file to the top, hamlib does not compile - obviously the order in the *mk file does not affect the situation.

If I remove the `libusb-1.0` from test file, but keep the manual LIBS='-lusb-1.0' \ at the configure part, hamlib compiles fine.

src/hamlib.mk
```diff
@@ -6,14 +6,18 @@ PKG             := hamlib
...
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 3.1
-$(PKG)_CHECKSUM := 682304c3e88ff6ccfd6a5fc28b33bcc95d2d0a54321973fef015ff62570c994e
+$(PKG)_VERSION  := 3.2
+$(PKG)_CHECKSUM := b55cb5e6a8e876cceb86590c594ea5a6eb5dff2e30fc13ce053b46baa6d7ad1d
 $(PKG)_GH_CONF  := hamlib/hamlib/releases/latest
 $(PKG)_DEPS     := cc libltdl libusb1 libxml2 pthreads
 
+#LIBS='-lusb-1.0' \
+
```
tonytheodore: Why the comment and whitespace? (also whitespace below)

src/hamlib.mk
```diff
@@ -6,14 +6,18 @@ PKG             := hamlib
...
 define $(PKG)_BUILD
+
     cd '$(BUILD_DIR)' && '$(SOURCE_DIR)/configure' \
         $(MXE_CONFIGURE_OPTS) \
+        LIBS='-lusb-1.0' \
```
tonytheodore: This only seems to be required for `i686-w64-mingw32.static` - there's some subtle failure in the detection for this target that is probably worth investigating.

src/hamlib.mk
```diff
@@ -6,14 +6,18 @@ PKG             := hamlib
...
         --with-included-ltdl \
```
tonytheodore: This `--with-included-ltdl` option is no longer recognised, and seemingly conflicts with the DEPS above. Both should be removed.

tonytheodore:
Hamlib provides it's own pkg-config file, can we use that instead of our custom one? Ours sets `Cflags: -DHAMLIB_LIB` but there's no `HAMLIB_LIB` define anywhere in the source. Also the pkg-config invocation for the test program includes a superfluous `pthreads` - this should be removed.

I know I'm being pedantic on this update - the package is very close to straight "mechanical" updates (just version and checksum) with no custom logic and it seems worth the stretch to get to that point.

src/hamlib.mk
```diff
@@ -6,31 +6,24 @@ PKG             := hamlib
...
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 3.1
-$(PKG)_CHECKSUM := 682304c3e88ff6ccfd6a5fc28b33bcc95d2d0a54321973fef015ff62570c994e
+$(PKG)_VERSION  := 3.2
+$(PKG)_CHECKSUM := b55cb5e6a8e876cceb86590c594ea5a6eb5dff2e30fc13ce053b46baa6d7ad1d
 $(PKG)_GH_CONF  := hamlib/hamlib/releases/latest
-$(PKG)_DEPS     := cc libltdl libusb1 libxml2 pthreads
+$(PKG)_DEPS     := cc libusb1 libxml2 pthreads
+
+#LIBS='-lusb-1.0' \
```
tonytheodore: Remove this comment and the new blank lines and it's good to go.

dl5rcw: done. Just missed the line in cleanup. sorry
Good remark - thank you!

PR #2111, opened 2018-05-01
Update to GStreamer 1.14.0

jonaski:


tonytheodore:
I think this may break OpenSceneGraph, I have a wip branch that updates both together but can't remember if I was just trying to bundle the updates.

Thanks!

PR #2113, opened 2018-05-05
poppler: disable reference docs (close #1820)

### LINKED ISSUE #1820: poppler fails to build unless this line
$(SED) -i 's,reference,'',g' '$(1)/glib/Makefile'
I put this after configure because it still tries to build and install reference docs and where it does not have permission.

darealshinji:
Fix is from @jjbenham (see #1820).


tonytheodore:
Normally try to avoid `sed` but there doesn't seem to be a proper config option and it's not worth patching since upstream is already far ahead.

Thanks!

PR #2123, opened 2018-05-20
openjpeg: explicitly enable pkg-config files

darealshinji:
Running the native cmake automatically enables `BUILD_PKGCONFIG_FILES` but the cross-toolchain wrapper disables it.
Related to https://github.com/mxe/mxe/issues/1446


tonytheodore:
Thanks!

PR #2127, opened 2018-05-21
ghostscript: improving cross-building

darealshinji:
First, run configure with the option `--with-drivers=ALL,display`. Upstream confirmed an error in the build process that keeps the display driver disabled when building a static library, but that driver is needed if you want to use libgs through the wrapper library libspectre. So creating a static library manually from the DLL object files isn't required anymore.

Also a massive update on the patch, mostly to enable real support for `--host` on the configure script (instead of relying on `uname`). Full support for `--build` wasn't added to keep the patch smaller and to not make things more complicated as they already are (for example configure runs itself recursively when CCAUX environment variable was set). ghostscript.mk now builds the needed auxiliary tools in a separate stage (as native binaries) and deletes the unwanted third-party library sources afterwards. This ensures those binaries will run on the build system and keeping the third-party sources on that step will ensure that configuring doesn't fail if i.e. native libjpeg development files aren't found.

Builds on all four MXE targets and the libraries are confirmed to work.

By the way this PR could need some tweaks if/when PR https://github.com/mxe/mxe/pull/2123 gets merged.


tonytheodore:
Build fails on Darwin during cross-build, native stage seems fine:
https://gist.github.com/tonytheodore/06bc5bc9d79ba4e5437616b4d954983b

darealshinji:
I had that issue before on Linux. Maybe the line `CPPFLAGS='$(CPPFLAGS) -DHAVE_SYS_TIMES_H=0'` should be kept.

By the I'm currently working on implementing real support for $build and $host and want to submit that to upstream. I could backport and test those changes on the release version 9.23 once I'm done.

darealshinji:
Here are changes that make it work with MXE: https://github.com/darealshinji/ghostscript/commit/c8013ed697afbc0214c5cd80a868c11f544e2a03
I will soon backport only those fixes relevant to MXE and the GS release version 9.23.

tonytheodore:
> Maybe the line CPPFLAGS='$(CPPFLAGS) -DHAVE_SYS_TIMES_H=0' should be kept.

Thanks, I can confirm that putting it back in fixes it.

The shared build fails at the test program since openjpeg doesn't install the *.pc file for shared builds (#2123).

darealshinji:
Hmm, it's present for me in `usr/x86_64-w64-mingw32.shared/lib/pkgconfig/libopenjp2.pc`

tonytheodore:
Okay, seems to be working now after forcing rebuild of openjpeg. Manually merged along with CPPFLAGS fix in https://github.com/mxe/mxe/compare/579070c1...b40d6b70.

Thanks!

P.S. I didn't see you're update about libpaper but that's good to go also.

Issue #2095, opened 2018-03-27
qt5-host-tools build failure


zgyarmati:
I'm trying to set up a host toolchain with Qt5, so on the current master i run:
make qt5-host-tools
with
MXE_PLUGIN_DIRS += plugins/examples/host-toolchain/  plugins/gcc7
with the target "x86_64-w64-mingw32.shared".
Unfortunately the qmake build fails: it seems that the build host's native g++ is invoked, which doesn't have support for the command line option ‘-fno-keep-inline-dllexport’, see the log at:
https://gist.github.com/zgyarmati/242ffba37dff0f15b5ce007d55229ab8

After looking into this, turned out that the '-host-option' configure argument is not supported in the current Qt version (removed with the commit: https://github.com/qt/qtbase/commit/42196f4061263d6d0e453c0561f7604cadc6d0a3#diff-24e7ddbbe95f1829bed41405948f2130 ), but it's still used in 'plugins/examples/host-toolchain/qt5-host-tools.mk', so it needs to be fixed, but so far i couldn't figure out how to get qmake built with the cross-compiler.

tonytheodore:
Without the `-host-option` the only way to use the cross-compiler is with a `sed` replacement since `configure` parses the `qmake.conf` files with `awk`:
```
$(SED) -i 's,$$$${CROSS_COMPILE},$(TARGET)-,g'   '$(SOURCE_DIR)/mkspecs/win32-g++/qmake.conf'
```
that will get you a little further, but then there's issues with `qfloat16-tables` - this seems to be a code generator that needs to be compiled with the native compiler and the `sed` breaks that.

PR #2131, opened 2018-05-26
host-toolchain: build make from source

### LINKED ISSUE #1924: Cross-compile GNU Make
From https://github.com/mxe/mxe/tree/master/plugins/examples/host-toolchain#make:

> Make is difficult to cross-compile so it is downloaded from the [source
recommended by the GNU Make team](https://git.savannah.gnu.org/cgit/make.git/tree/README.W32.template)

We really should build this ourselves;)

darealshinji:
Should close #1924


tonytheodore:
Nice - it runs on Windows though I can't test it with `qmake` since that's broken with later Qt versions (#2095).

Thanks!

PR #2138, opened 2018-06-07
Fix Clementine build

jonaski:


jonaski:
clementine_qt4 require the Qt4 version of qjson. Either drop Qt4 version of clementine or add a Qt4 version of qjson.
You can make clementine_qt4 compile by changing qjson.mk, set $(PKG)_QT4_BOOL  := ON and remove line $(PKG)_QT_SUFFIX := -qt5

Issue #2140, opened 2018-06-10
Package `ffmpeg` for `x86_64-w64-mingw32.static.posix` fails to build at `gnutls`


PsychoXIVI:
Logs: https://gist.github.com/PsychoXIVI/ef561e8ff666b18dc20ceea192dfad02

At https://github.com/raspberrypi/noobs/issues/470#issuecomment-376256295 user `franzgl` mentions:

> The error is with automake and perl v5.26.
> In perl v5.22, using a literal { in a regular expression was deprecated, and will emit a warning if it isn't escaped: \{. In v5.26, this won't just warn, it'll cause a syntax error.

I have no idea (at moment when I am writing this) how to create patch file for it, so here is an issue.

Edit: Fixes from above are fine, now I am working on the patch.

PsychoXIVI:
I created patch for `automake` (by hand :F). I will create pull request later, but here is the patch:

https://gist.github.com/PsychoXIVI/d45ee860f43ae5bb073b0013f99da01c

It will do the job while placed in `mxe/plugins/native`, next to `automake.mk`.

PR #2144, opened 2018-06-11
Patched `automake` - avoid deprecated syntax that fail.

PsychoXIVI:
Contains patch for `automake` from `native` plugin`. That allows using newer version of Perl, by avoiding deprecated syntax that causes failure in newer versions of Perl. 

At raspberrypi/noobs#470 (comment) user franzgl mentions:
> The error is with automake and perl v5.26.
> In perl v5.22, using a literal { in a regular expression was deprecated, and will emit a warning if it isn't escaped: {. In v5.26, this won't just warn, it'll cause a syntax error.

That solve an issue #2140

### DIFF
--- a/src/libechonest-1-fixes.patch
+++ b/src/libechonest-1-fixes.patch
@@ -1,18 +1,7 @@
-This file is part of MXE. See LICENSE.md for licensing information.
-
-Contains ad hoc patches for cross building.
-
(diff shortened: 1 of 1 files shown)


PsychoXIVI:
Auto-closed because of my mistake: I used the master branch of my fork to provide pull request commits, but it is bad practice, and I even felt that, when I want to create another pull request. Well sorry, I am learning from mistakes 😞

Issue #1802, opened 2017-07-02
ffmpeg picks up random dependencies


TobiX:
Hi,

since ffmpeg does feature detection, it might pick up random dependencies, which are compiled before it. I just noticed this with SDL2 - currently ffmpeg doesn't declare a dependency to SDL2, but if you build SDL2 and then ffmpeg, ffmpeg "magically" gets SDL2 support. I was about to write a patch to make this more explicit, but I'm not sure how the MXE project handles such dependencies. Maybe it is even intended for MXE packages to pick up "optional" dependencies? (Probably not...)

To make it easy for people to "disable" features of packages, would something like the following patch be desired?

```diff
diff --git a/src/ffmpeg.mk b/src/ffmpeg.mk
index 74f3930d..45185072 100644
--- a/src/ffmpeg.mk
+++ b/src/ffmpeg.mk
@@ -10,7 +10,7 @@ $(PKG)_FILE     := $(PKG)-$($(PKG)_VERSION).tar.bz2
 $(PKG)_URL      := https://ffmpeg.org/releases/$($(PKG)_FILE)
 $(PKG)_URL_2    := https://launchpad.net/ffmpeg/$(call SHORT_PKG_VERSION,$(PKG))/$($(PKG)_VERSION)/+download/$($(PKG)_FILE)
 $(PKG)_DEPS     := gcc bzip2 gnutls lame libass libbluray libbs2b libcaca \
-                   libvpx opencore-amr opus sdl speex theora vidstab \
+                   libvpx opencore-amr opus sdl2 speex theora vidstab \
                    vo-amrwbenc vorbis x264 xvidcore yasm zlib
 
 # DO NOT ADD fdk-aac OR openssl SUPPORT.
@@ -56,6 +56,7 @@ define $(PKG)_BUILD
         --enable-libopencore-amrnb \
         --enable-libopencore-amrwb \
         --enable-libopus \
+        --$(if $(filter sdl2,$($(PKG)_DEPS)),en,dis)able-sdl2 \
         --enable-libspeex \
         --enable-libtheora \
         --enable-libvidstab \
```

This would make SDL2 support explicit and leaves users of MXE an easy way to switch it off (just change the dependencies), if desired. 
If this is the way to go, I would take a look at all other option of ffmpeg which are marked `[autodetect]` (there are quite a few) and include them in a similar way.

PR #2145, opened 2018-06-11
ffmpeg: fix ffplay compilation by changing sdl dependency to sdl2

withmorten:
ffmpeg updated to using sdl2 a while ago, but the dependency didn't get changed to that, so ffplay doesn't get compiled anymore. This should fix that.

Edit: Related to (or fixes) https://github.com/mxe/mxe/issues/1802.


tonytheodore:
Thanks!

PR #2146, opened 2018-06-11
Patched `automake` - avoid deprecated syntax that fail.

PsychoXIVI:
Contains patch for `automake` from `native` plugin. That allows using newer version of Perl, by avoiding deprecated syntax that causes failure.

Solved thanks to comment of user `franzgl` from raspberrypi/noobs#470 that mentions:
> The error is with automake and perl v5.26.
> In perl v5.22, using a literal { in a regular expression was deprecated, and will emit a warning if it isn't escaped: {. In v5.26, this won't just warn, it'll cause a syntax error.

That solve an issue #2140


tonytheodore:
Thanks! Merged in https://github.com/mxe/mxe/commit/10c9aa5d78496f4a79629108f932a1702a2288d0

PR #2153, opened 2018-06-17
allow disabling usage of fakeroot via MXE_BUILD_PKG_NO_FAKEROOT

ashkulz:
fakeroot hangs in a docker container and is not required (already root), so don't call it if `MXE_BUILD_PKG_NO_FAKEROOT` is set in the environment.


tonytheodore:
Thanks!

PR #2167, opened 2018-07-13
Secure URLs

vszakats:


vszakats:
Any issues to address with this PR to have it merged?

src/neon.mk
```diff
@@ -1,7 +1,7 @@
 # This file is part of MXE. See LICENSE.md for licensing information.
 
 PKG             := neon
-$(PKG)_WEBSITE  := http://webdav.org/neon/
+$(PKG)_WEBSITE  := https://web.archive.org/web/20070925040920/www.webdav.org/neon/
 $(PKG)_DESCR    := HTTP and WebDAV client library (libneon)
```
tonytheodore: Why not use the most recent capture?

vszakats: Good point, can't remember what was the reason in this case (I took this change from an earlier patch to another project). Generally I also prefer the default/latest capture, but it's often not what's expected. In this case it is, so an update is commited.

vszakats: (and squashed)

tonytheodore:
Thanks!

Issue #1909, opened 2017-09-24
Establish HTTPS for mxe.cc and pkg.mxe.cc


starius:
We can use Let's Encrypt for this.
Proposed in https://github.com/mxe/mxe/pull/1899#issuecomment-331719814

tonytheodore:
How would `mxe.cc` be handled? It's currently a github pages site and most recommendations involve using an intermediary with re-directs (which doesn't seem ideal).

vszakats:
https://www.netlify.com/ does support custom domains with HTTPS, it also integrates well with GitHub, and it's free for projects like MXE. I believe it was or is used for the Homebrew homepage.

(Some put their GitHub hosted pages behind a CloudFlare proxy, but that's a non-solution, as the traffic remains unencrypted/unprotected between CloudFlare and GitHub. Not very honest/secure.)

tonytheodore:
Netlify looks good though I'm not sure if we want to add a service to serve static html - we don't really use any gh-pages functionality.

vszakats:
Anything blocking to enable this for `mxe.cc`? It may need some DNS record updates, but the bigger part of HTTPS is support is automatically handled by GitHub, all it needs is enabling support for it on the project settings page.

tonytheodore:
@vog, can you take a look at this? The DNS records need to be updated as noted in https://help.github.com/articles/troubleshooting-custom-domains/#custom-domains-configured-with-a-records

vog:
@tonytheodore Thanks for bringing this to my attention. I'll take care of this as soon as possible.

vog:
Okay, I just changed the A records for mxe.cc from:

* 192.30.252.153
* 192.30.252.154

to:

* 185.199.108.153
* 185.199.109.153
* 185.199.110.153
* 185.199.111.153

Some remarks:

* Please note that it may take up to 24h for those changes to propagate through the DNS caches all over the world
* I did not need to change www.mxe.cc, as this is a CNAME for mxe.cc.
* "you must remove and re-add your custom domain to the repository you’re using to publish your Pages site to trigger the process of enabling HTTPS" (from https://help.github.com/articles/troubleshooting-custom-domains/#custom-domains-configured-with-a-records)

@tonytheodore Would you mind to take care of that last step?

vszakats:
Thank you @vog and @tonytheodore. Can someone update the address on the header of this GitHub project? [ https://github.com/mxe/mxe/ ]

tonytheodore:
[...]

PR #2170, opened 2018-07-30
Secure URLs

vszakats:


vog:
Related issue: #1909

vog:
This looks very good to me, but I'd prefer to have another person have a look at it, just to be sure.

vszakats:
Sadly, the Lua code requires a new approach as the current one doesn't support security. Not a Lua expert here, so I'm not sure how to do this, especially on Windows.

Currently it falls back to HTTP automatically and results in this:
```
$ lua -e 'http = require "socket.http"; print(http.request("https://mxe.cc/"))'
<html>
<head><title>301 Moved Permanently</title></head>
<body bgcolor="white">
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>
	301	table: 0x7fac6ec06b40	HTTP/1.1 301 Moved Permanently
```

vszakats:
..but, since the actual URL used there seems irrelevant (both are just examples), it can be fixed by using an HTTP example URL. I'll commit it.

tonytheodore:
Looks good to me, thanks!

PR #2171, opened 2018-07-30
Add QJSON patch

drizt:
https://github.com/flavio/qjson/issues/48#ref-pullrequest-345139226


tonytheodore:
Thanks! Merged and consolidated in https://github.com/mxe/mxe/compare/8730bed...a99fed2

PR #2174, opened 2018-08-07
WIP: update glib and friends

tonytheodore:
- removes WinXP support
- requires posix toolchain for glibmm


tonytheodore:
Looks like [libsigc++3 is required for glibmm](https://github.com/GNOME/glibmm/commit/e3465005619f8428c84b6c53cfd73d2dbab5ed4c), but it hasn't been released yet and requires c++17 compiler support.

Janrupf:
Not 100% sure, but libsigc++3 is in libsigc++ 2.99.11. Seems a bit weird, but this is how it is.

tonytheodore:
That's just the odd/even release numbering used by some projects, it isn't actually required for the current release branch of glibmm.

atkmm is now also updated and I think I've got a general way to disable the non-posix builds and fix the update scripts.

PR #2178, opened 2018-08-13
secure/fix URLs

vszakats:


vszakats:
This patch:
- broken links with working secure ones
- removes two broken (insecure) mirrors
- fixes a regression in finding newer versions after switching to a secure download page
- removes references to bzip.org, which was not renewed and now for sale [<https://lwn.net/Articles/762264/>]
- upgrades some more URLs to secure ones
Let me know if there is any concern with it.

tonytheodore:
Looks good, bzip should point to the project archive at https://web.archive.org/web/20180801004107/http://www.bzip.org/ instead of wikipedia, otherwise it's good to merge.

vszakats:
I find referring to the old domain in any way to be risky/dangerous, because that domain is up for sale and any new content isn't guaranteed to be safe or genuine. For this reason the original domain was also removed from the wikipedia page as soon as it expired and went on sale at a domain broker a few days ago. A new owner may also trivially remove any past `archive.org` content. What do you think?

tonytheodore:
I didn't realise the new owner could change past content, wikipedia is better in that case.

Thanks!

PR #2181, opened 2018-08-17
use XGCC_W32_PREFIX for newer nsis

donbright:
new versions of NSIS require XGCC_W32_PREFIX not MINGW_CROSS_PREFIX, see http://nsis.sourceforge.net/Docs/AppendixG.html

Thanks!


tonytheodore:
Looks good, I think the patch can be removed also. `XGCC_W32_PREFIX ` is noted in the original issue http://sourceforge.net/tracker/index.php?func=detail&aid=3305366&group_id=22049&atid=373085

tonytheodore:
Thanks!

tonytheodore:
Patch and variable removed https://github.com/mxe/mxe/commit/13d6257891f94d539b29a786641cef60b0874233

PR #2182, opened 2018-08-17
scons use python2 when python3 is default python

donbright:
scons is broken on systems where python3 is default python, like ubuntu18.  this breaks nsis since it depends on scons.

this is because scons is actually a script that begins with #! /usr/bin/python

what we do here, is quite frankly a kludge, as i am not overly familiar with the language of Makefile , but A. it works, B, i hope it inspires someone else to improve it.

basically, we detect whether the system has a 'python2' binary in /usr/bin, this is my quick n dirty version of determining if a system is one that likely has python3 as default, since most systems that have done this will retain python2 as its own binary. 

then this copies scons from /usr/bin to our build dir, then modifies this copy of the scons script itself to use # /usr/bin/python2 instead of usr/bin/python (which is likely python3 on such machines)

then instead of calling 'scons' we call 'builddir/scons' which runs scons with python2

Thanks

### DIFF
--- a/src/nsis.mk
+++ b/src/nsis.mk
@@ -11,6 +11,12 @@ $(PKG)_FILE     := nsis-$($(PKG)_VERSION)-src.tar.bz2
 $(PKG)_URL      := https://$(SOURCEFORGE_MIRROR)/project/nsis/NSIS 3/$($(PKG)_VERSION)/$($(PKG)_FILE)
 $(PKG)_DEPS     := cc
 
+define SCONSBIN
+   $(if $(findstring python2,`[ -e /usr/bin/python2 ] && echo python2`), \
+      cp /usr/bin/scons $(1)/ && sed -i s/python/python2/g $(1)/scons && ./scons, \
+      scons)
+endef
+
 define $(PKG)_UPDATE
     $(WGET) -q -O- 'https://nsis.sourceforge.io/Download' | \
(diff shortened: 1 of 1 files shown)


tonytheodore:
I started [adding the scons-local package](https://gist.github.com/tonytheodore/4589f0be38a701afa7b1dc6cf8942b20) and was about to remove it from the requirements when I realised this approach isn't going to work - there's no easy way to guess the various platform conventions.

The package manager that installed scons is going to explicitly depend on python2 (see [archlinux scons](https://www.archlinux.org/packages/extra/any/scons/)) and embed the python used in the shebang line. If something changes after that, there's not a lot we can do in a portable way (short of building our own python2).

On Ubuntu, python3 doesn't install plain `/usr/bin/python` (compare [python3](https://packages.ubuntu.com/bionic/amd64/python3-minimal/filelist) and [python](https://packages.ubuntu.com/bionic/amd64/python-minimal/filelist)) - something else has changed that in your setup. I'm not an apt user, but you can check which package installed it with:

```
$ dpkg -S /usr/bin/python
python-minimal: /usr/bin/python
```

I'll think about it some more.

tonytheodore:
> there's no easy way to guess the various platform conventions.

It's not too hard, we just need any python2 and can test: python == 2 > python2 > python2.7. That covers most of what I've seen. I've done #2184 to replace this.

PR #2184, opened 2018-08-18
nsis: use scons-local with python2

tonytheodore:
Replaces #2182 

@donbright, can you confirm if this works on your system? I'll remove scons from the reqs if it does.


donbright:
tested, works, !!!! awesome.

PR #2179, opened 2018-08-13
gcc7 fixes

tonytheodore:
Better approach to #1860 plus various updates, selective `Wno-error`, and setting c++ standards.

PR #2185, opened 2018-08-18
mpc: update and fixed version checking

PsychoXIVI:
Updated and fixed version checking for `mpc` library, so its updates and compile now. :F 

Links to downloads expired:
https://gforge.inria.fr/scm/viewvc.php/tags/?root=mpc&sortby=date outdate from 3 years.
https://deb.debian.org/debian/pool/main/m/mpclib/mpclib_ - non existing for now.

Outdated version was causing compilation errors:
https://gist.github.com/PsychoXIVI/8ec2305c44e603d411a0da95e9cbd27a

### DIFF
--- a/src/mpc.mk
+++ b/src/mpc.mk
@@ -15,11 +15,7 @@ $(PKG)_DEPS     := cc gmp mpfr
 $(PKG)_DEPS_$(BUILD) := gmp mpfr
 
 define $(PKG)_UPDATE
-    $(WGET) -q -O- https://ftp.gnu.org/gnu/mpc/ | \
-	grep -oP 'mpc-[0-9.]*[0-9]' | \
-	grep -o '[0-9.]*' | \
-	sort --version-sort | \
-	tail -1
+    $(call GET_LATEST_VERSION, https://ftp.gnu.org/gnu/mpc/)
 endef
 
 define $(PKG)_BUILD_$(BUILD)
--- a/src/mpc.mk
+++ b/src/mpc.mk
@@ -4,21 +4,22 @@ PKG             := mpc
 $(PKG)_WEBSITE  := http://www.multiprecision.org/
 $(PKG)_DESCR    := GNU MPC
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 1.0.2
-$(PKG)_CHECKSUM := b561f54d8a479cee3bc891ee52735f18ff86712ba30f036f8b8537bae380c488
+$(PKG)_VERSION  := 1.1.0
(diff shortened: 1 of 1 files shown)


src/mpc.mk
```diff
@@ -4,21 +4,22 @@ PKG             := mpc
...
 $(PKG)_DEPS_$(BUILD) := gmp mpfr
 
 define $(PKG)_UPDATE
-    $(WGET) -q -O- 'https://gforge.inria.fr/scm/viewvc.php/tags/?root=mpc&sortby=date' | \
-    $(SED) -n 's,.*<a name="\([0-9][^"]*\)".*,\1,p' | \
-    head -1
+    $(WGET) -q -O- https://ftp.gnu.org/gnu/mpc/ | \
+	grep -oP 'mpc-[0-9.]*[0-9]' | \
+	grep -o '[0-9.]*' | \
+	sort --version-sort | \
```
tonytheodore: If you need to write `sort` manually, use `$(SORT)` for portability (it's `gsort` on freebsd/osx).

src/mpc.mk
```diff
@@ -4,21 +4,22 @@ PKG             := mpc
...
+	tail -1
 endef
```
tonytheodore: Try:
```
define $(PKG)_UPDATE
    $(call GET_LATEST_VERSION, https://ftp.gnu.org/gnu/mpc/)
endef
```
and look at the notes in https://github.com/mxe/mxe/blob/master/mxe.updates.mk#L3-L11 for further options (the above works with no other options - this one is the nice case of standard naming conventions).

tonytheodore:
Thanks for the update, it will take some time to merge as I'm trying to get [gcc4](https://github.com/mxe/mxe/commit/b7c8011aa83a0582a778a0c155aecf54c0433d41) through [gcc7](https://github.com/mxe/mxe/pull/2179) to build.

src/mpc.mk
PsychoXIVI: Tested and committed. I will try to make use of it in future, thanks for notice.

PsychoXIVI:
I am an idiot. I accidentally dropped my origin repository, because of GitKraken... :expressionless:

tonytheodore:
gcc4 looks good (there's a couple of unrelated errors) and that's the most likely cause of potential issues. 30 odd hours till gcc5 finishes and I think it's good to merge - the later gcc versions should be fine.

tonytheodore:
Same unrelated issues with gcc5 for `wt` and `itk` - probably their recent updates.

Thanks!

PR #2191, opened 2018-08-23
icu4u: updated to 62.1; fixed version detection due to moved page

PsychoXIVI:
:F


Lord-Kamina:
Have you made sure this builds the correct data file? I made a 61.1 package a while ago (didn't push it because it didn't build against mingw-w64 5.0.4, needed an unstable commit) and one of the biggest issues I had was that for _some reason_, the package as is, when used with newer versions of icu, ended up building only the stub data and not the complete thing (and thus I needed to add some extra hacky steps to make sure the real data file was built and installed/named properly)

PR #2192, opened 2018-08-23
openal: updated; patched new broken CMakeLists

PsychoXIVI:
;F

### DIFF
--- a/src/openal.mk
+++ b/src/openal.mk
@@ -3,8 +3,8 @@
 PKG             := openal
 $(PKG)_WEBSITE  := http://kcat.strangesoft.net/openal.html
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 1.16.0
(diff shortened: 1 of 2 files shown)


src/openal.mk
```diff
@@ -18,7 +18,13 @@ define $(PKG)_UPDATE
 endef
 
 define $(PKG)_BUILD
+    # Update OpenAL cross compilation script with mxe root path
+    $(SED) -i "s|CMAKE_FIND_ROOT_PATH \"/usr/\$${HOST}\"|CMAKE_FIND_ROOT_PATH \"$(PREFIX)/$(TARGET)\"|" $(1)/XCompile.txt
+    cat $(1)/XCompile.txt
```
mabrand: stray line?

PsychoXIVI: Yes, line 23 is unnecessary. I forgot to remove it while testing I think.

mabrand: The package URL doesn't work for me. I found
    http://www.openal-soft.org/openal-releases/
The current version seems to be 1.19.1 Do you want to update the pull request for that?

tonytheodore:
openal has since been updated

PR #2196, opened 2018-08-29
nasm: fix build with GCC >= 8.x.

tehnick:
Tested in Debian unstable.


tonytheodore:
Does it work with gcc<8?

tonytheodore:
It's fine with clang (but also was before):
```
$ gcc -v
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 9.1.0 (clang-902.0.39.2)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Can you push the patch upstream and annotate that in the patch?

tehnick:
Nope.  Their development infrastructure is not transparent for accidental visitor. (At least I failed to find necessary info in a sane amount of time.)

As far as you may see this patch is not my but from Fedora project. Package nasm in Debian repository has even more patches, but they are not required for building on packages in MXE. As we both know package maintainers are always in contact with upstream, so it will be sane to assume that upstream is already aware of this issue and the next stable release will be with appropriate fix for it.

tehnick:
Currently packages `nasm`, `x264`, `ffmpeg`, etc. from MXE are not buildable in Debian unstable/testing and Fedora >= 28. (Using default system compiler.)

tonytheodore:
I see, it's fixed upstream [[1]](http://repo.or.cz/nasm.git/commitdiff/9f45a77f46829e666b35530939b9237cf978d4dc?hp=5eb1838b4d3752fd863d19442943983a2a5ee87c) [[2]](http://repo.or.cz/nasm.git/commitdiff/5eb1838b4d3752fd863d19442943983a2a5ee87c?hp=ef7d18bfbb899fe8e44f4fae8d6a12e2742f69f5) but not released.

Thanks!

PR #2197, opened 2018-08-29
update protobuf

tehnick:


tehnick:
Hmm, it looks that I have build this package with `MXE_PLUGIN_DIRS += plugins/gcc8`. With default compiler it will not compile...

PR #2200, opened 2018-09-01
Update packages and fix some makefiles

jonaski:
It's all tested.
binutils 2.29.1 was the latest I could make work. I ran into some problems loading dll plugins in qt with the newer versions. Maybe someone could look into this.
I'm updating taglib to the latest git revision because of a bug in the current release 1.11.1 (2 years old) corrupting ogg files.


tonytheodore:
Thanks for looking at these updates, a couple of general comments:

  - updating taglib to follow master seems reasonable
  - binutils, gcc, isl, libtool, and qtbase are all manually updated, can you remove the changes from this PR?
  - can you `fixup` the last three commits into their original changes?
  - if you're adding more commits, use WIP in the PR title

other comments on their own commits.

src/cmake.mk
```diff
@@ -3,8 +3,8 @@
 PKG             := cmake
 $(PKG)_WEBSITE  := https://www.cmake.org/
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 3.10.0
-$(PKG)_CHECKSUM := b3345c17609ea0f039960ef470aa099de9942135990930a57c14575aae884987
+$(PKG)_VERSION  := 3.11.4
```
tonytheodore: There's a later cmake 3.12.1 - any reason not to use that?

src/chromaprint.mk
```diff
@@ -4,12 +4,12 @@ PKG             := chromaprint
...
-$(PKG)_VERSION  := 1.1
-$(PKG)_CHECKSUM := 6b14d7ea4964581b73bd3f8038c8857c01e446421c1ae99cbbf64de26b47cd12
-$(PKG)_SUBDIR   := $(PKG)-$($(PKG)_VERSION)
+$(PKG)_VERSION  := 1.4.3
+$(PKG)_CHECKSUM := ea18608b76fb88e0203b7d3e1833fb125ce9bb61efe22c6e169a50c52c457f82
+$(PKG)_SUBDIR   := $(PKG)-v$($(PKG)_VERSION)
 $(PKG)_FILE     := $(PKG)-$($(PKG)_VERSION).tar.gz
-$(PKG)_URL      := https://bitbucket.org/acoustid/chromaprint/downloads/$(PKG)-$($(PKG)_VERSION).tar.gz
-$(PKG)_DEPS     := cc ffmpeg
+$(PKG)_URL      := https://github.com/acoustid/chromaprint/releases/download/v$($(PKG)_VERSION)/$(PKG)-$($(PKG)_VERSION).tar.gz
```
tonytheodore: Can you switch this to using GH_CONF? If it looks tricky, I'll do it later

src/faad2.mk
```diff
@@ -17,7 +17,7 @@ define $(PKG)_UPDATE
 endef
 
 define $(PKG)_BUILD
-    cd '$(1)' && ./configure \
+    cd '$(1)' && autoreconf -fi && ./configure \
```
tonytheodore: Why `autoreconf` if no input file have changed?

src/cmake.mk
jonaski: Yes, cmake in boost fails.
Could not find the following Boost libraries: boost_context

tonytheodore: Okay, I'd suggest leaving cmake out of this PR in that case - later boost has [other issues](https://github.com/mxe/mxe/pull/2043#issuecomment-414269871) that the newest cmake may help with.

jonaski:
Could you merge the individual commits that don't need changes, and then I can just create new PR's for the remaining ones?
If not, I'll just create new PR. In that case do you want every change as individual PR's except those that are just automatic updates to latest versions?

PR #2208, opened 2018-09-17
V6

mabrand:
Update to mingw-w64 6.0


tonytheodore:
Nice - only just released! I've started a test build for the standard `{i686,x86_64}-w64-mingw32.{static,shared}` targets (though I'm not sure if a full build currently works so there may be unrelated issues).

It appears to use a new [Universal CRT](https://blogs.msdn.microsoft.com/vcblog/2015/03/03/introducing-the-universal-crt/) by default so simple programs fail on Win7 and WinXP, though I'm not too concerned about the latter. I hadn't heard about this before so don't know if if it's a simple option to `./configure`, if we need to manipulate gcc's spec files, or just a corresponding update to binutils (`file` reports different arch to `depends.exe`).

A related thing I'm working on is proper support for (potentially multiple) gcc versions via target specifiers e.g:
```
MXE_TARGETS=\
  x86_64-w64-mingw32.static.gcc8
  i686-w64-mingw32.static.gcc7
```
We could do something like 64-bit Win10 and 32-bit Win≤8.1 with:
```
MXE_TARGETS=\
  x86_64-w64-mingw32v6.static.gcc8
  i686-w64-mingw32.static.gcc7
```
So people could easily choose runtime and compiler without waiting for mxe defaults to change or using plugins and/or git workflows.

I'm not sure of the exact syntax of `*-mingw32v6*` but the general idea is to decouple the large subset of what is possible with mxe from what we can test and decide is stable in a reasonable timeframe. It makes it easier for early-adopters to try new combinations and easier for us to replicate issues. What do you think?

src/libmysqlclient-1-fixes.patch
```diff
@@ -471,3 +471,25 @@ index 1111111..2222222 100644
...
+--- a/mysys/my_gethwaddr.c
++++ b/mysys/my_gethwaddr.c
+@@ -138,8 +138,8 @@ my_bool my_gethwaddr(uchar *to)
+ #endif
+ 
+ // mprapi.h uses these undefined types
+-#define CERT_NAME_BLOB int
+-#define CRYPT_HASH_BLOB int
++//#define CERT_NAME_BLOB int
++//#define CRYPT_HASH_BLOB int
```
tonytheodore: Can you wrap this in `#if defined (__MINGW64_VERSION_MAJOR) && __MINGW64_VERSION_MAJOR >= 6`?

src/qtmultimedia-1-fixes.patch
```diff
@@ -0,0 +1,30 @@
...
+
+diff --git a/src/plugins/common/evr/evrdefs.h b/src/plugins/common/evr/evrdefs.h
+index 1111111..2222222 100644
+--- a/src/plugins/common/evr/evrdefs.h
++++ b/src/plugins/common/evr/evrdefs.h
+@@ -83,6 +83,7 @@ HRESULT WINAPI Direct3DCreate9Ex(UINT SDKVersion, IDirect3D9Ex**);
+ #define MF_E_TRANSFORM_NEED_MORE_INPUT ((HRESULT)0xC00D6D72L)
+ #endif
+ 
++#if 0
```
tonytheodore: #if defined (__MINGW64_VERSION_MAJOR) && __MINGW64_VERSION_MAJOR >= 6

src/libmysqlclient-1-fixes.patch
tonytheodore: Oops, negate that check (or test for the defs directly).

src/qtmultimedia-1-fixes.patch
tonytheodore: Oops again, negate the check.

jonaski:
qtmultimedia fails to build (shared):

```
.obj/release/evrcustompresenter.o:evrcustompresenter.cpp:(.rdata$.refptr.IID_IMFVideoPresenter[.refptr.IID_IMFVideoPresenter]+0x0): undefined reference to `IID_IMFVideoPresenter'
.obj/release/evrcustompresenter.o:evrcustompresenter.cpp:(.rdata$.refptr.IID_IMFVideoDeviceID[.refptr.IID_IMFVideoDeviceID]+0x0): undefined reference to `IID_IMFVideoDeviceID'
.obj/release/evrcustompresenter.o:evrcustompresenter.cpp:(.rdata$.refptr.IID_IMFTopologyServiceLookupClient[.refptr.IID_IMFTopologyServiceLookupClient]+0x0): undefined reference to `IID_IMFTopologyServiceLookupClient'

```

mabrand:
Made improvements suggested by tonytheodore and fixed qtmultimedia linking problem found by jonaski.

mabrand:
You're right about itk shared being an existing issue. I can confirm it fails on both i686-w64-mingw32.shared and itk_i686-w64-mingw32.shared on Mingw-w64 v5 too.

Lord-Kamina:
Is this ever gonna be merged? Asking for a friend. 😂

mabrand:
@tonytheodore: What do you think?

tonytheodore:
I haven't had a chance to look further into the new `ucrt` but think this is good to merge if we use `--with-default-msvcrt=msvcrt` as the default.

The new config options should help with the longer term goal of allowing people to choose the runtime without linking it to a specific mingw-w64 version.

mabrand:
Done. Thanks Tony, I was wondering whether to add that option. Feel free to merge if you are happy with it.

tonytheodore:
Thanks

Lord-Kamina:
Awesome; thanks!

PR #2211, opened 2018-09-23
Add shaderc

Akaricchi:
Required for taisei-project/taisei#148. Forgive me if this is crude, I've tried my best, and I'm sick and tired of fighting with these broken and braindead build systems already. I might improve it later, but hey, it installs a static library, so that's good enough for now. Here are a couple of known issues:

 * No pkg-config files, because Khronos is oblivious and Google doesn't give a crap (google/shaderc#392). Projects that depend on those libraries don't expect to find them via pkg-config as a sad result of that, so I made no effort to generate one.
 * Neither shaderc nor glslang have a soversion, probably for the same reasons they don't have .pc files.
 * shaderc has different library names for static and shared builds, because Google is fucking evil I guess.
 * I have no idea how I'm supposed to handle versions of this bloody mess.
 * Shared targets are not tested at all.
 * Patch files were generated manually via git format-patch, because patch-tool-mxe refused to cooperate.

### DIFF
--- a/src/glslang-src.mk
+++ b/src/glslang-src.mk
@@ -1,0 +1,12 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG_BASENAME    := glslang
+PKG             := $(PKG_BASENAME)-src
+$(PKG)_WEBSITE  := https://github.com/KhronosGroup/glslang
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 7.9.2888
+$(PKG)_SUBDIR   := $(PKG_BASENAME)-$($(PKG)_VERSION)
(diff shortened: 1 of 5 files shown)


Akaricchi:
glslang and spirv-headers fail with gcc8 apparently, will investigate tomorrow I guess.

Akaricchi:
Nevermind the previous comment, it actually builds fine on this branch. The build failures I was having are likely related to #2108, which I had merged locally.

Akaricchi:
Update: I've largely overhauled this. To make static linking actually work, I had to resort to compiling shaderc with vendored dependencies. Apparently that is how google wants it to work, and their build system is too broken to do it the sane way. I've converted the dependencies into source-only packages to achieve that.

PR #2215, opened 2018-09-26
new package: qtifw

muellni:
This installs qt installer framework in mxe. It only makes sense for static targets and the native tools depend on a native Qt installation.


mabrand:
Thanks. Looks like a nice addition.

Do the patches come from upstream? If so, please include the git header. If not, it would be good to submit them upstream and include a reference to the pull request.

muellni:
These patches come from me, I didn't submit them yet. I'm currently trying to figure out how to get them into Qts gerrit.

PR #2219, opened 2018-09-28
Sharing Kdenlive Windows build process

vpinon:
Hello,

I'm generating Kdenlive (KDE video editor) windows binaries for almost 2 years with MXE.
I have been waiting for long to share my work with you, mainly because I wanted to clean & upstream most of the patches.
Anyway here is the set I have been using again today... open to your wise comments :)

One thing that still bothers me is that some KDE frameworks generate binaries that are then called later in the process, the simplest solution I have found was to call my system installed ones :( 

And again, many many thanks for developing MXE, which is really a fantastic build/packaging system!

Vincent

### DIFF
--- a/plugins/apps/kdenlive-distribute.sh
+++ b/plugins/apps/kdenlive-distribute.sh
@@ -1,0 +1,50 @@
+#!/bin/bash -xe
+# run from mxe root dir after "make kdenlive"
+# > plugins/apps/kdenlive-distribute.sh
+MXE_DIR=$PWD
+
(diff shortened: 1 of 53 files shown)


scorpio810:
Hello,
is possible to ￼add KDE Frameworks 5 in MXE  plugins?
For now I use Vincent MXE fork.

scorpio810:
I added Vincent plugin in my local MXE, MXE build nice Qt 5 and KF5 Framework, but I have linking error at the end on compil, what wrong?

[gist:279f6586b109c78c9da111b328060505](url)

PR #2240, opened 2018-11-15
github release macro: add variable for archive extensions

tonytheodore:
From: https://github.com/mxe/mxe/commit/ac697f5bac7e7590fe1363d1a52113ae6ba442a9#r31310663

It's not easy to reliably parse/guess the extension for manually
uploaded archives using GitHub releases. Even if the naming convention
matches GitHub's automated rules, there are two problems:

  - mxe version may be on the ["next" page](https://github.com/fribidi/fribidi/releases?after=0.19.4)
  - using API may hit rate limiting if it's a live check at build time

This adds an "archive extension" variable to `GH_CONF` for the simple
cases where all other naming follows GitHub conventions.

### DIFF
--- a/mxe.github.mk
+++ b/mxe.github.mk
@@ -10,9 +10,10 @@
 #   Release:
 #     Manually uploaded distribution tarballs, especially useful for
 #     autotools packages with generated sources. No universal convention,
-#     but generally:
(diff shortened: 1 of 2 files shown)


tonytheodore:
Running `make check-gh-conf` now.

PR #2247, opened 2018-12-02
Update clementine

jonaski:
- Update Clementine to the latest git revisions
- Remove obsolete patching
- Remove libechonest, support is dropped in Clementine and the API is closed.
- Add a Qt 4 version of the qjson library: qjson-qt4 to still build the qt4 version of Clementine from master


pavelvat:
1)For unify the patch use mxe/tools/patch-tool-mxe, see https://github.com/mxe/mxe/pull/1277#issuecomment-257083463
2)$(PKG)_DEPS  must contain list of dependencies in alphabetical order
3)"qt5" is meta package see https://mxe.cc/build-matrix.html use only really needed qt5 modules
4)$(PKG)_DEPS automatically calculates dependencies, for example, you do not need to list all packages related with gstreamer

jonaski:
Fixed

pavelvat:
1)cc is the exception and should always be first.
2)package gst-plugins-bad already contain package gst-plugins-base in $(PKG)_DEPS, so there is no need to explicitly list it.
3)package liblastfm already contain package qtbase in $(PKG)_DEPS, so there is no need to explicitly list it.

PR #2249, opened 2018-12-06
ghostscript 9.26

darealshinji:
Update Ghostscript to version 9.26, always hide libgs symbols, use local lcms fork (lcms2mt).


tonytheodore:
Thanks!

Issue #2070, opened 2018-03-06
QtWebKit


aquiles2k:
Hi,
im currently in the process of updateing our application from ancient qt5.6.0 to newest qt 5.10.1. As usual its qtwebkit which gives me some headaches.

As you perhaps know theres now a qtwebkit version by Konstantin Tokarev which gets actively developed, tries to stay in sync with upstream webkit and is supported by qt.

See: 
http://code.qt.io/cgit/qt/qtwebkit.git/
https://github.com/annulen/webkit
http://download.qt.io/snapshots/ci/qtwebkit/5.212/latest/src/submodules/
http://download.qt.io/snapshots/ci/qtwebkit/5.212/latest/qtwebkit/
etc.

Is there any chance we can integrate this in mxe?

I tried to use their precompiled binaries together with our mxe compiled libs but unfortunately they are missing symbols like __emutls_v._ZSt11__once_call which is perhaps because we compile with .posix.dw2.

I could try to use the old webkit from qtwebkit.mk but i see its stuck at qt5.9.1 so i doubt that will be compatible with 5.10.x

Are there any plans how to go on with the webkit package?

mabrand:
The last time I checked, qtwebkit 5.9.1 at least built successfully with qt 5.10.1. Haven't actually tried to run anything though.

If you would be willing to put together a patch so that qtwebkit.mk will follow Konstantin's version, that would be great. It's probably not that that hard to do. For examples, look at some of the other .mk files in src that get their sources from github.

aquiles2k:
Im not sure if im experienced enough with mxe to get such a library to build :)

Anyway i made some attempt but didnt get very far. I tried to look at the msys2 build over at
https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-qtwebkit/PKGBUILD

But their dependencies and my local error cmake error seem to indicate that ruby is required. As far as i can see mxe doesnt provide its own ruby?

tonytheodore:
> As far as i can see mxe doesnt provide its own ruby?

No, ruby (and python/perl) would be too hard to manage across systems - better to use the one from your package manager.

tonytheodore:
Try https://github.com/mxe/mxe/commit/473f697ba37e2cfcbcd417bd6992e4f68e8cea66. If it doesn't work, you may have to add `-DCMAKE_SYSTEM_PROCESSOR=$(PROCESSOR)` to your cmake invocation.

aquiles2k:
[...]

PR #2263, opened 2019-01-04
Posix threads

tehnick:
See: https://github.com/mxe/mxe/issues/2258#issuecomment-450441252

PR #2237, opened 2018-11-14
[WIP] Update QtWebKit to 5.212

dirkhh:
All of the work was done by @aquiles2k - all I did was fix the install prefix
and create the pull request so that people can start discussing this proposed
change.

This has only been tested for shared builds and requires posix threads.
x86_64-w64-mingw32.shared.posix.dw2 and i686-w64-mingw32.shared.posix.dw2
appear to work.

Issue #2160, opened 2018-06-26
QtWebkit 5.9.1 is broken after update to Qt 5.11.x


tehnick:
diff shows:
```diff
@@ -3133,35 +3315,16 @@
 qt5/include/QtWebKit/5.9.1/QtWebKit/private/util.h
 qt5/include/QtWebKit/QtWebKit
 qt5/include/QtWebKit/QtWebKitDepends
-qt5/include/QtWebKit/QtWebKitVersion
 qt5/include/QtWebKit/qtwebkitversion.h
-qt5/include/QtWebKit/QWebDatabase
 qt5/include/QtWebKit/qwebdatabase.h
-qt5/include/QtWebKit/QWebElement
-qt5/include/QtWebKit/QWebElementCollection
 qt5/include/QtWebKit/qwebelement.h
-qt5/include/QtWebKit/QWebFullScreenVideoHandler
-qt5/include/QtWebKit/QWebHapticFeedbackPlayer
-qt5/include/QtWebKit/QWebHistory
 qt5/include/QtWebKit/qwebhistory.h
-qt5/include/QtWebKit/QWebHistoryInterface
 qt5/include/QtWebKit/qwebhistoryinterface.h
-qt5/include/QtWebKit/QWebHistoryItem
 qt5/include/QtWebKit/qwebkitglobal.h
-qt5/include/QtWebKit/QWebKitPlatformPlugin
 qt5/include/QtWebKit/qwebkitplatformplugin.h
-qt5/include/QtWebKit/QWebNotificationData
-qt5/include/QtWebKit/QWebNotificationPresenter
-qt5/include/QtWebKit/QWebPluginFactory
 qt5/include/QtWebKit/qwebpluginfactory.h
-qt5/include/QtWebKit/QWebSecurityOrigin
 qt5/include/QtWebKit/qwebsecurityorigin.h
-qt5/include/QtWebKit/QWebSelectData
-qt5/include/QtWebKit/QWebSelectMethod
-qt5/include/QtWebKit/QWebSettings
 qt5/include/QtWebKit/qwebsettings.h
-qt5/include/QtWebKit/QWebSpellChecker
-qt5/include/QtWebKit/QWebTouchModifier
 qt5/include/QtWebKitWidgets
 qt5/include/QtWebKitWidgets/5.9.1
 qt5/include/QtWebKitWidgets/5.9.1/QtWebKitWidgets
@@ -3170,21 +3333,43 @@
[...]

PR #2268, opened 2019-01-09
Update QtWebKit to 5.212 (final version)

tehnick:
After merging of https://github.com/mxe/mxe/pull/2263 we can finally update `qtwebkit` package and close issues https://github.com/mxe/mxe/issues/2070 and https://github.com/mxe/mxe/issues/2160.

Build with proposed changes is successfully tested for:
```
MXE_TARGETS := i686-w64-mingw32.shared x86_64-w64-mingw32.shared
```
As for usage with real applications I have tested only shared versions of libraries (both 32 and 64 bit) and only with program Psi+. And only during one month or so. 

As was discussed in https://github.com/mxe/mxe/issues/2070#issuecomment-438858029 (and below in that thread) I have enabled all possible features.

As you may see in the package, QtWebKit is built with enabled `ENABLE_VIDEO` option and with QtMultimedia backend for this by default. I explicitly added all three available options (`USE_GSTREAMER`, `USE_MEDIA_FOUNDATION`, `USE_QT_MULTIMEDIA`) for making them more noticeable for end users.

For testing purpose I have successfully built QtWebKit with GStreamer backend but failed to bring it show embedded video from YouTube. (It does not crash but just scold for required video format.)

As for build with [Media Foundation](https://en.wikipedia.org/wiki/Media_Foundation) I have tried to fix build with it (see the end of patch), but have not finished the work, because of lack of free time. This work may be delayed for future updates or finished in upstream (if project will become alive).

I believe that this PR is ready for merging, but any comments are welcome!

PS: Please close incomplete PR https://github.com/mxe/mxe/pull/2237.

### DIFF
--- a/src/qtwebkit-1-fixes.patch
+++ b/src/qtwebkit-1-fixes.patch
@@ -2,94 +2,287 @@ This file is part of MXE. See LICENSE.md for licensing information.
 
 Contains ad hoc patches for cross building.
 
-From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
-From: Tony Theodore <tonyt@logyst.com>
-Date: Wed, 18 Jan 2017 19:38:56 +1100
-Subject: [PATCH 1/3] fix msys build
-
-Taken from:
-https://github.com/Alexpux/Qt-builds/blob/master/patches/qt/5.0.x/qt-5.0.0-fix-build-under-msys.patch
-
-diff --git a/Source/WebCore/DerivedSources.pri b/Source/WebCore/DerivedSources.pri
-index 1111111..2222222 100644
---- a/Source/WebCore/DerivedSources.pri
-+++ b/Source/WebCore/DerivedSources.pri
-@@ -752,7 +752,7 @@ IDL_ATTRIBUTES_FILE = $$PWD/bindings/scripts/IDLAttributes.txt
- preprocessIdls.input = IDL_ATTRIBUTES_FILE
- preprocessIdls.script = $$PREPROCESS_IDLS_SCRIPT
- # FIXME : We need to use only perl at some point.
--win_cmd_shell: preprocessIdls.commands = type nul > $$IDL_FILES_TMP $$EOC
-+win_cmd_shell: preprocessIdls.commands = cat /dev/null > $$IDL_FILES_TMP $$EOC
- else: preprocessIdls.commands = cat /dev/null > $$IDL_FILES_TMP $$EOC
- for(binding, IDL_BINDINGS) {
-     # We need "$$binding" instead of "$$binding ", because Windows' echo writes trailing whitespaces. (http://wkb.ug/88304)
-
-From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
-From: Martchus <martchus@gmx.net>
-Date: Sun, 25 Sep 2016 22:21:54 +0200
-Subject: [PATCH 2/3] Include intrin.h for declaration of _mm_mfence
-
-taken from:
-https://aur.archlinux.org/cgit/aur.git/tree/0007-Include-intrin.h-for-declaration-of-_mm_mfence.patch?h=mingw-w64-qt5-webkit
-
-diff --git a/Source/WTF/wtf/Atomics.h b/Source/WTF/wtf/Atomics.h
-index 1111111..2222222 100644
---- a/Source/WTF/wtf/Atomics.h
-+++ b/Source/WTF/wtf/Atomics.h
-@@ -68,6 +68,8 @@
- #elif !COMPILER(GCC)
- extern "C" void _ReadWriteBarrier(void);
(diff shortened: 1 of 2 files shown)


tehnick:
Also unlike PR #2237 the simple test program (qtwebkit-test) is still used for testing the build with QtWebKit libraries.

tehnick:
BTW, is it any way to use ccache in MXE? It was a pain to rebuild qtwebkit each time after updating of the patch. And such clean rebuild from zero is almost useless, just a stupid heating of atmosphere...

tehnick:
> As for usage with real applications I have tested only shared versions of libraries (both 32 and 64 bit) and only with program Psi+.

One more note here: program and all libraries were built with gcc 6.x (from related MXE plugin), because 32 bit version of QtWebKit build with gcc 8.x had some unpredictable crashes (64 bit version worked fine for me).

As for older versions of gcc, I prefer not use them in our projects...

dirkhh:
#2237 was only intended to document a first step towards making this work. Thanks @tehnick for working on this!

tonytheodore:
Thanks for all your work, that looks quite comprehensive and will be a welcome update. I've started a build with the default gcc5 - do you know if there are new requirements like ruby?

Is `-fpermissive` really needed or just a way to cover new compiler warnings? If it's the latter, it can stay in as I have some wip around handling multiple gcc versions that makes this easier.

> BTW, is it any way to use ccache in MXE?

I haven't used it before and don't know much about it.

> It was a pain to rebuild qtwebkit each time after updating of the patch.

What I'd normally do is something like:
```
make init-git-qtwebkit
make qtwebkit qtwebkit_SOURCE_TREE=gits/qtwebkit-x..y.x
```
This will give you a build dir in tmp-qtwebkit-... and git controlled source tree in gits/qtwebkit-x..y.x. You can go make changes and commit in the source, incrementally build in tmp, and when you're done:
```
make export-patch-qtwebkit
make qtwebkit #as a final test
```
You could always set `qtwebkit_SOURCE_TREE` to some local git repo and produce a `git format-patch` style diff from there that can be standardised for mxe with:
```
make refresh-patch-qtwebkit
```

tehnick:
> I've started a build with the default gcc5 - do you know if there are new requirements like ruby?

I have `ruby` in my system and I do not know is it necessary for build. Here is one of my build logs with gcc5: [gist.github.com](https://gist.github.com/tehnick/cbc31abc57cfb7cce9a5fb31d74b38ed)

> Is `-fpermissive` really needed or just a way to cover new compiler warnings? If it's the latter, it can stay in as I have some wip around handling multiple gcc versions that makes this easier.

I have not tested build with gcc5 without it. For gcc6 and gcc8 it is required for "fixing" build for `i686-w64-mingw32.shared.sjlj` target. IIRC for `i686-w64-mingw32.shared.dw2` target library may be built without `-fpermissive`. But I am not sure. Build for `x86_64-w64-mingw32.shared.seh` was sucessfull without this option, build for `x86_64-w64-mingw32.shared.sjlj` I have not checked due to lack of interest.

tehnick:
>> BTW, is it any way to use ccache in MXE?
> I haven't used it before and don't know much about it.

Oh, you should try it. It is extremely convenient for development! In cases when one project is re-build many and many times.

tehnick:
> What I'd normally do is something like:
> ...

Hmm, this may be combined with my own practice to build using `make -k ...` and fix as much build errors as possible for each change of patch. Thanks!

tehnick:
> Here is one of my build logs with gcc5: [gist.github.com](https://gist.github.com/tehnick/cbc31abc57cfb7cce9a5fb31d74b38ed)

This file has been truncated in Web UI, but you may open raw file using related link.

tonytheodore:
It seems ruby is already a requirement, and the build succeeds with gcc5 (with or without `-fpermissive`).

This is [rebased and merged now, along with a retry for build failure due to resource exhaustion and converting the patch to format-patch style](https://github.com/mxe/mxe/compare/698001de...7148391b).

Thanks!

tehnick:
> build succeeds with gcc5 (with or without `-fpermissive`).

Yep, at night I have tested build without `-fpermissive` using gcc5 and all were built successfully. But for gcc6 and later this issue remains.

> along with a retry for build failure due to resource exhaustion

Are you talking about command below?
```
$(MAKE) -C '$(BUILD_DIR)' -j '$(JOBS)' VERBOSE=1 || $(MAKE) -C '$(BUILD_DIR)' -j '$(JOBS)' VERBOSE=1
```
Here you just try to build program second time when first attempt is failed. But how this should solve the problem with resource exhaustion?

As far as I remember QtWebKit always required a lot of memory for build. But not at compilation stage, but on linking one. And your workaround will not fix build if this is exactly that problem.

(Few years ago I had to enable swap file for building QtWebKit package successfully, but now I have enough RAM and could not check build with smaller memory size without using of virtual machine or such like.)

tehnick:
@tonytheodore I see. Thanks for the detailed answer!

PR #2269, opened 2019-01-10
add support for lzip archives

mabrand:
.tar.lz (lzip) archives seem to be increasing in popularity. As far as I can tell, the lzma, xz, and 7z utilities do not support this format, so lzip is required to decompress them.

Could the knowledgeable ones please check the distro-specific package names in this commit?

In the Makefile macro UNPACK_ARCHIVE, plain "tar xf" could be used more generally, since modern tar understands many compressed formats, including lzip. I imagine the support in tar also depends on lzip being installed. Are there still good reasons for explicitly invoking specific decompression tools based on file extension?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ WGET        = $(WGET_TOOL) --user-agent='$(or $($(1)_UA),$(DEFAULT_UA))'
 
 REQUIREMENTS := autoconf automake autopoint bash bison bzip2 flex \
                 $(BUILD_CC) $(BUILD_CXX) gperf intltoolize $(LIBTOOL) \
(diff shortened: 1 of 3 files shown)


tonytheodore:
Thanks, I've checked through the [package searches](https://github.com/mxe/mxe/commit/698001de0249ac72d03e83c4195549ef9347b5f5) and the names look good.

The default version of tar on darwin is bsdtar (libarchive) from 2010 and can't auto-detect *.lzma or *.lz. Even the macports gnutar version doesn't yet have support for *.lz. I might have a look at building our own recent version.

tonytheodore:
Seems like a worthwhile change given some potential [issues with xz](https://www.nongnu.org/lzip/xz_inadequate.html)

Issue #2272, opened 2019-01-20
Using ccache


tonytheodore:
@tehnick [wrote](https://github.com/mxe/mxe/pull/2268#issuecomment-452884914):
> BTW, is it any way to use ccache in MXE?

Using the [symlink method](https://ccache.samba.org/manual/latest.html#_run_modes) and adjusting the cmake toolchain file results in nice speedups for qtwebkit:
```
no cache     57m15.298s
cache 00     72m13.544s
cache 01     12m20.226s
cache 02     12m26.279s
cache 03     12m22.409s
```
Looks like a 20-25% penalty for the first cache run, then speedups just under 500%!

I think we'd use the `content` [compiler_check config](https://ccache.samba.org/manual/latest.html#_configuration_settings) but otherwise the defaults are very easy to get started with. The cache stats:
```
$ ccache -s
cache directory                     /Users/tonyt/.ccache
primary config                      /Users/tonyt/.ccache/ccache.conf
secondary config      (readonly)    /opt/local/etc/ccache.conf
stats updated                       Sun Jan 20 00:49:31 2019
cache hit (direct)                 10713
cache hit (preprocessed)              42
cache miss                          3585
cache hit rate                     75.00 %
called for link                      176
called for preprocessing            6608
compile failed                         4
preprocessor error                     4
no input file                         81
cleanups performed                     0
files in cache                      7388
cache size                         194.7 MB
max cache size                       5.0 GB
```
qtwebkit is the largest package so we may not even need to tweak the compression or cache size. I'm certainly going to look further into adding it to mxe.

tehnick:
> Looks like a 20-25% penalty for the first cache run, then speedups just under 500%!

Very interesting results. Not such impressive as at a build of native libraries and programs in my system. But nevertheless the speed up is huge.

> I'm certainly going to look further into adding it to mxe.

Excellent!

tonytheodore:
Progress so far is in #2273 though I'm thinking of removing support for a system ccache and only using our own version/config.

tehnick:
> Progress so far is in #2273 though I'm thinking of removing support for a system ccache and only using our own version/config.

Yes, this is a good idea. Users would not want to mix cache for system libraries and programs and cache for MXE libraries.

tonytheodore:
I'm happy with #2273 for now, thanks for the tip @tehnick - there's no going back once you start using it.

PR #2273, opened 2019-01-20
Add ccache and enable by default

tonytheodore:
See #2272


tonytheodore:
[Building qt5 & qtwebkit from scratch](https://gist.github.com/tonytheodore/7104a1d14413a8f2c7cdfff5db39dfd9):

```
no cache: 126m13.904s
cache 00: 150m29.873s
cache 01: 45m3.009s
cache 02: 45m4.558s
```

less impressive than some individual packages, but there's a lot of non-compile steps going on. I think it's especially useful when incrementally working on a single package. cmake encodes full paths so it's easy to toggle outside the mxe build environment, other build systems need to set $PATH.

I think this is good to merge as a first step. @tehnick @mabrand can you think of any reasons we shouldn't enable it by default?

tehnick:
> can you think of any reasons we shouldn't enable it by default?

I do not see such reasons.

tehnick:
> I think it's especially useful when incrementally working on a single package.

So do I. More over I see this option as very useful for development of MXE itself.

For example, my build environments for hobby tasks look like this:
```
$ du -hs /opt/mxe*
0       /opt/mxe
3,0G    /opt/mxe-hobby__2019-01-10
4,1G    /opt/mxe-hobby__2019-02-04
3,2G    /opt/mxe-hobby__2019-02-25
3,2G    /opt/mxe-hobby__2019-03-01
4,2G    /opt/mxe-hobby__devel
1,7G    /opt/mxe-hobby-winxp__2018-03-03
1,9G    /opt/mxe-hobby-winxp__2018-03-21
7,0G    /opt/mxe-master
```
And for work tasks:
```
$ du -hs /opt/mxe*
0       /opt/mxe
4,8G    /opt/mxe__2018-12-21
5,4G    /opt/mxe__2019-01-11
5,4G    /opt/mxe__2019-02-12
1,2G    /opt/mxe__master
2,0G    /opt/mxe-winxp__2018-03-14
6,1G    /opt/mxe__work
```
Do not pay an attention to symlinks (`/opt/mxe`).

As you see I have constant directories for updating and testing of MXE:
```
/opt/mxe-hobby__devel
/opt/mxe-master
/opt/mxe__master
/opt/mxe__work
```
They are re-built quite often. For example, when I update packages myself and prepare PRs for upstream (using separate git branches, so almost all paths are the same). Previously I had more permanent directories, like:
```
/opt/mxe__gcc6
/opt/mxe__gcc8
etc.
```
But now I do not have time for such experiments. And since I have found a lot of problems with gcc8 I use gcc6 in all my working branches.

As for directories with dates in their names: they are used by my [simple build server](https://github.com/sibuserv) for hobby and at work (see [special git tags](https://github.com/sibuserv/mxe/releases) in my fork of MXE). And I completely do not care if they will be built a bit longer.

tehnick:
Hope my example will be useful for understanding how different people use MXE.

tonytheodore:
Interesting, there's a few different scenarios. I have some WIP that will make some of those scenarios a little easier to manage:
  - target specified gcc versions e.g. x86_64-w64-mingw32.shared.gcc6
  - content checksums to avoid rebuilds when changing branches

I haven't tested sharing the MXE_CCACHE_DIR across multiple mxe directories, it should work in principle.

tonytheodore:
I've added an `MXE_CCACHE_BASE_DIR` variable which sounds like it will work [according to the docs](https://ccache.samba.org/manual/latest.html#_compiling_in_different_directories), but haven't tested it. In combination with `MXE_CCACHE_DIR` and the user config in `MXE_CCACHE_DIR/ccache.conf`, it should be possible to tweak this to handle multiple directories.

I think this is ready to go.

PR #2283, opened 2019-02-20
update vtk to 8.2

muellni:
I had to switch to VTK's libharu fork, since they added some functionality. 
Build has been tested for `i686-w64-mingw32.static`, `i686-w64-mingw32.static.posix` and `i686-w64-mingw32.shared.posix`.


tonytheodore:
Thanks, I'm testing the downstream pkgs now.

Note that since #2263, posix threads are the default.

tonytheodore:
It breaks `pcl` but the [latest version](https://github.com/mxe/mxe/commit/cbf35c9278c37a0f51cea1099606764da774fe5d) succeeds.

Thanks!

PR #2290, opened 2019-03-03
hunspell: update 1.6.1 --> 1.7.0

stloeffler:


tonytheodore:
Ideally, `-DHUNSPELL_STATIC` should be specified with `Cflags.private` in the [*.pc.in file](https://github.com/hunspell/hunspell/blob/master/hunspell.pc.in). If you have time to do an upstream PR, that would be great.

Thanks!

stloeffler:
Thanks. Reported upstream as https://github.com/hunspell/hunspell/pull/623

PR #2302, opened 2019-03-10
WIP: Add boost 1.66 as plugin

fiesh:
In reference to https://github.com/mxe/mxe/pull/2043, here's an attempt at making boost 1.66 available as a plugin.  (Boost 1.69, the most current version, would require a newer CMake version.)

There are two isses yet to be resolved:
* The tests are commented in `boost-overlay.mk` because I couldn't figure out how to make it *not* build the test in `src` but the one in `plugins/boost_1_66_0`.
* Only building the two static targets worked for me.  For shared targets, I got an internal compiler error.  This seems like a separate gcc8 bug, and I don't know how it would affect the state of this plugin.

Happy for feedback!

### DIFF
--- a/plugins/boost_1_66_0/boost-overlay.mk
+++ b/plugins/boost_1_66_0/boost-overlay.mk
@@ -1,0 +1,126 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := boost
+$(PKG)_WEBSITE  := https://www.boost.org/
+$(PKG)_DESCR    := Boost C++ Library
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 1.66.0
+$(PKG)_CHECKSUM := 5721818253e6a0989583192f96782c4a98eb6204965316df9f5ad75819225ca9
+$(PKG)_SUBDIR   := $(PKG)_$(subst .,_,$($(PKG)_VERSION))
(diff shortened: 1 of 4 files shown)


tonytheodore:
Thanks for looking at this, plugins can be useful but also have some complications (like the test you mentioned above).

I'm inclined to develop something like "first class" support for different boost versions, something along the lines of:
```
make DEFAULT_BOOST=boost166
# and/or
make MXE_TARGETS=i686-w64-mingw32.static.gcc8.boost166
```
then I can ensure that tests/patches/updates all work as expected internally and adding new versions won't be a chore. There's also the `Boost_ADDITIONAL_VERSIONS` cmake variable that can be used for newer boost versions while our cmake catches up.

Let me think about it, I have similar logic working for gcc so it shouldn't be too hard.

tonytheodore:
Actually, now I'm leaning towards supporting multiple boost versions with:
```
make MXE_USE_BOOST='boost160 boost166 boost167 boost168 boost169'
```
where the first version specified is used for downstream deps.

fiesh:
Is there a way to install multiple boost versions simultaneously?  I have never done it.

Also, I'd vote for a different versioning scheme since it'll become confusing once boost's minor version hits 100.

fiesh:
Ping.  Any news per chance?

orgads:
This can be closed.

Issue #1038, opened 2015-12-06
package zziplib is broken (Wheezy)


starius:
Commit 99af3f72565c0e75e2115d8233e49d9b58a551f6
Log: https://gist.github.com/starius/9423225b0714b46c056b

PR #2307, opened 2019-03-23
Enable mmap for zziplib

Hopper262:
This re-enables mmap functionality without triggering the compilation error in issue #1038.


tonytheodore:
Thanks!

PR #2308, opened 2019-03-24
Update zziplib: 0.13.62 -> 0.13.69

Hopper262:
* The project has moved to GitHub, as noted on the [old website](https://zziplib.sourceforge.io/). This PR switches to `$(PKG)_GH_CONF`.
* The existing mmap patch has been accepted to upstream and is no longer needed.
* The new patch (fixing a missing header and cast) was accepted to upstream 10 months ago, but there have been no new releases.


tonytheodore:
Can you fix the commit message (0.12.69) and add `mman-win32` to the DEPS?

Hopper262:
With #2307 in place, zziplib should not rely on `<sys/mman.h>` or even use it if present. The code path directly uses Win32 native APIs from `<winbase.h>` and `<memoryapi.h>` instead. So, `mman-win32` should not be a necessary dependency.

The configure script still detects `<sys/mman.h>` and claims that implementation will be used, because I did not patch the configure script to reflect the changes to `zzip/_mman.h`. I can provide a patch for `configure.ac` if that's helpful.

If there's anything else I can do, just ask! Thanks for all your efforts on MXE.

tonytheodore:
> The code path directly uses Win32 native APIs

Thanks for clarifying.

> I can provide a patch for configure.ac if that's helpful.

It's probably not necessary, maybe propose it upstream and we'll pick it up in the next update.

PR #2319, opened 2019-03-31
upgrade fdk-aac to 2.0.0

jcable:
Hi, this is a simple version upgrade. 2.0.0 includes decoding only support for the Digital Radio Mondiale flavours of AAC-HEv2 as well as decoding only support for the new xHE-AAC (USAC) codec. It's needed by a number of open source projects including drm.sourceforge.net which I'm working on a release and will try and package once it is working with MXE.


tonytheodore:
Thanks, this breaks downstream package `gst-plugins-bad` and may need to be updated together.

jcable:
If I get a chance I’ll look at that.

tonytheodore:
Unfortunately, `gst-plugins-bad` also indirectly depends on broken `gnutls` so you'll find it hard to test.

I'll merge this now and look at those separately - thanks!

PR #2326, opened 2019-04-29
secure URLs

vszakats:


src/isl.mk
```diff
@@ -1,7 +1,7 @@
 # This file is part of MXE. See LICENSE.md for licensing information.
 
 PKG             := isl
-$(PKG)_WEBSITE  := http://isl.gforge.inria.fr/
+$(PKG)_WEBSITE  := https://isl.gforge.inria.fr/
```
tonytheodore: From memory, they once used a self-signed cert and now seem to be redirecting to back to http - it's not really https.

vszakats: Yes, it's a redirect at the moment. Hopefully they will fix it someday.

tonytheodore:
Thanks

PR #2328, opened 2019-05-04
Switch to virtual-hosted style URLs for S3

vszakats:
Ref: https://forums.aws.amazon.com/ann.jspa?annID=6776

  * install .pc file,

Thanks!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ DEFAULT_MAX_JOBS   := 6
 PRINTF_COL_1_WIDTH := 13
 SOURCEFORGE_MIRROR := downloads.sourceforge.net
 MXE_MIRROR         := https://mirror.mxe.cc/pkg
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -19,7 +19,7 @@ DEFAULT_MAX_JOBS   := 6
 PRINTF_COL_1_WIDTH := 13
 SOURCEFORGE_MIRROR := downloads.sourceforge.net
 MXE_MIRROR         := https://mirror.mxe.cc/pkg
-PKG_MIRROR         := https://s3.amazonaws.com/mxe-pkg
+PKG_MIRROR         := https:///mxe-pkg.s3.amazonaws.com
```
tonytheodore: extra `/`

vszakats: Ops, thanks — fixed!

tonytheodore:
Thanks!

PR #2336, opened 2019-05-08
[WIP] imagemagick: switch to GH_CONF and update 6.9.0-0 --> 7.0.8-44

### LINKED ISSUE #2335: Upgrade ImageMagick 6.9.0-0 to last stable 7.0.8
ImageMagick package in MXE still a very old version 6.9.0.

New one is 7.0.8 commng with plenty of new format supported (XCF compressed, HEIF, FITS, etc...)

Please update this package. If 7.0.8 has problems to compile under MXE due to changes of API from 6 to 7, at least please upgrade to last 6.9.10-44, as this serie still maintained with API compatibility

Thanks in advance

Gilles Caulier

tonytheodore:
closes #2335

Builds on all four targets, breaks downstream pkg `cimg`.

### DIFF
--- a/src/imagemagick.mk
+++ b/src/imagemagick.mk
@@ -4,36 +4,22 @@ PKG             := imagemagick
 $(PKG)_WEBSITE  := https://www.imagemagick.org/
 $(PKG)_DESCR    := ImageMagick
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 6.9.0-0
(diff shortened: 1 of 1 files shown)


cgilles:
This patch has all requirements to fix #2335

tonytheodore:
`cimg` is actually a header-only package so we don't need to build the examples.

PR #2343, opened 2019-05-22
Add spandsp package

janbrummer:
Add spandsp package, a library of many DSP functions for telephony


src/spandsp.mk
```diff
@@ -0,0 +1,27 @@
...
+
+PKG             := spandsp
+$(PKG)_WEBSITE  := https://www.soft-switch.org
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 0.0.6pre21
+$(PKG)_CHECKSUM := bd152152bf0b204661ab9439c5a649098bcb8cefebcbfa959dd602442739aa50
+$(PKG)_SUBDIR   := spandsp-0.0.6
+$(PKG)_FILE     := spandsp-$($(PKG)_VERSION).tgz
+$(PKG)_URL      := https://download.soft-switch.org/downloads/spandsp/$($(PKG)_FILE)
+$(PKG)_DEPS     := cc tiff
```
tonytheodore: Does it really depend on tiff?

src/spandsp.mk
```diff
@@ -0,0 +1,27 @@
...
+define $(PKG)_UPDATE
+    $(WGET) -q -O- 'https://download.soft-switch.org/downloads/spandsp/' | \
+    grep '<a href="spandsp-' | \
+    $(SED)  -n "s,.*spandsp-\([0-9]\+\.[0-9]*[02468]\.[^']*\)\.tgz.*,\1,p" | \
+    $(SORT) -Vr |\
+    head -1
+endef
+
+define $(PKG)_BUILD
+    cd '$(1)' && ./configure \
```
tonytheodore: patching `configure.ac` requires an `autoreconf` step

src/spandsp.mk
```diff
@@ -0,0 +1,27 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := spandsp
+$(PKG)_WEBSITE  := https://www.soft-switch.org
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 0.0.6pre21
```
tonytheodore: That isn't the latest release, but you may be better off using snapshots if the release is so old.

src/spandsp.mk
janbrummer: Yes, it does. Fax format is tiff.

src/spandsp.mk
janbrummer: The changes between the latest stable release and snapshots are minimal, thus i would prefer the latest official version which is supported. Ok?

src/spandsp.mk
janbrummer: Good catch, i'll fix it.

src/spandsp.mk
janbrummer: Any feedback?

janbrummer: @tonytheodore  Another ping for a feedback

PR #2351, opened 2019-06-22
ghostscript: don't use the local lcms2mt fork

darealshinji:
lcms2mt's symbols clash with those from lcms2


darealshinji:
Instead of removing lcms2mt you could also add a suffix to all the cms symbols:
https://gist.github.com/darealshinji/59cf40931bb758c12f4c4d7fa4cbc062

tonytheodore:
Thanks!

PR #2360, opened 2019-07-14
gcc: Update to 8.3.0

Lord-Kamina:
Updated gcc because honestly, how come nobody had done this before?

To do it, removed a now-obsolete patch. I built a couple other packages after doing this so I'm pretty sure it's working.

Why 8.3.0 and not 9.1.0? Mainly because I figured 8.3.0 is modern enough and gcc8 has certainly had more time to smooth over any rough edges.


jonaski:
Can't make boost compile with gcc 8
`libs/log/src/spirit_encoding.hpp:56:18: internal compiler error: in make_rtl_for_nonlocal_decl, at cp/decl.c:6590`

Lord-Kamina:
@jonaski Can you see if this patch fixes it for you? Boost compiles just fine for me using this.

tonytheodore:
The current way to build with different versions of gcc is:
```
make MXE_PLUGIN_DIRS=plugins/gcc8
```
The patch is already included for gcc8 and not required in gcc9.

Lord-Kamina:
Oh. But, is there a point to keep the default four major versions behind? The reason I got into doing this was the icu4c PR I submitted alongside. That built with gcc5.5 but the log was littered with errors relating to not properly detecting c++11 for some reason. These things are just gonna keep happening more and more as time goes on.

jonaski:
I switched to gcc 9 instead and then it worked.

lotharsm:
Hi all, please excuse this quite novice question:

If I want to switch from the default 5.5.0 compiler to 8.x or 9.x, is it sufficient to run
`make cc MXE_PLUGIN_DIRS=plugins/gcc8` only?

mabrand:
In the meantime there is gcc 9, available as overlay, and probably soon 10.

PR #2361, opened 2019-07-14
icu4c: Update to 64.2

Lord-Kamina:
This PR incorporates #2191, and it also fixes some issues with the previous port, resulting in icu linking against the stub data library instead of the actual data.


jonaski:
Thanks for this, I tried to update icu4c but was unsuccessful.

tonytheodore:
Can you check your editor for tab/space handling? It's hard to make sense of some changes.

Also, the `autoreconf` steps don't seem to be required any longer.

Lord-Kamina:
> Can you check your editor for tab/space handling? It's hard to make sense of some changes.
> 
> Also, the `autoreconf` steps don't seem to be required any longer.

Should I prefer tabs or spaces? My editor can easily convert between either.

Lord-Kamina:
Fixed spacing on this and integrated the url update commit.

src/icu4c.mk
```diff
@@ -20,35 +19,42 @@ define $(PKG)_UPDATE
...
 
 define $(PKG)_BUILD_COMMON
-    cd '$(1)/source' && autoreconf -fi
-    mkdir '$(1).native' && cd '$(1).native' && '$(1)/source/configure' \
+    cd '$(SOURCE_DIR)/source' && autoreconf -fi
+    mkdir '$(BUILD_DIR)/$(PKG).native'
+    cd '$(BUILD_DIR)/$(PKG).native' && '$(SOURCE_DIR)/source/configure' \
         CC=$(BUILD_CC) CXX=$(BUILD_CXX)
-    $(MAKE) -C '$(1).native' -j '$(JOBS)'
+        $(MAKE) -C '$(BUILD_DIR)/$(PKG).native' -j $(JOBS) VERBOSE=1
```
tonytheodore: whitespace?

src/icu4c.mk
```diff
@@ -20,35 +19,42 @@ define $(PKG)_UPDATE
...
-        --with-cross-build='$(1).native' \
-        CFLAGS=-DU_USING_ICU_NAMESPACE=0 \
-        CXXFLAGS='--std=gnu++0x' \
-        SHELL=bash
-
-    $(MAKE) -C '$(1).cross' -j '$(JOBS)' install
+        --with-cross-build='$(BUILD_DIR)/$(PKG).native' \
+        CFLAGS='-DU_USING_ICU_NAMESPACE=0' \
+        CXXFLAGS='-std=c++11' \
+        LIBS='-lmsvcr100' \
```
tonytheodore: As far as I know, different `msvcrt` requires separate install of re-dist components, this isn't something we can put in a build rule.

The next default will be the new `ucrt`.

src/icu4c.mk
```diff
@@ -20,35 +19,42 @@ define $(PKG)_UPDATE
...
+        SHELL=$(SHELL)
+    
+    rm -fv '$(PREFIX)/$(TARGET)/lib/libicu*.dll.a'
```
tonytheodore: Check your shell quoting, this does nothing.

src/icu4c.mk
```diff
@@ -20,35 +19,42 @@ define $(PKG)_UPDATE
...
+    $(MAKE) -C '$(BUILD_DIR)/$(PKG).cross' -j $(JOBS) all-recursive VERBOSE=1 SHELL=$(SHELL)
+    $(MAKE) -C '$(BUILD_DIR)/$(PKG).cross' -j 1 install VERBOSE=1 SHELL=$(SHELL)
     ln -sf '$(PREFIX)/$(TARGET)/bin/icu-config' '$(PREFIX)/bin/$(TARGET)-icu-config'
 endef
 
 define $(PKG)_BUILD_SHARED
-    $($(PKG)_BUILD_COMMON)
+    $(call $(PKG)_BUILD_COMMON)
```
tonytheodore: Why add `call`?

src/icu4c.mk
```diff
@@ -20,35 +19,42 @@ define $(PKG)_UPDATE
...
     # icu4c installs its DLLs to lib/. Move them to bin/.
-    mv -fv $(PREFIX)/$(TARGET)/lib/icu*.dll '$(PREFIX)/$(TARGET)/bin/'
+    mv "$(PREFIX)/$(TARGET)/lib/libicudt$($(PKG)_MAJOR).dll" "$(PREFIX)/$(TARGET)/lib/icudt$($(PKG)_MAJOR).dll"
+    cd "$(PREFIX)/$(TARGET)/lib" && ln -fs "icudt$($(PKG)_MAJOR).dll" "icudt.dll"
+    mv -fv $(PREFIX)/$(TARGET)/lib/icu*.dll "$(PREFIX)/$(TARGET)/bin/"
```
tonytheodore: This seems to clobber the real data file with the stub data. Is that what you wanted?

src/icu4c.mk
```diff
@@ -20,35 +19,42 @@ define $(PKG)_UPDATE
...
     # add symlinks icu*<version>.dll.a to icu*.dll.a
-    for lib in `ls '$(PREFIX)/$(TARGET)/lib/' | grep 'icu.*\.dll\.a' | cut -d '.' -f 1 | tr '\n' ' '`; \
+    for lib in $(shell ls '$(PREFIX)/$(TARGET)/lib/' | grep 'icu.*\.dll\.a' | cut -d '.' -f 1 | tr '\n' ' '); \
```
tonytheodore: Why add `shell`?

src/icu4c.mk
Lord-Kamina: I don't think so. Perhaps added extra indent accidentally.

src/icu4c.mk
Lord-Kamina: Maybe it technically requires separate install but in practice it builds with MXE and that runtime should be present on most if not all windows installs after win7.

src/icu4c.mk
Lord-Kamina: Except it does, I'll explain below.

src/icu4c.mk
Lord-Kamina: No reason, just cleaner I think.

src/icu4c.mk
Lord-Kamina: No; at this point the stub data should have been replaced by the real data file. This just ensures dlls have proper names and locations.

src/icu4c.mk
Lord-Kamina: Without adding shell like this, build fails with an obscure error on some systems because of different `echo` implementations (Support for `-n`), on some systems we end uo generating broken makefiles if `sh` is used instead of bash.

src/icu4c.mk
Lord-Kamina: Might be leftover from a previous iteration because it's technically not needed here I think, but it's better anyway because the '``' syntax is deprecated and discouraged as far as I know. The correct way to do it is using $() but that obviously won't work in a makefile.

src/icu4c.mk
Lord-Kamina: Anyway, newer versions of icu won't work with msvcrt; any icu calls will produce a crash on runtime due to a missing symbol definition.

src/icu4c.mk
Lord-Kamina: I think you might be right on this; I'm doing some more changes to this, as well as checking whether plain msvcrt will work.

src/icu4c.mk
tonytheodore: I understand the `SHELL=$(SHELL)`, we've seen that often in the past.

I meant this line:
```
rm -fv '$(PREFIX)/$(TARGET)/lib/libicu*.dll.a'
```

will try to delete a file literally named `libicu*.dll.a`.

src/icu4c.mk
tonytheodore: It's cleaner to only use `call` if you are actually sending parameters.

src/icu4c.mk
tonytheodore: The `$()` syntax can be used in makefiles:
```diff
-    for lib in `ls '$(PREFIX)/$(TARGET)/lib/' | grep 'icu.*\.dll\.a' | cut -d '.' -f 1 | tr '\n' ' '`; \
+    for lib in $$(ls '$(PREFIX)/$(TARGET)/lib/' | grep 'icu.*\.dll\.a' | cut -d '.' -f 1 | tr '\n' ' '); \
```
see two lines down where `$${lib}.dll.a` is used.

src/icu4c.mk
Lord-Kamina: Ah. You're right.
Still, I think this one is not quite ready for primetime yet. The thing that is puzzling me right now... I've swiped some of the msys2 patches to install dlls in bin; however I cannot understand why it installs dlls in bin (including a stub data) and import libs in lib, where it also places the full data dll and a dll.a for the stub data file...

I just don't understand that.
I'm taking a look at the makefiles and it seems the full data lacks logic for creating/installing the import library that the stubdata has. Do you know what could be the reasoning behind this?

tonytheodore: the *.dll files are used for deployment and need to be in the runtime directory to be found without a modifying %PATH%. The *.dll.a are used for linking and will usually be looked for in `lib`. 

I'm guessing they are trying to default to linking against the stub data so you can have more flexibility at runtime?

Lord-Kamina: > the *.dll files are used for deployment and need to be in the runtime directory to be found without a modifying %PATH%. The *.dll.a are used for linking and will usually be looked for in lib.

That I understand.

> I'm guessing they are trying to default to linking against the stub data so you can have more flexibility at runtime?

I'm guessing that _could_ make sense? Excpt right now it's the other way around (stub in bin, full in lib; although that could partially be my fault)

Anyway, what do you think we should do? Keep the way you say (stub for linking, full data in bin)? Or should we have full data both for runtime and linking? I'd say the latter but I don't know.

tonytheodore: The latter sounds more reasonable as a default, using the stub data could lead to unexpected surprises when deploying.

I'll see if there's a simple test program to try.

Lord-Kamina: Would we necessarily need an import lib or is it feasible to just drop the dll (or a symlink to it) in `lib`?

As for test programs... havent' actually tested it yet but I got one from a blog...

Lord-Kamina: https://gist.github.com/Lord-Kamina/9b8dbe079c58420ef9850f3922231740

tonytheodore: Thanks, I've added that https://github.com/mxe/mxe/pull/2421/commits/2ec080e7e05e86c91976c22367458899d6685317 in #2421

It builds a zip that packages the correct dlls and executes on Win7 using msvcrt.

Lord-Kamina:
#2421 made this obsolete.

PR #2363, opened 2019-07-19
Update binutils

jonaski:


ZachBacon:
seems it maybe prudent to update this to 2.34, there's apparently a backport patch to fix a lot of the major issues that have been introduced, however it seems coming up there will be 2.35 that will have the fix integrated.

jonaski:
Do you have a link to the patch? It would be useful to see if it's the same patching I use now.
I have updated to the latest version of binutils (2.34) in my own MXE repository that I use for the official releases of Strawberry. I've been using that version since February.
Everything else is at the latest version too.
It's here in case anyone is interested: https://github.com/strawberrymusicplayer/strawberry-mxe
Feel free to pick up anything you need, and submit it to the official repository if you want.
I've got some serious back pain / issues so I probably won't have time to submit a PR for everything.
I'll close this one since it's too old now.

ZachBacon:
@jonaski sorry, just been busy the last few days but apparently the commit in question is located here https://sourceware.org/git/?p=binutils-gdb.git;a=commitdiff;h=463ec189fe9eca199edf87cda2c31efbe850390d

PR #2373, opened 2019-07-19
Update glib to 2.58.3

jonaski:
This might still need the static patch, the others are no longer needed.
It might also need a newer gcc than 5.
2.58 is the last major version that uses GNU configure, in 2.59 they switched to meson which I'm not familiar with.


tonytheodore:
glib updates usually require updating all the gnome deps as well (https://github.com/mxe/mxe/pull/2174). Have you tested those?

Lord-Kamina:
This fails in macOS while compiling kqueue because `gioenumtypes.h` does not exist.

Adding 
`$(MAKE) -C '$(BUILD_DIR)/gio'     -j '$(JOBS)'` to `$(PKG)_BUILD_DARWIN` takes care of generating that include file and so the build succeeds.

jonaski:
I got the glib, glib-networking, pango updated to the latest versions, and libsoup updated to the previous version 2.67.1 (I think there is a bug in the latest so I plan to wait for the next).
It's in my repository here: https://github.com/jonaski/strawberry-mxe
I don't use the other gnome libraries, but I can try to update those too and submit a new PR with all updated gnome packages, but that involves some work so I need to know that it will be accepted first.

Lord-Kamina:
Your package won't build static (Or rather it will but will make other libs fail when building static); I'm trying to fix this version and have so far had a little success.  I took a look at your meson version and you literally just removed the logic for building in macOS... Why would you do that?

jonaski:
If it is still needed it's needs to be redone as the latest version of glib uses meson.

Lord-Kamina:
@jonaski I think I've got it now. I'd still rather we update to the latest build if possible (because more things will keep on moving to meson) but there are bigger issues with your meson version than this one. I think I've managed to fix this version and apparently am able to build and link gdk-pixbuf, cairo, pango, and librsvg among others; both dynamically and statically, when using my fixes.

I'm currently doing a few final tests and then will push my changes somewhere so you can see them.

PR #2377, opened 2019-07-27
Update to SDL 2.0.10

lotharsm:
This updates SDL to version 2.0.10.

I also updated the sdl2-1-fixes.patch file just barely to match the new file names (`configure.in` was renamed to `configure.ac` in 2.0.10).

However, I am a bit unsure if we even need the patch(es) in the current way - for a first test, I built the packages (with all four MXE_TARGETS enabled) with the patch disabled, and things just work fine. ~~I'm especially unsure about the patch regarding fixing the static build since I had to *disable* the patch in order to achieve a fully functional static build of SDL 2.0.9.~~

EDIT: I must have done something wrong with my old SDL 2.0.9 tests since I can't replicate the issues I had with this new 2.0.10 package - everything works as intended.


tonytheodore:
Thanks!

PR #2378, opened 2019-07-28
New package: fluidsynth

lotharsm:
This adds a package for the FluidSynth library, a free Soundfont compatible MIDI synthesizer. This library is for example used by VLC and is supported by ScummVM.

### DIFF
--- a/src/fluidsynth.mk
+++ b/src/fluidsynth.mk
@@ -10,16 +10,22 @@ $(PKG)_GH_CONF  := FluidSynth/fluidsynth/tags,v
 $(PKG)_DEPS     := cc glib
 
 define $(PKG)_BUILD
-    cd '$(BUILD_DIR)' && '$(TARGET)-cmake' '$(SOURCE_DIR)' \
-        -DBUILD_SHARED_LIBS=$(CMAKE_SHARED_BOOL) \
-	-DCMAKE_BUILD_TYPE=Release
+    cd '$(BUILD_DIR)' && '$(TARGET)-cmake' '$(SOURCE_DIR)'
     $(MAKE) -C '$(BUILD_DIR)' -j '$(JOBS)' VERBOSE=1
-    $(MAKE) -C '$(BUILD_DIR)' -j 1 install VERBOSE=1
+
+    $(INSTALL) -d '$(PREFIX)/$(TARGET)/include/fluidsynth'
+    $(INSTALL) -v '$(BUILD_DIR)/include/fluidsynth.h'    '$(PREFIX)/$(TARGET)/include/'
+    $(INSTALL) -v '$(BUILD_DIR)/include/fluidsynth/'*.h  '$(PREFIX)/$(TARGET)/include/fluidsynth/'
+    $(INSTALL) -v '$(SOURCE_DIR)/include/fluidsynth/'*.h '$(PREFIX)/$(TARGET)/include/fluidsynth/'
+    $(INSTALL) -v '$(BUILD_DIR)/fluidsynth.pc'           '$(PREFIX)/$(TARGET)/lib/pkgconfig/' 
+    $(INSTALL) -v '$(BUILD_DIR)/src/libfluidsynth.a' '$(PREFIX)/$(TARGET)/lib/'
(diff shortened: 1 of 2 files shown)


lotharsm:
Building fluidsynth itself works, I currently only have some issues with the fluidsynth-test.c file (complains about missing DSound references... guess I have to check my LDFLAGS.

lotharsm:
Okay folks, I guess I need some help here.

fluidsynth itself compiles without issues. However, when I try to compile the test file, I get tons and tons of  ``undefined reference to `g_$insert-more-glib-stuff-here` and to `dsound`, even though I compiled fluidsynth statically.

What's more interesting: When I build other applications that use fluidsynth (using ScummVM here as an example), everything compiles and links as expected with the same .pc file.

Any idea what I am missing here? :)

tonytheodore:
I'm only seeing undefined references to `dsound` - other apps you're building may link this in their own config.

Ideally, CMake should put this and the `glib` dependency in the *.pc file.

src/fluidsynth.mk
```diff
@@ -0,0 +1,25 @@
...
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 2.0.5
+$(PKG)_CHECKSUM := 69b244512883491e7e66b4d0151c61a0d6d867d4d2828c732563be0f78abcc51
+$(PKG)_GH_CONF  := FluidSynth/fluidsynth/tags,v
+$(PKG)_DEPS     := cc glib
+
+define $(PKG)_BUILD
+    cd '$(BUILD_DIR)' && '$(TARGET)-cmake' '$(SOURCE_DIR)' \
+        -DBUILD_SHARED_LIBS=$(CMAKE_SHARED_BOOL) \
+	-DCMAKE_BUILD_TYPE=Release
```
tonytheodore: These two lines aren't necessary, BUILD_SHARED_LIBS and CMAKE_BUILD_TYPE are set by the toolchain file.

src/fluidsynth.mk
```diff
@@ -0,0 +1,25 @@
...
+    $(MAKE) -C '$(BUILD_DIR)' -j '$(JOBS)' VERBOSE=1
+    $(MAKE) -C '$(BUILD_DIR)' -j 1 install VERBOSE=1
```
tonytheodore: Install only the lib and headers, not the *.exe

src/fluidsynth.mk
```diff
@@ -0,0 +1,25 @@
...
+
+    # compile test
+    '$(TARGET)-gcc' \
+        -W -Wall -ansi -pedantic \
+        '$(TEST_FILE)' -o '$(PREFIX)/$(TARGET)/bin/test-fluidsynth.exe' \
+        `'$(TARGET)-pkg-config' --cflags --libs fluidsynth` \
+	`'$(TARGET)-pkg-config' --cflags --libs glib-2.0`
```
tonytheodore: tab character

lotharsm:
> 
> 
> I'm only seeing undefined references to `dsound` - other apps you're building may link this in their own config.
> 
> Ideally, CMake should put this and the `glib` dependency in the *.pc file.

Thanks a lot, I will try again with your suggestions! The "dsound" dependency gets indeed pulled in by the dependency on sdl2 for ScummVM, so this issue is solved. However, I can't explain the issues with glib since neither ScummVM nor it's other dependencies rely on glib and still build fine. :D

tonytheodore:
Ideally, cmake should (conditionally?) add:
```
Requires: glib-2.0
```
in fluidsynth.pc. That will mean it doesn't need to be specified in the test build invocation.

lotharsm:
Everything's working as intended now :)

tonytheodore:
> Everything's working as intended now :)

Did you test the shared builds?

lotharsm:
> Did you test the shared builds?

Thanks for pointing that out - I just re-tested my last commit and there are indeed problems with the shared builds :(

Thanks a lot for taking care of this!

tonytheodore:
`CMake` takes care of most of it, so it's easier to patch the input than manually adjust the output.

Do you want see if some version of the patches will be accepted upstream?

Issue #2380, opened 2019-08-02
curl: Error 60 - SSL peer certificate or SSH remote key was not OK


lotharsm:
Hi folks,

I have an issue establishing an SSL connection to a server with an application built with libcurl from mxe. The server I'm trying to connect to is using an SSL certificate issued by Let's Encrypt.

When connecting, I get the error code "60 - SSL peer certificate or SSH remote key was not OK". When I build the same application with MSYS2 using it's libcurl build, I have no issues at all.

I think the mxe build is somehow missing a current certficate store. Any idea how to solve this? :)

Thanks!

PR #2381, opened 2019-08-03
curl: switch to winssl/schannel

lotharsm:
Currently, libcurl is not able to use the SSL certificates from the certificate store in Windows due to the dependency on gnutls. This patch fixes this by switching to Schannel/WinSSL, Windows' native SSL library which has access to the certificate store of the OS.

I originally stumbled across this issue in https://github.com/mxe/mxe/issues/2380.

If you don't want any changes in the current curl.mk package, I'll happily split this into a seperate curl-winssl.mk file.


tonytheodore:
This seems like it will only potentially affect old versions, from https://curl.haxx.se/docs/install.html:

>Schannel (from Windows SSPI), is the native SSL library in Windows. However, Schannel in Windows <= XP is unable to connect to servers that no longer support the legacy handshakes and algorithms used by those versions. If you will be using curl in one of those earlier versions of Windows you should choose another SSL backend such as OpenSSL.

Thanks!

PR #2384, opened 2019-08-14
qtifw: update and build test

muellni:
I switched to the unix-download, hence the changed lineendings in the patch.
The unix-download is a tar archive with a false tar.gz file suffix, so I had to change the filename and enable tar archives in mxe.

This package will now create an example installer as test executable.


mabrand:
thanks

PR #2389, opened 2019-08-29
Add gssdp package

janbrummer:
Add A GObject-based API for handling resource discovery and announcement over SSDP

Please note that the latest release 1.2 would need a newer glib version (>= 2.54).


janbrummer:
Anything left before it can be merged or is it fine now?

src/gssdp.mk
```diff
@@ -0,0 +1,24 @@
+PKG             := gssdp
```
tonytheodore: Add standard file header above

janbrummer: Ok, done.

tonytheodore:
Build fails for [i686-w64-mingw32.static](https://gist.github.com/tonytheodore/d4a17327b66e77b62ab63ecf08b775a5), other targets succeed.

janbrummer:
> Build fails for [i686-w64-mingw32.static](https://gist.github.com/tonytheodore/d4a17327b66e77b62ab63ecf08b775a5), other targets succeed.

This is strange and i cannot find the reason for it. Is is legit to disable it for i686?

tonytheodore:
It's probably using newer api methods that aren't available in i686 by default. Try adding something like:
```
CFLAGS='-D_WIN32_WINNT=0x0600 -DWINVER=0x0600'
```

janbrummer:
Works fine on my machine now, thanks for you feedback! LGTM.

janbrummer:
Is it possible to merge this one?

PR #2391, opened 2019-08-29
Add glib-networking package

janbrummer:
Newer versions of glib-networking would require complete meson build system support as autotools has been dropped.

### DIFF
--- a/src/glib-networking.mk
+++ b/src/glib-networking.mk
@@ -1,0 +1,24 @@
+PKG             := glib-networking
+$(PKG)_WEBSITE  := https://www.gnome.org
+$(PKG)_DESCR    := Network-related GIO modules for glib
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 2.54.1
+$(PKG)_CHECKSUM := eaa787b653015a0de31c928e9a17eb57b4ce23c8cf6f277afaec0d685335012f
+$(PKG)_SUBDIR   := glib-networking-$($(PKG)_VERSION)
+$(PKG)_FILE     := glib-networking-$($(PKG)_VERSION).tar.xz
+$(PKG)_URL      := https://download.gnome.org/sources/$(PKG)/$(call SHORT_PKG_VERSION,$(PKG))/$($(PKG)_FILE)
+$(PKG)_DEPS     := cc gnutls glib
+
+define $(PKG)_UPDATE
+    $(WGET) -q -O- 'https://gitlab.gnome.org/GNOME/glib-networking/-/tags' | \
+    $(SED) -n "s,.*glib-networking-\([0-9]\+\.[0-9]*[0-9]*\.[^']*\)\.tar.*,\1,p" | \
+    $(SORT) -Vr | \
+    head -1
+endef
+
+define $(PKG)_BUILD
+    cd '$(BUILD_DIR)' && '$(SOURCE_DIR)/configure' \
+         $(MXE_CONFIGURE_OPTS)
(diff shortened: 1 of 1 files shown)


src/glib-networking.mk
```diff
@@ -0,0 +1,25 @@
+PKG             := glib-networking
+$(PKG)_WEBSITE  := https://www.gnome.org
+$(PKG)_DESCR    := Network-related GIO modules for glib
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 2.54.1
+$(PKG)_CHECKSUM := eaa787b653015a0de31c928e9a17eb57b4ce23c8cf6f277afaec0d685335012f
+$(PKG)_SUBDIR   := glib-networking-$($(PKG)_VERSION)
+$(PKG)_FILE     := glib-networking-$($(PKG)_VERSION).tar.xz
+$(PKG)_URL      := https://download.gnome.org/sources/$(PKG)/$(call SHORT_PKG_VERSION,$(PKG))/$($(PKG)_FILE)
+$(PKG)_DEPS     := cc gnutls
```
jonaski: Add glib

src/glib-networking.mk
```diff
@@ -0,0 +1,25 @@
...
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 2.54.1
+$(PKG)_CHECKSUM := eaa787b653015a0de31c928e9a17eb57b4ce23c8cf6f277afaec0d685335012f
+$(PKG)_SUBDIR   := glib-networking-$($(PKG)_VERSION)
+$(PKG)_FILE     := glib-networking-$($(PKG)_VERSION).tar.xz
+$(PKG)_URL      := https://download.gnome.org/sources/$(PKG)/$(call SHORT_PKG_VERSION,$(PKG))/$($(PKG)_FILE)
+$(PKG)_DEPS     := cc gnutls glib
+
+define $(PKG)_UPDATE
+   $(WGET) -q -O- 'https://git.gnome.org/browse/glib-networking/refs/tags' | \
```
jonaski: This update does not work. try this:
    $(WGET) -q -O- 'https://gitlab.gnome.org/GNOME/glib-networking/-/tags' | \
    $(SED) -n "s,.*glib-networking-\([0-9]\+\.[0-9]*[0-9]*\.[^']*\)\.tar.*,\1,p" | \
    $(SORT) -Vr | \

janbrummer:
Is this one ok to merge?

janbrummer:
> Build out-of-source with `BUILD_DIR` etc. see glib.mk as an example.

Ok, done.

tonytheodore:
Thanks!

tonytheodore:
@janbrummer I've had to [disable ca-certificates](https://github.com/mxe/mxe/commit/03ba05e2d88c9202f8641aed479a8d8836deb8de) as configure was failing without them being found. Is it likely that your build was picking a native path for those?

janbrummer:
Interesting. Do you have a bug report or a log file?

tonytheodore:
Log is here https://gist.github.com/tonytheodore/9d3521a3e48983b09aa42e85c2ca1ef6

From configure:15759
```
        if test -z "$with_ca_certificates"; then
	    for f in /etc/pki/tls/certs/ca-bundle.crt \
	       	     /etc/ssl/certs/ca-certificates.crt \
		     /etc/ssl/ca-bundle.pem; do
		if test -f "$f"; then
		    with_ca_certificates="$f"
		fi
```

it's looking for native paths and effectively disabled anyway.

I think we can ignore it till we get meson and the latest version.

janbrummer:
Yes, makes sense.

PR #2394, opened 2019-09-07
Fix fontconfig update

jonaski:


tonytheodore:
The pre-release versions are intentionally excluded.

PR #2402, opened 2019-09-17
secure/fix a few URLs

vszakats:
Detected via https-everywhere rules.

### DIFF
--- a/src/libmpeg2.mk
+++ b/src/libmpeg2.mk
@@ -1,18 +1,18 @@
 # This file is part of MXE. See LICENSE.md for licensing information.
 
 PKG             := libmpeg2
-$(PKG)_WEBSITE  := https://plib.sourceforge.io/
+$(PKG)_WEBSITE  := https://libmpeg2.sourceforge.io/
 $(PKG)_DESCR    := libmpeg2 - a free MPEG-2 video stream decoder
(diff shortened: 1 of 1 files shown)


vszakats:
Something else to be done with this PR to get it merged?

src/icu4c.mk
```diff
@@ -13,7 +13,7 @@ $(PKG)_URL      := https://ssl.icu-project.org/files/$(PKG)/$($(PKG)_VERSION)/$(
 $(PKG)_DEPS     := cc
 
 define $(PKG)_UPDATE
-    $(WGET) -q -O- 'http://bugs.icu-project.org/trac/browser/icu/tags' | \
+    $(WGET) -q -O- 'https://ssl.icu-project.org/trac/browser/icu/tags' | \
```
jonaski: This site doesn't work for me with https, I get connection refused.

vszakats: Indeed, thanks! — Their SSL support is not operating, even though it has been for many years.
I've undone this update.

vszakats:
With the icu4c link update undone, is there some remaining concern?

tonytheodore:
Thanks!

Note `make cleanup-style` will fix tab issues - there's also one in curl.mk but I don't have a machine I can push from at the moment.

PR #2406, opened 2019-09-21
Glib: Update to 2.58.3 (supersedes @jonaski's #2373)

Lord-Kamina:
I wanted to create a PR at his fork instead but apparently github doesn't allow for that.

Anyway, this included #2373 as well as includes some extra fixes to allow building statically, among other things.


Lord-Kamina:
Fixed this to keep format-patch style.

Lord-Kamina:
Hello, just a friendly reminder that it'd be really cool if this got merged. :P

janbrummer:
Thanks for your PR. Not knowing your PR I've added a recent glib version using meson on my local system, is this of interest?

Meson is the way to go for all gtk/gnome stuff so a jump to meson is an option?

Lord-Kamina:
> Thanks for your PR. Not knowing your PR I've added a recent glib version using meson on my local system, is this of interest?
> 
> Meson is the way to go for all gtk/gnome stuff so a jump to meson is an option?

It should be, I think. Take a look at the patches applied for static builds and such, though.

janbrummer:
Before starting with my work: This means meson will be a requirement to compile glib/gtk & co and is no longer a plugin. Is this correct?

jonaski (approved):

mabrand (approved):

tonytheodore:
I've started a test build, [glib historically breaks other packages](https://github.com/mxe/mxe/pull/2373#issuecomment-513525119).

danger89:
Ownoooh

danger89:
This is no longer needed if we create this PR: https://github.com/danger89/mxe/pull/1

PR #2410, opened 2019-10-03
icu4c: Fix urls and update

jonaski:


src/icu4c.mk
```diff
@@ -1,23 +1,22 @@
 # This file is part of MXE. See LICENSE.md for licensing information.
 
 PKG             := icu4c
-$(PKG)_WEBSITE  := https://ssl.icu-project.org/
+$(PKG)_WEBSITE  := http://site.icu-project.org/
```
tonytheodore: In the absence of https url, point to the project at `https://github.com/unicode-org/icu` - this has a link back to the main project site and is potentially more useful to mxe users.

tonytheodore: http://blog.unicode.org/2018/07/icu-moves-to-github-and-jira.html

jonaski: I could update it to use GH_CONF, but they have tagged their releases with - instead of .
So when I use $(PKG)_GH_CONF  := unicode-org/icu/tags, release-
The version will be 65-1 instead of 65.1, is there a way to workaround that?

jonaski:
Is this good?

tonytheodore:
Thanks!

PR #2415, opened 2019-10-12
Set tries and timeout for wget in update rules

jonaski:
The default timeout for wget is 15 minutes.
It's pretty annoying having to wait 15 minutes for `make update` to continue for each update where the site is actually down.
Right now that's the case for wavpack.
This sets the timeout to 6 seconds and tries to 2, so it doesn't try more than 2 times.


tonytheodore:
I'm reasonably sure you can change it once in the [`WGET` variable](https://github.com/mxe/mxe/blob/master/Makefile#L49). When that variable is [used in download rules](https://github.com/mxe/mxe/blob/master/Makefile#L296) the later options will take precedence.

jonaski:
Yeah, that's probably better. Closing this then.

PR #2416, opened 2019-10-12
Set tries and timeout for wget

jonaski:
The default timeout for wget is 15 minutes.
It's pretty annoying having to wait 15 minutes for make update to continue for each update where the site is actually down.
Right now that's the case for wavpack.
This sets the timeout to 6 seconds and tries to 2, so it doesn't try more than 2 times.


tonytheodore:
6 seconds seems a little optimistic, but I have nothing to base that on - lets see how it goes.

Thanks!

jonaski:
Bear in mind, it's not the actual time it takes to download the file, it's the time when no data is read at all. (also connect timeout and dns timeout is included in the same parameter).

PR #2417, opened 2019-10-12
Add brotli

jonaski:
brotli is required by recent versions of libsoup


tonytheodore:
This installs both static and shared libs, along with the *.exe that should be disabled.

Can you add a simple test program build?

Issue #1037, opened 2015-12-04
Add libraries for Octave and Shogun


tonytheodore:
Various maths and scientific libraries are common across both toolkits and there's a lot of prior work to build on:

#304
http://hg.octave.org/mxe-octave

Long term goal is to incorporate all of mxe-octave (and it's getting harder as time passes), but we'll start with the libs.

tonytheodore:
Also https://github.com/schloegl/mxe

PR #2420, opened 2019-10-19
Devel001

schloegl:
Please consider including the packages libb64 and biosig into mxe. 
- libb64 is a small library for base64 encoding/decoding, used by some tools in biosig. 
- biosig provides some libraries (libbiosig, libgdf, libphysicalunits), and some cli tools (save2gdf, biosig_fhir, physicalunits). libbiosig is a library used in a number of additional tools (sigviewer, stimfit), and is important for various language bindings of biosig to octave, matlab, python, R, etc.  

I tried to follow the suggested rules below, let me know if there is anything need to be changed. 
 

  * install .pc file

Thanks!

### DIFF
--- a/src/biosig.mk
+++ b/src/biosig.mk
@@ -1,0 +1,96 @@
+# This file is part of MXE.
+# See index.html for further information.
+
+PKG             := biosig
+$(PKG)_WEBSITE  := http://biosig.sf.net/
+$(PKG)_DESCR    := biosig
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 1.9.5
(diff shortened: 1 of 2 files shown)


schloegl:
Can I expect a review of these pull requests anytime soon ? Is there something I should fix?

tonytheodore:
Hi @schloegl, thanks for the updates!

Adding these packages has been on my list [for a long time](https://github.com/mxe/mxe/issues/1037#issuecomment-162718823). 

Don't try to make this PR mergeable, I can use `git diff` to pick out the packages that make sense in core mxe, and give you some guidelines for the rest.

PR #2421, opened 2019-10-21
icu4c: update 56.1 --> 65.1 and modernise

tonytheodore:
- autoreconf and patch no longer required
- use GH_CONF
- separate native build
- use $($(PKG)_CONFIGURE_OPTS)
- bundle test with `peldd` to verify actual data dll deployment

replaces #2361 and #2191


Lord-Kamina:
Just some lingering comments from the other thread:

1) After some digging: The msvcr100 issue was better fixed with a patch in msys2, and was caused by functions missing pre-win7. This issue appeared in 57 and disappeared in icu62 because those files were nuked altogether, the caveat is they might at some point make their way back and we might need that patch in the future.

2) Before going to bed last night, I tried a build using just the bindir patches over at msys2 (could be replicated by moving things post-build). I left the full data in bin and the  stub `.dll.a` in lib; this produced a functional install with which I could compile both the test and freetype. I then tried the test executable in a win10 VM; it didn't work as expected but I suspect that has to do with differences in how windows and unix handle stream redirection.

3) Based on the above, we could still make a better test probably, although that's not a huge probability now. Still, I might be able to look into it tomorrow.

src/icu4c.mk
```diff
@@ -4,52 +4,80 @@ PKG             := icu4c
...
 endef
 
 define $(PKG)_BUILD_COMMON
-    cd '$(1)/source' && autoreconf -fi
-    mkdir '$(1).native' && cd '$(1).native' && '$(1)/source/configure' \
-        CC=$(BUILD_CC) CXX=$(BUILD_CXX)
-    $(MAKE) -C '$(1).native' -j '$(JOBS)'
-
-    mkdir '$(1).cross' && cd '$(1).cross' && '$(1)/source/configure' \
+    rm -fv $(shell echo "$(PREFIX)/$(TARGET)"/{bin,lib}/{lib,libs,}icu'*'.{a,dll,dll.a})
```
Lord-Kamina: Didn't know this syntax was a thing. Looks nice; will it work in bash 3.22, though?

Lord-Kamina:
In general this looks like a very nice change.
Just a couple things... Wouldn't it be better to use the create the versioned libraries, symlink non-versioned and patch pkg-config to use non-versioned names as msys2 does?

tonytheodore:
> The msvcr100 issue was better fixed with a patch in msys2, and was caused by functions missing pre-win7.

Not knowing what the issue was, I'd normally look at `WINVER` and/or `_WIN32_WINNT` rather than the runtime. The test program fails to execute on WinXP due to those missing functions (expected) - but they are present in msvcrt and Win7 runs fine.

> the caveat is they might at some point make their way back and we might need that patch in the future.

We will have easily selectable runtimes in the near future, December is creeping up but I'll start on it again next weekend.

> I tried a build using just the bindir patches over at msys2 (could be replicated by moving things post-build)

I'm happy to continue moving these post-build, unless the patches are submitted upstream. Ideally, we'd have no patches across updates.

> I left the full data in bin and the stub .dll.a in lib

I know now, there isn't a stub `.dll.a`, only a stub `.dll` that is placed in `bin` by default. The linking is always going cause a lookup for the same file in the runtime search path.

> Based on the above, we could still make a better test probably, although that's not a huge probability now. Still, I might be able to look into it tomorrow.

Better tests are always welcome, but the test only needs to confirm correct linking - which I think it does. There's too much to read about [minimising icu data](http://userguide.icu-project.org/icudata), this mimics the previous behaviour.

> Wouldn't it be better to use the create the versioned libraries, symlink non-versioned and patch pkg-config to use non-versioned names as msys2 does?

It would certainly seem more correct, but that correctness needs to come from upstream. They don't provide versioned *.pc files so they're not thinking about side-by-side installs...unless this is their solution to side-by-side installs with the default being "current"?

src/icu4c.mk
tonytheodore: bash on darwin is old:
```
% which bash
/bin/bash
% `which bash` --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin19)
Copyright (C) 2007 Free Software Foundation, Inc.
```
But brace expansion is nice. The use of `$(shell ...)` here defers the expansion of `*` so it can be provided as a literal for the actual command to evaluate.

tonytheodore: I never really thought about it before, but the pervasive use of backtick expansion in mxe is probably caused by reducing the amount of deferred expansions between `make` and the shell.

Lord-Kamina: > bash on darwin is old:

That's why I was asking but I already checked the manpage and it should work.

Lord-Kamina:
> > The msvcr100 issue was better fixed with a patch in msys2, and was caused by functions missing pre-win7.
> 
> Not knowing what the issue was, I'd normally look at `WINVER` and/or `_WIN32_WINNT` rather than the runtime. The test program fails to execute on WinXP due to those missing functions (expected) - but they are present in msvcrt and Win7 runs fine.
> 

The thing is the function was missing but apparently mingw-w64 headers included it; so it would compile and then fail at runtime.

> > the caveat is they might at some point make their way back and we might need that patch in the future.
> 
> We will have easily selectable runtimes in the near future, December is creeping up but I'll start on it again next weekend.
> 

Either way, as I said, should it be the case in the future (and if we still care about Win7 then), there is a patch already available over in the MSYS repositories.

> > I tried a build using just the bindir patches over at msys2 (could be replicated by moving things post-build)
> 
> I'm happy to continue moving these post-build, unless the patches are submitted upstream. Ideally, we'd have no patches across updates.
> 

Fair enough.

> > I left the full data in bin and the stub .dll.a in lib
> 
> I know now, there isn't a stub `.dll.a`, only a stub `.dll` that is placed in `bin` by default. The linking is always going cause a lookup for the same file in the runtime search path.
> 
> > Based on the above, we could still make a better test probably, although that's not a huge probability now. Still, I might be able to look into it tomorrow.
> 
> Better tests are always welcome, but the test only needs to confirm correct linking - which I think it does. There's too much to read about [minimising icu data](http://userguide.icu-project.org/icudata), this mimics the previous behaviour.
> 

Damn was I sleepy; I meant it was probably not a huge priority. Anyway, it definitely works for checking everything links properly so I guess we're good.

> > Wouldn't it be better to use the create the versioned libraries, symlink non-versioned and patch pkg-config to use non-versioned names as msys2 does?
> 
> It would certainly seem more correct, but that correctness needs to come from upstream. They don't provide versioned *.pc files so they're not thinking about side-by-side installs...unless this is their solution to side-by-side installs with the default being "current"?

I'm not so sure about this; consider this: http://userguide.icu-project.org/howtouseicu#TOC-C-Namespace

tonytheodore:
> I'm not so sure about this; consider this: http://userguide.icu-project.org/howtouseicu#TOC-C-Namespace

Nice, that is now off by default.

>> Wouldn't it be better to use the create the versioned libraries, symlink non-versioned and patch pkg-config to use non-versioned names as msys2 does?
>
> It would certainly seem more correct, but that correctness needs to come from upstream.

That is actually the usual behaviour, but has special handling for mingw, aix, and os400. That can be disabled with a make variable and we can get rid of our manual links.

These are in https://github.com/mxe/mxe/pull/2421/commits/5e45dd3054937f6f50ea9d08510c6f343a0c5686

tonytheodore:
I'm happy with this now, rebased.

PR #2423, opened 2019-10-24
Add gtksourceview3

AltGr:
The repository had gtk3 and gtksourceview2, but not gtksourceview3.

This is my first package so I am not totally sure everything is done
correctly. I had it working successfully for w64 static libraries.


AltGr:
Ah, note that, while everything else is working correctly, I had trouble when trying to display a completion widget:
```
Error while loading the completion UI: The resource at '/org/gnome/gtksourceview/ui/gtksourcecompletion.ui' does not exist
```
The contents of the corresponding resource file are present in the `.a`, though, so it might be an error while linking my application.

PR #2431, opened 2019-11-06
Update qt5 freeze example

tehnick:
Old QtWebkit has known security issues. Also its JavaScript interpreter is very outdated and cannot execute scripts on many modern websites. So I propose update this package even for old version of Qt.

I have successfully tested not only the build of this version of QtWebkit library, but also its work in WinXP. But any additional tests are welcome!


tehnick:
BTW, I have tested builds with GCC 5.x and GCC 6.x. But work -- only for GCC 6.x builds.

tehnick:
> I have tested builds with GCC 5.x and GCC 6.x.

Sorry for misinformation. It appears that all builds were tested with:
```
MXE_PLUGIN_DIRS += plugins/gcc6
```

tehnick:
Now I have explicitly checked that with GCC 5.x `qtwebkit` package is built fine. But I cannot check it in our real applications because they require GCC >= 6.x..

tonytheodore:
Thanks!

PR #2443, opened 2019-12-14
qca: fix build after recent changes

tehnick:
Tested for:
```
MXE_TARGETS := x86_64-w64-mingw32.shared i686-w64-mingw32.shared
MXE_PLUGIN_DIRS += plugins/gcc9
```


tonytheodore:
Thanks!

PR #2445, opened 2019-12-17
add Qt5Network library dependencies

justdan96:
This is a simple edit, I've tested this on my own machine and it works. Without this CMake fails to link executables that require Qt5Network.


mabrand:
thanks

PR #2447, opened 2019-12-20
Upgraded boost to 1.72.0

highperformancecoder:
Note I had to change the test example, as it wouldn't correctly compile and link under MXE. I just picked a simple test example (quick.cpp) from the system library that exercises boost under make and cmake.


ooxi:
Why does the old example no longer compile and link correctly?

highperformancecoder:
On Wed, Jan 01, 2020 at 08:56:33AM -0800, ooxi wrote:
> Why does the old example no longer compile and link correctly?
> 

There was clearly an API change, at least with respect to Win32
targets. I found numerous questions on StackOverflow about how to
build boost::context using code under MinGW, but not much by the way
of answers. I did get the example to compile by adding the compile
flag -DBOOST_USE_WINFIB, but still got link time errors - undefined
symbols that weren't in any of the boost libraries.

-- 

----------------------------------------------------------------------------
Dr Russell Standish                    Phone 0425 253119 (mobile)
Principal, High Performance Coders
Visiting Senior Research Fellow        hpcoder@hpcoders.com.au
Economics, Kingston University         http://www.hpcoders.com.au
----------------------------------------------------------------------------

src/boost.mk
```diff
@@ -72,11 +72,10 @@ define $(PKG)_BUILD
         -W -Wall -Werror -ansi -U__STRICT_ANSI__ -pedantic \
         '$(PWD)/src/$(PKG)-test.cpp' -o '$(PREFIX)/$(TARGET)/bin/test-boost.exe' \
         -DBOOST_THREAD_USE_LIB \
-        -lboost_serialization-mt \
-        -lboost_thread_win32-mt \
-        -lboost_system-mt \
-        -lboost_chrono-mt \
-        -lboost_context-mt
+        -lboost_serialization-mt-x32 \
```
fiesh: This breaks linking of the test program since `-x32` is explicitly added, which is wrong for most architectures.

highperformancecoder:
On Fri, Feb 14, 2020 at 05:44:54AM -0800, fiesh wrote:

> +        -lboost_serialization-mt-x32 \
> 
> This breaks linking of the test program since -x32 is explicitly added, which
> is wrong for most architectures.

I hadn't noticed that, so thanks for pointing it out. I'm building 32
bit only ATM for maximum platform compatibility.


-- 

----------------------------------------------------------------------------
Dr Russell Standish                    Phone 0425 253119 (mobile)
Principal, High Performance Coders     hpcoder@hpcoders.com.au
                      http://www.hpcoders.com.au
----------------------------------------------------------------------------

PR #2452, opened 2020-01-11
update to wxwidgets 3.1.3 / fix update check:

abma:
sourceforge.net seems to be unmaintained


src/wxwidgets.mk
```diff
@@ -4,25 +4,31 @@ PKG             := wxwidgets
...
 endef
 
 define $(PKG)_CONFIGURE_OPTS
-        $(MXE_CONFIGURE_OPTS) \
+       --host='$(TARGET)' \
+       --build='$(BUILD)' \
+       --prefix='$(PREFIX)/$(TARGET)' \
+       $(if $(BUILD_STATIC), \
+          --disable-shared , \
+          --enable-shared )  \
```
tonytheodore: Why not keep using `MXE_CONFIGURE_OPTS`?

abma: the configure script errors out when unknown parameters are passed, i.e. --enable-static is unknown.

introduced by https://github.com/wxWidgets/wxWidgets/commit/aa7e10bb0919982bfdc510bd9cdc8b84d621d80b

abma: should --disable-option-checking be added instead?

i can only agree to the commit message:

"Give an error if an unknown option is specified because this is very helpful
for catching typos which happen quite frequently in practice, considering the
number of options that we have."

janvdijk: I recommend to just pass --enable-option-checking (without arguments) to configure to fix the problem. The advantage is that configure succeeds, but you still get useful warnings: configure finishes with smth. like:

configure: WARNING: unrecognized options: --disable-static, --enable-doc, --enable-gtk-doc, --enable-gtk-doc-html, --enable-gtk-doc-pdf, --with-html-dir, --disable-doxygen

PR #2453, opened 2020-01-17
Fix qt5 cmake

### LINKED ISSUE #1642: Qt5 + cmake doesn't seem to work at all

### LINKED ISSUE #2166: CMake: undefined reference to `qt_static_plugin_QWindowsIntegrationPlugin()'
I have build my application against a statically linked Qt 5.11 with the CMake wrapper. However, when I try to run the application on Windows I get: `"This application failed to start because no Qt platform plugin could be initialized."`

After some desperate googling I guess that I need to add:

`#include <QtPlugin>`

`Q_IMPORT_PLUGIN(QWindowsIntegrationPlugin)`

...but this leads to this `undefined reference to qt_static_plugin_QWindowsIntegrationPlugin()` when building the application. 

How should I do this with CMake? Running the application on Windows has worked before without doing anything so I guess something has changed in the recent Qt versions.

### LINKED ISSUE #2430: Qt 5.13 fails to link with openssl
Hello,

When cross-compiling with cmake a program depending on qt5, I have the following errors at linking:

```
[100%] Linking CXX executable DLAutoParam.exe
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qnetworkinterface_win.o):qnetworkinterface_win.cpp:(.text$_ZN24QNetworkInterfaceManager22interfaceIndexFromNameERK7QString+0x19): undefined reference to `ConvertInterfaceNameToLuidW@8'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qnetworkinterface_win.o):qnetworkinterface_win.cpp:(.text$_ZN24QNetworkInterfaceManager22interfaceIndexFromNameERK7QString+0x3c): undefined reference to `ConvertInterfaceLuidToIndex@8'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qnetworkinterface_win.o):qnetworkinterface_win.cpp:(.text$_ZN24QNetworkInterfaceManager22interfaceNameFromIndexEj+0x24): undefined reference to `ConvertInterfaceIndexToLuid@8'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qnetworkinterface_win.o):qnetworkinterface_win.cpp:(.text$_ZN24QNetworkInterfaceManager22interfaceNameFromIndexEj+0x43): undefined reference to `ConvertInterfaceLuidToNameW@12'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qnetworkinterface_win.o):qnetworkinterface_win.cpp:(.text$_ZN9QHostInfo15localDomainNameEv+0x2b): undefined reference to `GetNetworkParams@8'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qnetworkinterface_win.o):qnetworkinterface_win.cpp:(.text$_ZN9QHostInfo15localDomainNameEv+0x12e): undefined reference to `GetNetworkParams@8'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qnetworkinterface_win.o):qnetworkinterface_win.cpp:(.text$_ZL16interfaceListingv+0x47): undefined reference to `GetAdaptersAddresses@20'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qnetworkinterface_win.o):qnetworkinterface_win.cpp:(.text$_ZL16interfaceListingv+0x19a): undefined reference to `ConvertInterfaceLuidToNameW@12'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qnetworkinterface_win.o):qnetworkinterface_win.cpp:(.text$_ZL16interfaceListingv+0x746): undefined reference to `GetAdaptersAddresses@20'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qsslsocket_openssl_symbols.o):qsslsocket_openssl_symbols.cpp:(.text$_Z23q_ASN1_STRING_get0_dataPK14asn1_string_st+0x1): undefined reference to `ASN1_STRING_get0_data'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qsslsocket_openssl_symbols.o):qsslsocket_openssl_symbols.cpp:(.text$_Z18q_OPENSSL_init_sslyPK21ossl_init_settings_st+0x1): undefined reference to `OPENSSL_init_ssl'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qsslsocket_openssl_symbols.o):qsslsocket_openssl_symbols.cpp:(.text$_Z21q_OPENSSL_init_cryptoyPK21ossl_init_settings_st+0x1): undefined reference to `OPENSSL_init_crypto'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qsslsocket_openssl_symbols.o):qsslsocket_openssl_symbols.cpp:(.text$_Z9q_BIO_newPK13bio_method_st+0x1): undefined reference to `BIO_new'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qsslsocket_openssl_symbols.o):qsslsocket_openssl_symbols.cpp:(.text$_Z11q_BIO_s_memv+0x1): undefined reference to `BIO_s_mem'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qsslsocket_openssl_symbols.o):qsslsocket_openssl_symbols.cpp:(.text$_Z12q_BN_is_wordP9bignum_stj+0x1): undefined reference to `BN_is_word'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qsslsocket_openssl_symbols.o):qsslsocket_openssl_symbols.cpp:(.text$_Z22q_EVP_CIPHER_CTX_resetP17evp_cipher_ctx_st+0x1): undefined reference to `EVP_CIPHER_CTX_reset'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qsslsocket_openssl_symbols.o):qsslsocket_openssl_symbols.cpp:(.text$_Z18q_EVP_PKEY_base_idP11evp_pkey_st+0x1): undefined reference to `EVP_PKEY_base_id'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qsslsocket_openssl_symbols.o):qsslsocket_openssl_symbols.cpp:(.text$_Z10q_RSA_bitsP6rsa_st+0x1): undefined reference to `RSA_bits'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qsslsocket_openssl_symbols.o):qsslsocket_openssl_symbols.cpp:(.text$_Z10q_DSA_bitsP6dsa_st+0x1): undefined reference to `DSA_bits'
/home/user/mxe/usr/i686-w64-mingw32.static/qt5/lib/libQt5Network.a(qsslsocket_openssl_symbols.o):qsslsocket_openssl_symbols.cpp:(.text$_Z16q_OPENSSL_sk_numP8stack_st+0x1): undefined reference to `OPENSSL_sk_num'
```

Thanks

ArtemPisarenko:
Couple of changes which makes work at least simple Qt5 applications built with CMake.
One of them is Qt patch I [submitted upstream](https://codereview.qt-project.org/c/qt/qtbase/+/286837).

Fixes #1642
Fixes #2430
Fixes #85
Fixes #2166

### DIFF
--- a/src/qtbase.mk
+++ b/src/qtbase.mk
@@ -96,8 +96,9 @@ define $(PKG)_BUILD
         -I'$(1)/test-$(PKG)-pkgconfig' \
         `'$(TARGET)-pkg-config' Qt5Widgets$(BUILD_TYPE_SUFFIX) --cflags --libs`
 
-    # setup cmake toolchain
+    # setup cmake toolchain and test
     echo 'set(CMAKE_SYSTEM_PREFIX_PATH "$(PREFIX)/$(TARGET)/qt5" ${CMAKE_SYSTEM_PREFIX_PATH})' > '$(CMAKE_TOOLCHAIN_DIR)/$(PKG).cmake'
+    $(CMAKE_TEST)
 
     # batch file to run test programs
     (printf 'set PATH=..\\lib;..\\qt5\\bin;..\\qt5\\lib;%%PATH%%\r\n'; \
--- a/src/qtbase.mk
+++ b/src/qtbase.mk
@@ -9,7 +9,7 @@ $(PKG)_CHECKSUM := 4ef921c0f208a1624439801da8b3f4344a3793b660ce1095f2b7f5c4246b9
 $(PKG)_SUBDIR   := $(PKG)-everywhere-src-$($(PKG)_VERSION)
 $(PKG)_FILE     := $(PKG)-everywhere-src-$($(PKG)_VERSION).tar.xz
 $(PKG)_URL      := https://download.qt.io/official_releases/qt/5.14/$($(PKG)_VERSION)/submodules/$($(PKG)_FILE)
-$(PKG)_DEPS     := cc dbus fontconfig freetds freetype harfbuzz jpeg libmysqlclient libpng openssl pcre2 postgresql sqlite zlib zstd jasper libmng tiff libwebp lcms lzma
+$(PKG)_DEPS     := cc dbus fontconfig freetds freetype harfbuzz jpeg libmysqlclient libpng openssl pcre2 postgresql sqlite zlib zstd
 $(PKG)_DEPS_$(BUILD) :=
 $(PKG)_TARGETS  := $(BUILD) $(MXE_TARGETS)
 
--- a/src/qtbase.mk
+++ b/src/qtbase.mk
@@ -9,7 +9,7 @@ $(PKG)_CHECKSUM := 4ef921c0f208a1624439801da8b3f4344a3793b660ce1095f2b7f5c4246b9
 $(PKG)_SUBDIR   := $(PKG)-everywhere-src-$($(PKG)_VERSION)
 $(PKG)_FILE     := $(PKG)-everywhere-src-$($(PKG)_VERSION).tar.xz
 $(PKG)_URL      := https://download.qt.io/official_releases/qt/5.14/$($(PKG)_VERSION)/submodules/$($(PKG)_FILE)
-$(PKG)_DEPS     := cc dbus fontconfig freetds freetype harfbuzz jpeg libmysqlclient libpng openssl pcre2 postgresql sqlite zlib zstd
+$(PKG)_DEPS     := cc dbus fontconfig freetds freetype harfbuzz jpeg libmysqlclient libpng openssl pcre2 postgresql sqlite zlib zstd jasper libmng tiff libwebp lcms lzma
 $(PKG)_DEPS_$(BUILD) :=
 $(PKG)_TARGETS  := $(BUILD) $(MXE_TARGETS)
 
--- a/src/qtbase.mk
+++ b/src/qtbase.mk
@@ -105,14 +105,6 @@ define $(PKG)_BUILD
      printf 'test-qt5.exe\r\n'; \
      printf 'test-qtbase-pkgconfig.exe\r\n';) \
      > '$(PREFIX)/$(TARGET)/bin/test-qt5.bat'
-
-    # add libs to CMake config of Qt5Core to fix static linking
-    $(if $(BUILD_STATIC), \
-        $(SED) -i 's^set(_Qt5Core_LIB_DEPENDENCIES \"\")^set(_Qt5Core_LIB_DEPENDENCIES \"ole32;uuid;ws2_32;advapi32;shell32;user32;kernel32;mpr;version;winmm;z;pcre2-16;netapi32;userenv;zstd\")^g' '$(PREFIX)/$(TARGET)/qt5/lib/cmake/Qt5Core/Qt5CoreConfig.cmake' && \
-        $(SED) -i 's^set(_Qt5Gui_LIB_DEPENDENCIES \"Qt5::Core\")^set(_Qt5Gui_LIB_DEPENDENCIES \"Qt5::Core;dwmapi;winspool;wtsapi32;jasper;mng;tiff;webpdemux;webpmux;d3d11;dxgi;dxguid;harfbuzz;cairo;gobject-2.0;fontconfig;freetype;usp10;msimg32;pixman-1;ffi;expat;bz2;png16;harfbuzz_too;freetype_too;glib-2.0;shlwapi;pcre;intl;iconv;gdi32;comdlg32;oleaut32;imm32;opengl32;mpr;userenv;version;pcre2-16;netapi32;ws2_32;advapi32;kernel32;ole32;shell32;uuid;user32;winmm;lcms2;pthread;webp;zstd;lzma;jpeg;z;m\")^g' '$(PREFIX)/$(TARGET)/qt5/lib/cmake/Qt5Gui/Qt5GuiConfig.cmake' && \
-        $(SED) -i 's^set(_Qt5Network_LIB_DEPENDENCIES \"Qt5::Core\")^set(_Qt5Network_LIB_DEPENDENCIES \"Qt5::Core;ssl;crypto;ws2_32;gdi32;crypt32;iphlpapi\")^g' '$(PREFIX)/$(TARGET)/qt5/lib/cmake/Qt5Network/Qt5NetworkConfig.cmake' && \
-        $(SED) -i 's^set(_Qt5Widgets_LIB_DEPENDENCIES \"Qt5::Gui;Qt5::Core\")^set(_Qt5Widgets_LIB_DEPENDENCIES \"Qt5::Gui;Qt5::Core;gdi32;comdlg32;oleaut32;imm32;opengl32;png16;harfbuzz;ole32;uuid;ws2_32;advapi32;shell32;user32;kernel32;mpr;version;winmm;z;pcre2-16;shell32;uxtheme;dwmapi\")^g' '$(PREFIX)/$(TARGET)/qt5/lib/cmake/Qt5Widgets/Qt5WidgetsConfig.cmake',
-    )
 endef
 
 define $(PKG)_BUILD_$(BUILD)
--- a/src/qtbase-1-fixes.patch
+++ b/src/qtbase-1-fixes.patch
@@ -63,3 +63,102 @@ index 19312d245d..0164f9b49b 100644
 -- 
 2.20.1
 
+
+From b8fd37d8cd7642599b3b022e4d8fa1aff8f68a1e Mon Sep 17 00:00:00 2001
+From: Artem Pisarenko <artem.k.pisarenko@gmail.com>
+Date: Thu, 16 Jan 2020 14:47:00 +0600
+Subject: [PATCH] Fix win32 support for linking against static builds in CMake
+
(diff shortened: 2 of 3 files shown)


src/qtbase.mk
```diff
@@ -9,7 +9,7 @@ $(PKG)_CHECKSUM := 4ef921c0f208a1624439801da8b3f4344a3793b660ce1095f2b7f5c4246b9
 $(PKG)_SUBDIR   := $(PKG)-everywhere-src-$($(PKG)_VERSION)
 $(PKG)_FILE     := $(PKG)-everywhere-src-$($(PKG)_VERSION).tar.xz
 $(PKG)_URL      := https://download.qt.io/official_releases/qt/5.14/$($(PKG)_VERSION)/submodules/$($(PKG)_FILE)
-$(PKG)_DEPS     := cc dbus fontconfig freetds freetype harfbuzz jpeg libmysqlclient libpng openssl pcre2 postgresql sqlite zlib zstd
+$(PKG)_DEPS     := cc dbus fontconfig freetds freetype harfbuzz jpeg libmysqlclient libpng openssl pcre2 postgresql sqlite zlib zstd jasper libmng tiff libwebp lcms lzma
```
mabrand: Aren't these extra deps introduced by qtimageformats rather than qtbase?

ArtemPisarenko: Hmm. Indeed, looks like these deps introduced by separate Qt internal plugin `Qt Image Formats`, which in turn provided by `qtimageformats` package in MXE. But if so, then my app build should fail due to missing plugin itself. I did neither install this package, nor Qt application being build has dependency on corresponding plugin. Only `qtbase`. And only after I installed these extra packages it was build successfully. Otherwise o failed at link stage with `cannot find -l<...>` message for each extra dep. That's why I decided they correspond to qtbase.
I guess we discovered another Qt bug(feature?). Libraries list for linker is being generated by mechanism implemented in qtbase which tracks dependencies for all qt addon modules and plugins not accounting whether they are actually used in specific app build they generated for, i.e. module just "reports" dependencies of all plugins it uses, and although app isn't actually link with them at runtime and may be built, these libs are stupidly passed to linker which complains, because cannot find them. But that's just a guess, because otherwise I don't have any other explanation. Need to check...

ArtemPisarenko: Ok, I've got the point. It turned out to be a documented feature rather than bug and, of course, it's more general than this specific `QtCore -> Qt Image Formats` dependency relation. [Qt Cmake documentations](https://doc.qt.io/qt-5/qtcore-cmake-qt5-import-plugins.html) describe it:

> ...
> Qt provides plugin types such as **imageformats**, platforms, and sqldrivers.
> If `qt5_import_plugins()` isn't called, the target **automatically links against a sane set of default plugins**, for each Qt module that the target is linked against.
> ...
> Note: This macro imports plugins from static Qt builds only. On shared builds, it does nothing.
> ...

Actually this behavior isn't much different from what I guessed above, because it doesn't link against plugin itself but does against its dependencies, so I would consider this a bug. But the behavior itself imposes a problem into MXE, regardless of this bug being fixed or not, because now MXE Qt packages dependencies needs to be rewritten in a way I can't imagine of - for each MXE package containing such a default plugin (where decision of being default or not, depends on Qt internal logic, possibly version-specific) there is bidirectional dependency between it and a package containing corresponding module (such as `qtbase`). Maybe it makes sense to find out how similar problem being solved in other packaging systems such as Linux distributions...

tonytheodore: > If qt5_import_plugins() isn't called, the target automatically links against a sane set of default plugins

I'd say we should set this explicitly (maybe in the toolchain file includes) and investigate how subsequent calls in project files interact with that setting.

tonytheodore: > I'd say we should set this explicitly (maybe in the toolchain file includes)

Ignore that, it's an executable-specific setting.

A [simple cmake test](https://github.com/ArtemPisarenko/mxe/pull/1) builds and runs fine without the extra image deps or setting `qt5_import_plugins`.

ArtemPisarenko: Yes, it turned out to be I've got things wrong again :) Seems like commit c943c9b changes Qt behavior and, indeed, these changes are inapropriate now.

I've tested all cominations of build/run Qt applications:
- app doesn't depend on any non-default Qt plugin (similar to simple test you added);
- app using extra Webp image format, having `qt5_import_plugins(mxe_qt_app INCLUDE_BY_TYPE imageformats Qt5::QWebpPlugin)` in cmake file and calling `QImage image(100, 100, QImage::Format_RGB32); Q_ASSERT(image.save("image.webp"));`;
- linking against static Qt;
- linking against shared Qt;
- building/deploying from MXE environment which has only `qtbase` package installed;
- building/deploying from MXE environment which has `qtbase` and `qtimageformats` packages installed.

App is being built and run successfully in all combinations above and its second variant doesn't fail at Q_ASSERT in runtime in combinations where it being built/deployed from MXE with `qtimageformats` installed. In static variants Qt cmake configuration is smart enough to import only those plugins it found itself.

I personally like this behavior. Although, correct deploying app with plugins isn't simple, especially for beginers, because app doesn't report any diagnostic info why it failed (although maybe I just don't know how to debug it). I mean that it isn't enough to collect DLLs dependencies from app .exe only and copy necessary plugins, but also each plugin DLL dependencies must be collected too, e.g. in my test I have to do `mkdir deploy && (x86_64-w64-mingw32.shared-peldd mxe_qt_app.exe /mxe/usr/x86_64-w64-mingw32.shared/qt5/plugins/imageformats/qwebp.dll -a -w d3d11.dll -w dxgi.dll | xargs cp -t deploy/) && cp -r /mxe/usr/x86_64-w64-mingw32.shared/qt5/plugins/platforms /mxe/usr/x86_64-w64-mingw32.shared/qt5/plugins/imageformats deploy/`

tonytheodore:
Looks good to me, all downstream packages build (with a couple of unrelated fixes for Qt 5.14 that I'll push shortly). I'll test some known projects like openscad and pulseview, but this is good to go.

Thanks!

luntik2012:
For now I cannot link with qquickcontrols2 using mxe/cmake, getting `undefined reference to `qt_static_plugin_*`

(I'm trying to do it like this https://github.com/luntik2012/qml_hello_world/blob/master/CMakeLists.txt)

PR #2456, opened 2020-01-29
Switch to pkgconf at more places in the docs

vog:
Switch to pkgconf at more places in the docs. In particular, list pkgconf instead of pkg-config in the list of dependencies.

Moreover, use pkgconf instead of pkg-config for the Debian installation command. The old command switched Debian systems back from pkgconf to pkg-config, which is certainly not what we want.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -729,7 +729,7 @@ USE_OSGPLUGIN(&lt;plugin2&gt;)
     p7zip-full \
     patch \
     perl \
-    pkg-config \
+    pkgconf \
     python \
     ruby \
     sed \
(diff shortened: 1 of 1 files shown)


tonytheodore:
Thanks for looking at this, I'm not certain either is a direct requirement any longer - I'll do some testing on a minimal debian install.

vog:
Indeed, that would be interesting to know. Thanks for checking!

tonytheodore:
This was removed on 2014-10-29[1], but the docs weren't updated.

  - cross packages use the prefixed ${target}-pkg-config wrapper
  - native packages use usr/${build}/bin/pkgconf
  - native reqs may install whichever pkg-config the package manager
    specifies, but it isn't used by mxe


[1] a063f82#diff-b67911656ef5d18c4ae36cb6741b7965

PR #2459, opened 2020-02-17
Two fixes

fiesh:
In an attempt to fix #2447 I ran into some minor issues.  Two of them are fixed here.

In general, this begs the following question: Would it be impossible to have a pipeline that builds everything?  We handle our internal MXE docker image like this, but of course we don't build everything.  It seems that MXE could really use a CI pipeline that ensures everything can be built in a certain environment.  But maybe that should be discussed in a separate issue. I would be happy to provide a sample Dockerfile.

### DIFF
--- a/src/qtifw.mk
+++ b/src/qtifw.mk
@@ -6,7 +6,7 @@ $(PKG)_DESCR    := Qt Installer Framework
 $(PKG)_IGNORE   :=
 $(PKG)_VERSION  := 3.2.0
 $(PKG)_CHECKSUM := 7cd0d7b86fbe63b101f263b6532485df3a2d336cc0026b47eb866dae134e319f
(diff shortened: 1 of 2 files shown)


tonytheodore:
> Would it be impossible to have a pipeline that builds everything?

It's not impossible, but takes a lot of processing resource. At the moment, I try to run testing locally overnight and then on a cloud VM on the weekend. Of course, life intrudes and that gets missed.

I've just received some [AWS open source credits](https://lists.nongnu.org/archive/html/mingw-cross-env-list/2020-02/msg00002.html) specifically for this purpose, so we should be able to make some progress. 

> I would be happy to provide a sample Dockerfile.

Any help with setting up a CI environment is welcome.

Thanks for the fixes!

PR #2486, opened 2020-04-16
The default msvcrt must be the same for mingw-w64 headers and crt.

kalibera:
The default msvcrt must be the same for ming-w64 headers and crt, so it is better to explicitly set both in the rules file, not just one.


tonytheodore:
Thanks, I've got this in [another set of changes](https://github.com/tonytheodore/mxe/commit/674dbde76ab4ced84af450021674c0bffe6cfbfb#diff-d30e6bdb5660ed818ad8ed6aa4a177e3R68), but that will take some time.

PR #2491, opened 2020-05-03
fribidi: bump to version 1.0.8

lotharsm:
Bumping the FriBiDi lib to version 1.0.8.

The latest 1.0.9 archives seems to be broken. The .tar.bz2 archive got removed, the .tar.xz seems to be corrupt and the .tar.gz and .zip archives seem to be incomplete. I'll try to contact the author, so we might be able to get an updated build.

For now, this is as current as we can get :)

### DIFF
--- a/src/fribidi.mk
+++ b/src/fribidi.mk
@@ -4,9 +4,9 @@ PKG             := fribidi
 $(PKG)_WEBSITE  := https://fribidi.org/
 $(PKG)_DESCR    := FriBidi
 $(PKG)_IGNORE   :=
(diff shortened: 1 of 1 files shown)


tonytheodore:
> the .tar.gz and .zip archives seem to be incomplete

These typically require an `autoreconf` step to generate the files that aren't kept under source control.

> For now, this is as current as we can get :)

Thanks!

danger89:
Just curious, why is MXE saying `MXE Warning! Downloading fribidi from second URL.`?

PR #2493, opened 2020-05-03
libvpx: bump version to 1.8.2

lotharsm:
This patch updates libvpx to the current version 1.8.2. Switching to github, since the files seem to be not available through the Google Storage anymore. Also dropping 3 patches since the files do not exist anymore.


tonytheodore:
One of the patched files has been renamed (`gen_msvs_proj.sh --> gen_msvs_vcxproj.sh`), but they [seem to still exist](https://github.com/webmproject/libvpx/tree/master/build/make) and have `#!/bin/bash`.

lotharsm:
Thanks for the heads up, I’ll fix it. I wonder if those are necessary though, or if we could omit those patches since it seems to have no real effect since we are not using the MSVC files...
    	
    	

    	
    
  




On Mon, May 4, 2020 at 1:48 PM +0200, "Tony Theodore" <notifications@github.com> wrote:















One of the patched files has been renamed (gen_msvs_proj.sh --> gen_msvs_vcxproj.sh), but they seem to still exist and have #!/bin/bash.

tonytheodore:
Best option is to report it upstream and remove the patches, they have other [attempts](https://github.com/webmproject/libvpx/commit/3526f1cd5e24006b82b9b26062993d74ee4d73b7) at [removing `bashisms`](https://github.com/webmproject/libvpx/commit/eb2fbea62131aa5cb3b3c0886308e0e7620bcbb2) and should be open to fixing it. At very least, they could require `bash` and rely on `#!/usr/bin/env bash` to find it.

lotharsm:
I re-added and fixed the patch with the correct filename. It turned out that only one filename was changed. I'll try to get in touch with the upstream devs.

tonytheodore:
Thanks!

PR #2502, opened 2020-05-22
gdk-pixbuf: Enable dynamic modules in shared builds

janbrummer:
GTK3 makes use of svg icons, but required librsvg module cannot be build in static
gdk-pixbuf so allow dynamic modules for shared builds.


tonytheodore:
Thanks!

Issue #2499, opened 2020-05-19
add rust compiler


alex-tee:
I was trying to update librsvg but it needs rust

kleisauke:
Here is an example for building the latest librsvg with Rust:
https://github.com/libvips/build-win64-mxe/blob/master/8.9/rust.mk
https://github.com/libvips/build-win64-mxe/blob/master/8.9/rust-std-x86_64.mk
https://github.com/libvips/build-win64-mxe/blob/1964b876290d151c124abd80902a7b12fe464184/8.9/overrides.mk#L541-L568
https://github.com/libvips/build-win64-mxe/blob/1964b876290d151c124abd80902a7b12fe464184/8.9/patches/librsvg-2-fixes.patch

I could try to make a PR, but I'm not sure if there's any interest in it.

alex-tee:
thanks

i'm packaging [zrythm](https://www.zrythm.org) and i'm having rendering issues with some SVGs. using the latest librsvg would likely fix it. merging this upstream would probably be helpful for GTK apps using SVGs

yutannihilation:
Let me ask a super noob question. There's already a Rust package on MSYS2, so I naively guess it won't be hard to add the Rust compiler to MXE. What's the blocker here?

https://packages.msys2.org/base/mingw-w64-rust

PR #2409, opened 2019-10-03
Bye bye winthreads; relating to #2385

Lord-Kamina:
So, here's a big one...

I removed essentially all instances of hardcoded win32 threads; adding the necessary patches where required.

Additionally, I updated some of the packages as well, most notably boost which is now at 1.71.0 and building properly against pthreads. 🎉

Issue #2330, opened 2019-05-05
Add support to ARM and ARM64 Windows targets.


woachk:
Related to #558.

ARM and ARM64 Windows targets depend on LLVM-MinGW https://github.com/mstorsjo/llvm-mingw, and aren't usable with a binutils + GCC based toolchain.

kleisauke:
fwiw, I've created a MXE-plugin that uses https://github.com/mstorsjo/llvm-mingw as toolchain, see:
https://github.com/libvips/build-win64-mxe/tree/master/8.8/plugins/llvm-mingw

I'm considering making a PR to integrate it into MXE as well. Is there any interest in this?

(Still need to test the [`armv7-w64-mingw32` and `aarch64-w64-mingw32` targets](https://github.com/libvips/build-win64-mxe/blob/master/8.8/plugins/llvm-mingw/llvm-mingw.mk#L81-L83) but I don't have the necessary hardware to test it on, unfortunately)

lunixbochs:
@kleisauke I am in favor of an in-tree LLVM plugin, I keep running into issues with GCC

boswelja:
Any updates on this?

tonytheodore:
Adding clang is the next thing on my list now, gcc doesn't yet support newer Macs with M1 chips.

The [repo above](https://github.com/libvips/build-win64-mxe/tree/master/build/plugins/llvm-mingw) looks very promising, thanks @kleisauke.

markghayden:
Is there any progress on this?  The Windows ARM-based Surface tablets are becoming more common and we'd like to compile for them.

PR #2503, opened 2020-05-22
librsvg: Enable module generation for shared build

janbrummer:


Lord-Kamina:
I forgot to commit it but I have a fork somewhere to build the newer librsvgs requiring rust.

kleisauke:
@Lord-Kamina I've also build the latest librsvg that requires Rust within MXE, see: https://github.com/mxe/mxe/issues/2499#issuecomment-633209023.

I could try to make a PR, but I'm not sure if there's enough interest in this.

As an aside, thanks for the fixes at OpenEXR!

Lord-Kamina:
> @Lord-Kamina I've also build the latest librsvg that requires Rust within MXE, see: [#2499 (comment)](https://github.com/mxe/mxe/issues/2499#issuecomment-633209023).
> 
> I could try to make a PR, but I'm not sure if there's enough interest in this.
> 
> As an aside, thanks for the fixes at OpenEXR!

Thanks. That was borne as a side project from #2409 

I think your approach to rust as very similar to mine; however, mine has the merit of working on more platforms so we should probably combine them. I'll upload my branch later.

kleisauke:
Let me know when you've uploaded it. Do you build Rust from scratch? 

I'm using the pre-built Rust libraries which has disadvantages in terms of compatibility. Only the [`x86_64-w64-mingw32.*`](https://github.com/libvips/build-win64-mxe/blob/master/8.9/rust-std-x86_64.mk) and [`i686-w64-mingw32.*`](https://github.com/libvips/build-win64-mxe/blob/master/8.9/rust-std-i686.mk) targets are supported. The [`aarch64-w64-mingw32.*`](https://github.com/libvips/build-win64-mxe/blob/master/8.9/rust-std-aarch64.mk) and [`armv7-w64-mingw32.*`](https://github.com/libvips/build-win64-mxe/blob/master/8.9/rust-std-armv7.mk) LLVM targets (see #2330) are currently unsupported. It may also cause problems when targeting UCRT since the pre-built Rust libraries are built using a MinGW-w64 compiled with `--with-default-msvcrt=msvcrt`.

Lord-Kamina:
> Let me know when you've uploaded it. Do you build Rust from scratch?
> 
> I'm using the pre-built Rust libraries which has disadvantages in terms of compatibility. Only the [`x86_64-w64-mingw32.*`](https://github.com/libvips/build-win64-mxe/blob/master/8.9/rust-std-x86_64.mk) and [`i686-w64-mingw32.*`](https://github.com/libvips/build-win64-mxe/blob/master/8.9/rust-std-i686.mk) targets are supported. The [`aarch64-w64-mingw32.*`](https://github.com/libvips/build-win64-mxe/blob/master/8.9/rust-std-aarch64.mk) and [`armv7-w64-mingw32.*`](https://github.com/libvips/build-win64-mxe/blob/master/8.9/rust-std-armv7.mk) LLVM targets (see #2330) are currently unsupported. It may also cause problems when targeting UCRT since the pre-built Rust libraries are built using a MinGW-w64 compiled with `--with-default-msvcrt=msvcrt`.

No, I also use pre-built. Initially I attempted to build from scratch but besides taking hours, you couldn't mix-and-match due to rust not being abi-stable, so essentially you needed yo rebuild everything everytime and it was a mess.

tonytheodore:
Thanks!

I'll try to look at rust in the next few weeks.

PR #2507, opened 2020-05-26
[PATCH] Fixed abscence of libzstd.dll file at bin directory

Vitozz:

  * install .pc file,

Thanks!


tonytheodore:
I've submitted both [the current patch](https://github.com/facebook/zstd/pull/2183) and [this one](https://github.com/facebook/zstd/pull/2185) upstream (after seeing the CLA delays). Small edits can be done in github and don't require any local repos.

With those, we can [consolidate the patches easily](https://github.com/mxe/mxe/commit/cdabfdc85239267f07efbb0fb51145fac1b741a0):
```sh
make init-git-zstd
cd gits/zstd-1.4.5
curl 'https://patch-diff.githubusercontent.com/raw/facebook/zstd/pull/2183.patch' | git am
curl 'https://patch-diff.githubusercontent.com/raw/facebook/zstd/pull/2185.patch' | git am
git rebase dist -i
# reword commits with patch urls
cd ../..
make export-patch-zstd
```

PR #2510, opened 2020-05-30
VTK+gcc-10.1 fix: XMLPUBVAR: patch the mingw block, not the cygwin

janvdijk:
(A small follow up on (completed) pull request #2506)

In a small part of the patch that is now in MXE, somehow the line
numbers got messed up. The result is that the XMLPUBVAR is now
fixed for the cygwin case, not for mingw. This patch fixes that. It adds
'extern' to the __declspec(dllexport) definition of XMLPUBVAR.

The result is what is also in upstream libxml2, and in the
stand-alone version that is shipped with mxe. For details, see:
https://github.com/GNOME/libxml2/commit/1eabc31401b7b8c3b5273993778f37eeef37a055

Please merge this fix. And sorry for the inconvenience.


janvdijk:
@mabrand, since you pull in my change set,  could you please take a look at this tiny follow-up? I apologize for the inconvenience, I am sure the error was mine.

PR #2511, opened 2020-05-31
fltk: update 1.3.4-2 --> 1.3.5

blueghost-tbs:
This new version was released more than 1 year ago and contains important bug fixes.


tonytheodore:
Thanks!

Issue #2476, opened 2020-04-05
luajit compile error


dbyoung720:
luajit compile error
[luajit_i686-w64-mingw32.zip](https://github.com/mxe/mxe/files/4432788/luajit_i686-w64-mingw32.zip)

tonytheodore:
Strange, I'm seeing a different error though nothing has changed recently.

tonytheodore:
Building on Ubuntu 16.04 still succeeds.

dbyoung720:
I compile on wsl , retry three time, error.

gitfxt:
I got the same error on wsl ubuntu 18044.
Is the OS problem?

ZachBacon:
I'm definitely experiencing the same issue on ubuntu 20.04 wsl, the log seems to be erroring out after building minilua, seems it's trying to build something with it that caused a syntax error. https://gist.github.com/ZachBacon/f85c092e0f1bf06dc22d2e3aaca9c7e8

tonytheodore:
I've recently upgraded pkg.mxe.cc to 20.04 and it's working there.

There are [some env vars](https://repo.or.cz/luajit-2.0.git/blob/0bf80b07b0672ce874feedcc777afe1b791ccb5a:/src/Makefile#l159) that might be causing an issue on WSL.

tonytheodore:
Also look at https://mxe.cc/#requirements-wsl, it may or may not be related.

ZachBacon:
Don't think so, everything is setup under /home/username, as for the temp. Perhaps, but I doubt it.

tonytheodore:
I'll try installing WSL and see if I can replicate it.

ZachBacon:
It's a weird and a unique one, I'm not really looking so much as a solution since I generally prefer a linux box over wsl2, but I still find it handy from time to time.

tonytheodore:
```
/usr/include/limits.h:26:10: fatal error: bits/libc-header-start.h: No such file or directory
 #include <bits/libc-header-start.h>
```
Bizarrely, there is no `/usr/include/bits` directory on WSL, on a standard install it's a symlink:
```
$ ll /usr/include/bits
lrwxrwxrwx 1 root root 21 Aug 28  2014 /usr/include/bits -> x86_64-linux-gnu/bits/
```

tonytheodore:
Now I get it:
```
$ dpkg -S include/bits
libc6-dev-i386: /usr/include/bits
```
[...]

PR #2512, opened 2020-06-02
Move ocaml packages to plugin

tonytheodore:
These packages are heavily patched and very difficult to maintain and
update. They're starting to fail on various linux systems and it
doesn't seem worth trying to fix the current versions.

Move to a plugin in case anyone wants to revisit and take ownership
of them.


vog:
If we do that, I propose to make a visible announcement on the mailing list as well, just to be sure. Maybe that motivates someone to step up.

tonytheodore:
Seems there is [another package (luajit) that enforces 32-bit native code for 32-bit cross targets](https://github.com/mxe/mxe/issues/2476#issuecomment-647011418), so the multilib reqs can't be simply removed.

Scorbie:
Is it possible to warn this when installing OCaml in the homepage until this is merged?
I was misleaded that ocaml% packages are buildable if I setup multi-lib compiler environment.
https://mxe.cc/#requirements (See Issues without a 32-bit compiler)
(BTW, I think luajit should also be included there)

jonaski (approved):

PR #2514, opened 2020-06-03
librsvg: Update to 2.46.4; also adds rustc/cargo/rust-std packages.

Lord-Kamina:
As I mentioned in #2503, I made this a while ago and forgot to make a PR. I'm using a fairly similar approach as the one @kleisauke posted in https://github.com/mxe/mxe/issues/2499#issuecomment-633209023

My rust is older than the version he used (because I originally worked on this PR back near the end of 2019); updating the RUST packages will likely prove to be a bit of work but then again, I'm not really sure they need to be updated all that often. IIRC, I downloaded a directory listing and then used Regex and perhaps a bash script to get the hashes of the relevant files; the idea being to be able to potentially support as many host systems as possible (I developed the PR running MXE on macOS, for example)

Anyway... it might have a few rough edges but it builds and is functional. One doubt I do genuinely have, though, is I'm not sure it will properly build with gcc 5.5.0, since I used 8.3.0


Lord-Kamina:
I just noticed this apparently doesn't work on linux; but it works perfectly on mac.
Anybody have an idea as to why this might be happening?

It seems the issue on linux is it isn't downloading the source files, so it tries to extract them and obviously there's nothing there.

On linux I'm also getting 
```
bash: line 0: [: too many arguments
bash: line 0: [: too many arguments
```
errors that are not present in mac, but I don't know a) whether they are relevant to this and b) where they're coming from.

tonytheodore:
The BUILD_TRIPLET detection is better achieved using [rust's own naming conventions](https://github.com/rust-lang/rustup/pull/2375) and adding that file to our `ext` directory (the same as we do with `config.guess`).

I'll think about multiple files/checksums.

Lord-Kamina:
> The BUILD_TRIPLET detection is better achieved using [rust's own naming conventions](https://github.com/rust-lang/rustup/pull/2375) and adding that file to our `ext` directory (the same as we do with `config.guess`).
> 
> I'll think about multiple files/checksums.

I'm using their conventions. The idea essentially is to try and convert whatever mxe says to rust triplets. I found the bug BTW; rust does not have pc-linux, only unknown-linux.

So, changing 
``` diff
ifneq (, $(findstring darwin,$(BUILD)))
BUILD_TRIPLET = $(firstword $(call split,-,$(BUILD)))-apple-darwin
else
-    ifneq (, $(findstring ibm-linux,$(BUILD)))
+    ifneq (, $(findstring linux,$(BUILD)))
    BUILD_TRIPLET = $(firstword $(call split,-,$(BUILD)))-unknown-linux-gnu
    else
    BUILD_TRIPLET = $(BUILD)
    endif
endif
```
Fixes my logic.

Lord-Kamina:
Mind you @tonytheodore; the get-triplet-logic on rust-std.mk is still shaky. i.e. it works but only sometimes (you sometimes need to remove the install file and try again or even make a failed attempt and retry) because it's a really nasty hack. I'm of course open to better solutions for that, or ideally if it's possible, solve the core issue in mxe that necessitated that fix, i.e.:

` $(TARGET)` is not available outside the `$(PKG)_BUILD` functions.

src/rust-std.mk
```diff
@@ -0,0 +1,49 @@
...
+else
+    ifneq (, $(findstring linux,$(BUILD)))
+        BUILD_TRIPLET = $(firstword $(call split,-,$(BUILD)))-unknown-linux-gnu
+    else
+        BUILD_TRIPLET = $(BUILD)
+    endif
+endif
+
+define DEP_TARGET
+$(shell [[ -f $(MXE_TMP)/get-target-triplet.txt ]] && cat $(MXE_TMP)/get-target-triplet.txt || echo ERROR)
```
JL2210, line 20: If using double-brackets is an issue you can quote the argument like so:
```
[ -f "$(MXE_TMP)/get-target-triplet.txt" ]
```

Lord-Kamina: Actually yeah, I'm not entirely sure why I used the double brackets instead of the alternate test format.

danger89:
Maybe update with upstream master?

Issue #2517, opened 2020-06-08
[PATCH] Qt 5.15 fails to build on Ubuntu 18.04


mbunkus:
I've just filed [QTBUG-84792](https://bugreports.qt.io/browse/QTBUG-84792) but thought it would be good to have here, too:

Qt 5.15 fails to build on Ubuntu 18.04 due to `zstd` being too old. The build process uses the option `--single-thread` when invoking `zstd`, and the version coming with Ubuntu 18.04 simply doesn't support that one yet.

It affects all architectures, both static & dynamic versions. My MXE is at f5fe0a81. I'm using the gcc 10 overlay, though that should be of no consequence to the failure in question.

Relevant error message is:

```
perl /opt/mxe/tmp-qtbase-x86_64-w64-mingw32.static/qtbase-everywhere-src-5.15.0/src/corelib/mimetypes/mime/generate.pl --zstd mimetypes/mime/packages/freedesktop.org.xml > .rcc/release/qmimeprovider_database.cpp
Incorrect parameters
Failed to run zstd -cq19 --single-thread < mimetypes/mime/packages/freedesktop.org.xml at /opt/mxe/tmp-qtbase-x86_64-w64-mingw32.static/qtbase-everywhere-src-5.15.0/src/corelib/mimetypes/mime/generate.pl line 82.
```

I'm now using [this small additional patch](https://gist.github.com/mbunkus/529f6799b7e13ad57fc82211bcb2a980) which simply patches the option out.

tonytheodore:
I wonder if we should simply build the native zstd binary, the [lib is now used by gcc10](https://gcc.gnu.org/install/prerequisites.html) so it will generally be available. It also has a Makefile so doesn't really require cmake for the native build.

tonytheodore:
@mbunkus, see #2520.

You can avoid a full rebuild with:
```
make -t qtbase
touch src/zstd.mk
make qtbase
```

mbunkus:
Thanks, I'll give it a try once #2522 has been resolved.

mbunkus:
#2520 fixed the issue for me. Thanks!

PR #2520, opened 2020-06-09
Enable native zstd bins for qtbase and libs for gcc10

tonytheodore:
WIP because the gcc detection needs more overrides to actually use zstd.

@mbunkus I'm pretty sure the [qtbase issue](https://github.com/mxe/mxe/issues/2517) can be tested without a full rebuild - those overlay plugins don't use the [undocumented](https://github.com/mxe/mxe/commit/24d37efa5dae04fbe143c33b8c6710a8606c5451#diff-fffe09b3b50640845a354114057e2a80R7) addition of [_FILE_DEPS](https://github.com/mxe/mxe/commit/24d37efa5dae04fbe143c33b8c6710a8606c5451#diff-b67911656ef5d18c4ae36cb6741b7965R528).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -507,9 +507,9 @@ PKG_DEPS = \
                       $(filter $($($(DEP)_PKG)_TYPE),$(BUILD_PKG_TYPES))), \
                 $($(DEP)_TGT)/installed/$($(DEP)_PKG))))
 
-# order-only package deps unlikely to need target lookup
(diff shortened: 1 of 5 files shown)


tonytheodore:
Turns out zstd is only used in gcc10, even though the [docs were updated before the 9.2 release](https://web.archive.org/web/20190722051344/http://gcc.gnu.org:80/install/prerequisites.html).

PR #2532, opened 2020-06-23
This allows the msw theme to be built in, 

ZachBacon:
but it doesn't allow themes in general to be enabled.

This also condenses a lot of the patches a bit that goes over certain files over again, but the important thing though is that the msw theme is built in.

### DIFF
--- a/src/gtk2-1-fixes.patch
+++ b/src/gtk2-1-fixes.patch
@@ -6,222 +6,10 @@ From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: "fix@me" <fix@me>
 Date: Fri, 24 Sep 2010 14:43:43 +0200
 Subject: [PATCH 01/12] s,DllMain,static _disabled_DllMain,
-
-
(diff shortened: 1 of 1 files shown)


tonytheodore:
> This also condenses a lot of the patches a bit

That is the exact opposite of what the 12 discrete patches are trying to achieve. We use a git-style workflow to patches:
```
make init-git-gtk2
make import-patch-gtk2
# apply your changes
make export-patch-gtk2
```
That "apply your changes" could be dropping or replacing any of the previous commits, or simply adding to the patch chain.

To test the intermediate results of the working copy you can run:
```
make gtk2 gtk2_SOURCE_TREE=gits/gtk+-2.24.29
```
That will bypass any patches and build against the source tree, committed or not. When you're happy with your changes, you can commit and update the patch with
```
make export-patch-gtk2
#optional but recommended
make clean
make gtk2
```

tonytheodore:
> apply your changes

Update:
```
# apply your changes in gits/gtk+-2.24.29
```

PR #2533, opened 2020-06-23
nsis: Unicode builds, fix x86_64 build.

Lord-Kamina:
Here we go.

### DIFF
--- a/src/nsis.mk
+++ b/src/nsis.mk
@@ -18,14 +18,17 @@ define $(PKG)_UPDATE
 endef
 
 define $(PKG)_SCONS_OPTS
-    XGCC_W32_PREFIX='$(TARGET)-' \
+    CCACHE_DISABLE=1 \
+    PATH='$(PREFIX)/bin:$(PATH)' \
     PREFIX='$(PREFIX)/$(TARGET)' \
-    `[ -d /usr/local/include ] && echo APPEND_CPPPATH=/usr/local/include` \
-    `[ -d /usr/local/lib ]     && echo APPEND_LIBPATH=/usr/local/lib` \
-    $(if $(findstring x86_64-w64-mingw32,$(TARGET)),\
-        SKIPPLUGINS='System' TARGET_ARCH=amd64) \
+    XGCC_W32_PREFIX='$(TARGET)-' \
+    APPEND_CPPPATH="['$(PREFIX)/$(TARGET)/include'$(shell [ -d /usr/local/include ] && echo , \'/usr/local/include\')]" \
+    APPEND_LIBPATH="['$(PREFIX)/$(TARGET)/lib'$(shell [ -d /usr/local/lib ] && echo , \'/usr/local/lib\')]" \
+    TARGET_ARCH=$(if $(findstring x86_64,$(TARGET)),amd64,x86) \
+    TARGET_OS='windows' \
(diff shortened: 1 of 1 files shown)


src/nsis.mk
```diff
@@ -18,25 +18,27 @@ define $(PKG)_UPDATE
...
+    CCACHE_DISABLE=1 \
+    PATH='$(PREFIX)/bin:$(PATH)' \
     PREFIX='$(PREFIX)/$(TARGET)' \
-    `[ -d /usr/local/include ] && echo APPEND_CPPPATH=/usr/local/include` \
-    `[ -d /usr/local/lib ]     && echo APPEND_LIBPATH=/usr/local/lib` \
-    $(if $(findstring x86_64-w64-mingw32,$(TARGET)),\
-        SKIPPLUGINS='System' TARGET_ARCH=amd64) \
+    XGCC_W32_PREFIX='$(TARGET)-' \
+    CPPPATH_APPEND="['#$(PREFIX)/$(TARGET)/include'$(shell [ -d /usr/local/include ] && echo , \'#/usr/local/include\')]" \
+    LIBPATH_APPEND="['#$(PREFIX)/$(TARGET)/lib'$(shell [ -d /usr/local/lib ] && echo , \'#/usr/local/lib\')]" \
```
tonytheodore, line 26: - why rename `APPEND_CPPPATH` -> `CPPPATH_APPEND`:
```
% grep -r CPPPATH_APPEND gits/nsis-3.05-src
# no result
% grep -r APPEND_CPPPATH gits/nsis-3.05-src
gits/nsis-3.05-src/SConstruct:opts.Add(PathVariable('APPEND_CPPPATH', 'Additional paths to search for include files', None))
gits/nsis-3.05-src/SConstruct:defenv.Append(CPPPATH = Split('$APPEND_CPPPATH'))
```
  - what are the hashes for? I'm used to switching between shell and make, but I can't parse those lines.

tonytheodore:
Looking good - building on all four targets with gcc[5|8|10].

I just don't understand the `[CPP|LIB]PATH_APPEND` changes. For background, those are FreeBSD compatibility fixes and can be moved to comments till someone confirms they are still relevant.

src/nsis.mk
Lord-Kamina: > * why rename `APPEND_CPPPATH` -> `CPPPATH_APPEND`:

Because I'm dumb and/or had slept too little.

> ```
> % grep -r CPPPATH_APPEND gits/nsis-3.05-src
> # no result
> % grep -r APPEND_CPPPATH gits/nsis-3.05-src
> gits/nsis-3.05-src/SConstruct:opts.Add(PathVariable('APPEND_CPPPATH', 'Additional paths to search for include files', None))
> gits/nsis-3.05-src/SConstruct:defenv.Append(CPPPATH = Split('$APPEND_CPPPATH'))
> ```
> 
> * what are the hashes for? I'm used to switching between shell and make, but I can't parse those lines.

They're probably not strictly needed but I added them when nothing else was working (can't remember why); anyway, the `#` tells scons that it should use the top-level directory as reference; although I don't really think that's relevant for absolute paths.

tonytheodore: Given that `scons` wasn't seeing them at all, can you remove those lines? 

I'm happy if you pop them out to a comment since it's not clear they are relevant any longer.

PR #2534, opened 2020-06-24
This bakes in the msw theme into gtk2

ZachBacon:
This is a revised PR with proper patch support that allows the msw theme be baked into gtk2 itself


ZachBacon:
Bump

ZachBacon:
bump

ZachBacon:
just doing one more bump

src/gtk2-1-fixes.patch
```diff
@@ -567,3 +567,161 @@ index 1111111..2222222 100644
...
+     }
+ }
+ 
+-G_MODULE_EXPORT void
+-theme_init (GTypeModule * module)
++#ifdef G_OS_WIN32
++#define MODULE_ENTRY(type, function) type _gtk_theme_wimp_ ## function
++#else
++#define MODULE_ENTRY(type, function) G_MODULE_EXPORT type theme_ ## function
++#endif
```
mabrand, line 691: Are you sure G_MODULE_EXPORT isn't needed for G_OS_WIN32? Have you tested static and shared builds?

ZachBacon: This was mostly based on an old patch that I migrated to a newer gtk, however as for 64bit shared builds, gtk2 never built for me in shared form period so even before this patch it still didn't exactly compile period in that form on any of my rigs. My test case showed it does work, but I have not extensively tested it outside of static builds.

PR #2565, opened 2020-09-25
SE-3395 Draco

atciiiatx:
Add MXE support for the Draco compression library


fiesh:
On a side note, we've been using draco for our projects integrated as a `meson` subproject for quite a while now, works like a charm also in the MXE pipeline.

(There's also a more recent 1.4.1 by now.)

PR #2574, opened 2020-11-04
binutils: update to 2.36.1

orgads:


orgads:
@mabrand ?

orgads:
ping

tonytheodore:
At the moment, these versions (along with #2576, #2577, #2579, and #2580) are tied to the default compiler. 

That will be changing soon, but in the mean time, newer versions should be in their respective plugins.

orgads:
Why? I successfully built the older compilers with these versions.

PR #2580, opened 2020-11-04
isl: update to 0.18

orgads:


orgads:
ping

jonaski:
I don't think they want to update isl to the latest version. I have previously submitted a PR for the same that was ignored for a year.

orgads:
The latest version is 0.22 IIRC. I picked 0.18 because it is compatible with all the supported versions of gcc (at least build-wise).

jonaski:
I see. I wasn't aware, because the latest on https://gcc.gnu.org/pub/gcc/infrastructure/ is 0.18.
But I see it now when I visit https://isl.gforge.inria.fr/

tonytheodore:
> compatible with all the supported versions of gcc (at least build-wise)

This is the key - gcc will build, possibly all of mxe. Will any downstream users of mxe be affected (or worse, silently affected)?

I haven't seen issues with gmp, mpc, or mpfr - but have seen issues with binutils, isl, and cloog (previously). It's preferable to keep the toolchain deps stable after they've been through a period of validation.

To wit, binutils has two issues since 2.35 that are exposed by building all of mxe (https://github.com/mxe/mxe/commit/ef9b92b82ef9d6118d5b8bbd12198861479137b4 and https://github.com/mxe/mxe/commit/40cdd2a78f4110b857406f353f583384ecf6b6be). I'm happy for those issues to be worked through in the latest gcc plugins, but not in the default.

The default gcc is, of course, the main issue. I've restarted my target-specified toolchain work and should have some WIP to share by the end of this or the next weekend. That will decouple the toolchain deps from both the cross versions and [differences between gcc versions](https://gist.github.com/tonytheodore/58e047ead7b76e4cfa97030767a7fa31).

orgads:
Ok. I'll abandon all my upgrades then.

PR #2581, opened 2020-11-04
mingw-w64: update to 8.0.0

orgads:


orgads:
Can you suggest a way to avoid the patch duplication? The same patch applies to all GCC versions except gcc4.

mabrand:
BTW, In the discussion at the link below, I asked about the build failure related to a change in Mingw-w64 8, which the GCC patches address. According to a later contributor, the failure needed to be fixed in GCC, not Mingw-w64. So this pull request is in line with that.

https://sourceforge.net/p/mingw-w64/bugs/853/

Seems good to me.

tonytheodore:
> Can you suggest a way to avoid the patch duplication?

You could do something like:
```
$(PKG)_PATCHES  := \
    $(dir $(lastword $(MAKEFILE_LIST)))/gccX.patch \
    $(dir $(lastword $(MAKEFILE_LIST)))/../gcc-common/dont-hard-code-printf.patch \
    $(dir $(lastword $(MAKEFILE_LIST)))/../gcc-common/dont-hard-code-mingw.patch \
    $(dir $(lastword $(MAKEFILE_LIST)))/../gcc-common/allow-native-cpu-detection-clang.patch
```
etc. but it's probably easier to keep them separate (and duplicated).

Note also, except gcc4.patch all of these have:
```
+#include <stdio.h>
+#include <stdio.h>
```

PR #2593, opened 2020-12-05
Improve python and mesa

tonytheodore:
Tested on darwin with all four targets and both python2/3.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,7 @@ SHELL      := bash
 
 MXE_TMP := $(PWD)
 
+ORIG_PATH  := $(call merge,:,$(filter-out $(PREFIX)/$(BUILD)/bin $(PREFIX)/bin,$(call split,:,$(PATH))))
 BUILD_CC   := $(shell (gcc --help >/dev/null 2>&1 && echo gcc) || (clang --help >/dev/null 2>&1 && echo clang))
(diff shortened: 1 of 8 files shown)


tonytheodore:
Also tested on Ubuntu 20.04 where:
```
$ `which python` --version
Python 2.7.18
```
and setuptools isn't installed by default.

I think it's ready for broader testing.

mabrand:
Nice work. Works for me on Ubuntu 21.04 (dev). I don't see any reason not to merge this.

tonytheodore:
For the other packages affected by python2, [`jack` has been updated](https://github.com/mxe/mxe/commit/beab89154df93ca39d723563c977f20ab8b6d8ce) and I'm nearly finished with `aubio` and `waf`.

jonaski:
This does not work on 64 bit linux systems where the libdir is lib64:
It tries to install to: usr/x86_64-pc-linux-gnu/lib64/python3.8/site-packages/ instead of usr/x86_64-pc-linux-gnu/lib/python3.8/site-packages.
I worked around it by adding --install-lib to setup.py. But that requires specifying the whole path.

`PYTHON_SETUP_INSTALL = cd '$(SOURCE_DIR)' && $(BUILD)-python$(PY_XY_VER) setup.py install --prefix='$(PREFIX)/$(TARGET)' --install-lib='$(PREFIX)/$(TARGET)/lib/python3.8/site-packages'`

PR #2598, opened 2020-12-23
sdl2: Bump to 2.0.14

lotharsm:
This patch bumps SDL2 to the recently released version 2.0.14. Additionally, I'm removing an now obsolete patch which got merged upstream and is no longer necessary here.


tonytheodore:
Thanks!

PR #2600, opened 2020-12-25
Add binutils 2.35.1 as a plugin

mspieth:
Add binutils 2.35.1 to help support usage of later gcc plugins


tonytheodore:
This should be added to the [gcc overlay](https://github.com/mxe/mxe/blob/master/plugins/gcc10/gcc10-overlay.mk) in the respective plugin (see `isl` as a guide).

I'll have proper support for multi-version gcc and deps soon.

PR #2601, opened 2020-12-27
URL_2 for expat

sadhen:
Failed to download expat from SourceForge. Add an alternative URL solves the problem.


tonytheodore:
Thanks!

PR #2604, opened 2020-12-29
Add a wxWidgets-3.1.4 plugin

janvdijk:
This makes a recent wxWidgets version available as a plugin.

This is related to pull request #2452 that proposes to update the default
wxWidgets version in mxe. Note the following differences wrt the code in
that pull request:

 * this plugin uses the usual MXE_CONFIGURE_OPTS, addressing the
   comments of @tonytheodore
 * configure complaints about unused options that result from this choice
   produce warnings, rather than errors, by passing --enable-option-checking
   to configure (without argument). This was suggested by @janvdijk (me)
   in the audit trail of pull request #2452
 * it makes version 3.1.4 available, rather than 3.1.3
 * no patches are needed to build this version of wxWidgets with the
   default gcc or with gcc9 or gcc10 (available as plugins in mxe).

This patch allows us to do in-the-wild testing of wxWidgets, before
making 3.1.4 the default, thereby obsoleting pull request #2452.
(I think that would be a good idea, even soon, given the facts that 3.0.x
does not see much maintenance (if at all), and that 3.1.x releases have
proven to be VERY stable and reliable.)

### DIFF
--- a/plugins/wx314/wx314-overlay.mk
+++ b/plugins/wx314/wx314-overlay.mk
@@ -1,0 +1,71 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := wxwidgets
+$(PKG)_WEBSITE  := https://www.wxwidgets.org/
+$(PKG)_DESCR    := wxWidgets
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 3.1.4
+$(PKG)_CHECKSUM := 3ca3a19a14b407d0cdda507a7930c2e84ae1c8e74f946e0144d2fa7d881f1a94
+$(PKG)_SUBDIR   := wxWidgets-$($(PKG)_VERSION)
+$(PKG)_FILE     := wxWidgets-$($(PKG)_VERSION).tar.bz2
+$(PKG)_URL      := https://github.com/wxWidgets/wxWidgets/releases/download/v$($(PKG)_VERSION)/$($(PKG)_FILE)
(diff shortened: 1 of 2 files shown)


tonytheodore:
Thanks for the update, it looks good and I'm building it now.

> I think that would be a good idea, even soon, given the facts that 3.0.x
does not see much maintenance (if at all), and that 3.1.x releases have
proven to be VERY stable and reliable.

I think this can be made the default immediately, a short-lived plugin may be confusing and plugins generally are better suited to either pinning old versions or testing alpha/beta type releases.

tonytheodore:
Build succeeds on all four targets.

> no patches are needed to build this version of wxWidgets

Only issue is that *.dlls are installed in `lib` instead of `bin` - this is currently patched but there may be some configure option for it.

tonytheodore:
> I think this can be made the default immediately

Indeed, the current version is failing on shared builds either due to mingw-w64 or libtiff updates.

janvdijk:
@tonytheodore, thanks for your feedback.

Since you have already done the polishing (thanks!), will you commit and push  this change? Otherwise, please let me know what I can do to promote this change.

janvdijk:
> Only issue is that *.dlls are installed in `lib` instead of `bin` - this is currently patched but there may be some configure option for it.

Just for the record: this is https://trac.wxwidgets.org/ticket/14601

tonytheodore:
Thanks, I see there's some progress on the ticket.

PR #2599, opened 2020-12-25
[WIP] Qt6 proof of concept

tonytheodore:

PR #2605, opened 2021-01-07
Add detection of --install to cmake wrapper.

jacobly0:
The `--install` command line option behaves similarly to `--build`, so treat it the same way.


tonytheodore:
Thanks! I think this tricked me a few weeks ago working on #2599

PR #2609, opened 2021-01-16
Quote asterisk in shell command for portability with zsh

vog:
The shell command

    apt-cache search mxe.*sdl_net

has two issues. First, it will lead to wrong results if a file like `mxe.FOOBARsdl_net` happens to exist in the current directory. Second, it will fail in `zsh` which is more strict about failed globs.

The proposed pull request fixes that.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -371,7 +371,7 @@ sudo apt-get update</pre>
     Note that all package and target names have underscores replaced with
     dashes. For example:
     </p>
-    <pre>apt-cache search mxe.*sdl_net</pre>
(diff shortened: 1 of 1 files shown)


tonytheodore:
Thanks for pointing this out and the deliberate use of single quotes - instructive as always.

I've been using zsh locally and was about to give up on it (due to similar such strictness e.g. `setopt rmstarsilent`), but now think I should persist and adopt better practices.

vog:
As a side note, I recommend to run shell scripts always through the excellent [shellcheck](https://www.shellcheck.net/) tool. (This is a command line tool, but also available as web service.)

PR #2628, opened 2021-03-24
Fix exiv2 updating

vog:
The Exiv2 updater didn't work anymore, which is why Exiv2 got stuck at version 0.25.

This pull request fixes the updater, and updates Exiv2 to its latest version 0.27.3.

### DIFF
--- a/src/exiv2.mk
+++ b/src/exiv2.mk
@@ -4,11 +4,11 @@ PKG             := exiv2
 $(PKG)_WEBSITE  := https://www.exiv2.org/
 $(PKG)_DESCR    := Exiv2
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 0.25
-$(PKG)_CHECKSUM := c80bfc778a15fdb06f71265db2c3d49d8493c382e516cb99b8c9f9cbde36efa4
-$(PKG)_SUBDIR   := exiv2-$($(PKG)_VERSION)
-$(PKG)_FILE     := exiv2-$($(PKG)_VERSION).tar.gz
-$(PKG)_URL      := https://www.exiv2.org/releases/$($(PKG)_FILE)
+$(PKG)_VERSION  := 0.27.3
+$(PKG)_CHECKSUM := a79f5613812aa21755d578a297874fb59a85101e793edc64ec2c6bd994e3e778
+$(PKG)_SUBDIR   := exiv2-$($(PKG)_VERSION)-Source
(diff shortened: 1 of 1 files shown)


src/exiv2.mk
```diff
@@ -4,34 +4,22 @@ PKG             := exiv2
...
-        --disable-visibility \
-        --disable-nls \
-        --with-expat \
-        LIBS='-lmman' \
-        $(if $(BUILD_SHARED),\
-            lt_cv_deplibs_check_method='file_magic file format (pe-i386|pe-x86-64)' \
-            lt_cv_file_magic_cmd='$$OBJDUMP -f')
-    $(MAKE) -C '$(1)/xmpsdk/src' -j '$(JOBS)'
-    $(MAKE) -C '$(1)/src'        -j '$(JOBS)' install-lib
+    cd '$(BUILD_DIR)' && $(TARGET)-cmake '$(SOURCE_DIR)' -DEXIV2_BUILD_SAMPLES=OFF
```
tonytheodore, line 22: Nice to see switching to cmake where possible, our version is recent enough to specify directories:
```
    $(TARGET)-cmake -S '$(SOURCE_DIR)' -B '$(BUILD_DIR)' \
        -DEXIV2_BUILD_EXIV2_COMMAND=OFF \
        -DEXIV2_BUILD_SAMPLES=OFF
```

vog:
Hi Tony,

Thanks for the hint. I improved my pull request accordingly.

However, I'm a bit uneasy with setting `-DEXIV2_BUILD_EXIV2_COMMAND=OFF`, so I didn't add that for now. The reason is that for Exiv2 it is important to have some linker test, especially for static linking, and the `exiv2` command line tool spares us from having to create our own dummy test program. Moreover, the `exiv2.exe` command line tool is very handy, even when run only through Wine, to debug issues related to Exiv2 (and/or its dependency libraries) if something works on Linux but not on Windows.

vog:
Do you agree, or should I really add `-DEXIV2_BUILD_EXIV2_COMMAND=OFF` to my commit?

tonytheodore:
Hi Volker,

> important to have some linker test, especially for static linking, and the exiv2 command line tool spares us from having to create our own dummy test program

Dummy programs can also check external configuration the package provides - that can be a losing battle though and this is a sufficient proxy for our purposes. 

> the exiv2.exe command line tool is very handy

100% agree with including handy tools.

Thanks!

PR #2632, opened 2021-04-13
Fix gettext builds (gettext bug 59929)

janvdijk:
This patch fixes a gettext configury bug that prevents building MXE on my OpenSUSE Tumbleweed system with autoconf 2.69 and automake 1.16.3, producing an error about obsolete invocations of AC_INIT, AM_INIT_AUTOMAKE  It has been taken from the gettext bugtracker, see https://savannah.gnu.org/bugs/?59929 for details.

### DIFF
--- a/src/gettext.mk
+++ b/src/gettext.mk
@@ -12,6 +12,7 @@ $(PKG)_URL_2    := https://ftpmirror.gnu.org/gettext/$($(PKG)_FILE)
 # native gettext isn't technically required, but downstream
 # cross-packages may need binaries and/or *.m4 files etc.
 $(PKG)_DEPS     := cc libiconv $(BUILD)~$(PKG)
+$(PKG)_PATCHES  := $(TOP_DIR)/src/gettext-1-fixes.patch
 
 $(PKG)_TARGETS       := $(BUILD) $(MXE_TARGETS)
 $(PKG)_DEPS_$(BUILD) := libiconv
--- a/src/gettext-1-fixes.patch
+++ b/src/gettext-1-fixes.patch
@@ -1,0 +1,19 @@
+Fix https://savannah.gnu.org/bugs/?59929
+'AC_INIT with proper package and version'
+
+--- gettext-0.21/libtextstyle/configure.ac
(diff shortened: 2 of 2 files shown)


jonaski:
Thanks for the patch, I can confirm it works.
What's preventing this from getting merged?

Error:

```
configure.ac:25: installing '../build-aux/missing'
configure.ac:25: error: AC_INIT should be called with package and version arguments
: gl_AM_INIT_AUTOMAKE is expanded from...
/usr/share/aclocal-1.16/init.m4:29: AM_INIT_AUTOMAKE is expanded from...
configure.ac:25: the top level
autom4te: /usr/bin/m4 failed with exit status: 1
aclocal: error: /usr/bin/autom4te failed with exit status: 1
autoreconf: aclocal failed with exit status: 1
```

jonaski (approved):

src/gettext.mk
```diff
@@ -12,6 +12,7 @@ $(PKG)_URL_2    := https://ftpmirror.gnu.org/gettext/$($(PKG)_FILE)
 # native gettext isn't technically required, but downstream
 # cross-packages may need binaries and/or *.m4 files etc.
 $(PKG)_DEPS     := cc libiconv $(BUILD)~$(PKG)
+$(PKG)_PATCHES  := $(TOP_DIR)/src/gettext-1-fixes.patch
```
tonytheodore, line 15: This isn't necessary, it's only required if the usual logic of `pkg-[0-9]*.patch` isn't suitable.

tonytheodore:
I'm not certain we need the `autoreconf` step at all. That most likely should have been removed when any patch touching the configury was removed.

janvdijk:
@tonytheodore  thank you for the review, and for educating me about `$(PKG)_PATCHES`.

I can confirm your and @jonaski's comment that everything works fine (without any gettext patches) if the two lines containing `autoreconf` are removed from `src/gettext.mk`. If that is a better resolution, would you be willing to make that change and close the bug and pull request?

tonytheodore:
Done! Darwin still needs the autoconf step so that has been moved to the pinned version.

PR #2637, opened 2021-04-16
update cmake

jonaski:


tonytheodore:
I usually check all cmake-based packages with:
```
make `make show-downstream-deps-cmake` MXE_TARGETS="i686-w64-mingw32.static i686-w64-mingw32.shared x86_64-w64-mingw32.static x86_64-w64-mingw32.shared"
```
I'll try to start a build after catching up on some other changes.

tonytheodore:
Only one issue https://github.com/mxe/mxe/commit/03d3fcff6477145350f5737374ee23b6f889edda

PR #2414, opened 2019-10-11
Update qtwebkit

tehnick:

PR #2642, opened 2021-05-07
update qtwebkit

tehnick:
Tested for:
```
MXE_TARGETS := x86_64-w64-mingw32.shared i686-w64-mingw32.shared
MXE_PLUGIN_DIRS += plugins/gcc10
```

### DIFF
--- a/src/qtwebkit.mk
+++ b/src/qtwebkit.mk
@@ -4,10 +4,10 @@ PKG             := qtwebkit
 $(PKG)_WEBSITE  := https://github.com/annulen/webkit
 $(PKG)_DESCR    := QtWebKit
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 5.13.1
-$(PKG)_CHECKSUM := f688e039e2bdc06e2e46680f3ef57715e1b7d6ea69fd76899107605a8f371ea3
-$(PKG)_SUBDIR   := qtwebkit-everywhere-src-$($(PKG)_VERSION)
-$(PKG)_FILE     := qtwebkit-everywhere-src-$($(PKG)_VERSION).tar.xz
+$(PKG)_VERSION  := 5.212
+$(PKG)_CHECKSUM := 10cbdaba60aac79d27016aa05bae9ab3ec7b0aed4df163debfbf8fddd66adc14
+$(PKG)_SUBDIR   := qtwebkit-opensource-src-$($(PKG)_VERSION)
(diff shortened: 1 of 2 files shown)


src/qtwebkit.mk
```diff
@@ -4,10 +4,10 @@ PKG             := qtwebkit
 $(PKG)_WEBSITE  := https://github.com/annulen/webkit
 $(PKG)_DESCR    := QtWebKit
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 5.13.1
-$(PKG)_CHECKSUM := f688e039e2bdc06e2e46680f3ef57715e1b7d6ea69fd76899107605a8f371ea3
-$(PKG)_SUBDIR   := qtwebkit-everywhere-src-$($(PKG)_VERSION)
-$(PKG)_FILE     := qtwebkit-everywhere-src-$($(PKG)_VERSION).tar.xz
+$(PKG)_VERSION  := 5.212
```
tonytheodore, line 7: This looks like going back to the old version in #2414?

tehnick: qtwebkit-opensource-src-5.212.tar.xz != qtwebkit-everywhere-src-5.13.1.tar.xz

tehnick: And yes, I don't understand why they did this...

tehnick:
PR is ready for merging. I have tested this qtwebkit in builds for Psi+ project.

tehnick:
@annulen  
Could you comment strange situation with official tarballs for qtwebkit which is described in https://github.com/mxe/mxe/pull/2642#issuecomment-834417113 ?

For now I just use latest tarball (based on timestamps).

tonytheodore:
Since it's cmake-based, can we use a tagged repo download instead of a tarball?

tehnick:
@tonytheodore 
What repo do you have in mind?
1) https://github.com/qtwebkit/qtwebkit
2) https://code.qt.io/cgit/qt/qtwebkit.git/
3) https://github.com/qt/qtwebkit

I never checked them in a whole and not compared sources from git tags with ones from tarballs above. 
(As these repos are too huge, I cloned them only with `--depth 1` option of git.)

In any case qtwebkit with changes proposed in this PR works well. At least it was tested by Psi+ users.

tonytheodore:
@tehnick, the tags in those repos don't shed any light on versioning, and [one warns about using the prepared tarballs and "not" the auto-generated repo archives](https://github.com/qtwebkit/qtwebkit/releases/tag/qtwebkit-5.212.0-alpha4).

Looks good to me - thanks!

PR #2643, opened 2021-05-07
qtbase: add missing limits includes

jonaski:
This should fix the compile errors with qtbase when using GCC 11.

I also have a patch for Qt 6 base here:
https://github.com/strawberrymusicplayer/strawberry-mxe/blob/master/src/qt6base-3-fixes.patch

### DIFF
--- a/src/qtbase-2-fixes.patch
+++ b/src/qtbase-2-fixes.patch
@@ -1,0 +1,60 @@
+This file is part of MXE. See LICENSE.md for licensing information.
+
+Contains ad hoc patches for cross building.
+
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Jonas Kvinge <jonas@jkvinge.net>
+Date: Wed, 5 May 2021 14:57:25 +0200
+Subject: [PATCH 1/1] Add missing limits include
+
+
+diff --git a/src/corelib/global/qendian.h b/src/corelib/global/qendian.h
+index 1111111..2222222 100644
+--- a/src/corelib/global/qendian.h
++++ b/src/corelib/global/qendian.h
+@@ -44,6 +44,8 @@
+ #include <QtCore/qfloat16.h>
+ #include <QtCore/qglobal.h>
+ 
++#include <limits>
++
+ // include stdlib.h and hope that it defines __GLIBC__ for glibc-based systems
+ #include <stdlib.h>
+ #include <string.h>
+diff --git a/src/corelib/global/qfloat16.h b/src/corelib/global/qfloat16.h
+index 1111111..2222222 100644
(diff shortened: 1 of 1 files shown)


mabrand:
d0ddb14a9a528d1362a127b00f445d72c4edf9cb fixed these, except for src/corelib/tools/qoffsetstringarray_p.h. Does this one actually fail to build in your configuration?

Jan200101:
qoffsetstringarray_p.h fails to build for me

```
qoffsetstringarray_p.h:70:22: error: ‘numeric_limits’ is not a member of ‘std’
   70 |         Last <= std::numeric_limits<quint8>::max(),
```

mabrand:
> qoffsetstringarray_p.h fails to build for me
> 
> ```
> qoffsetstringarray_p.h:70:22: error: ‘numeric_limits’ is not a member of ‘std’
>    70 |         Last <= std::numeric_limits<quint8>::max(),
> ```

Did that happen during the build of qtbase (qt5)? Do you have any changes of your own in qtbase.mk? I'm just trying to understand why I don't get this same error.

Jan200101:
> Did that happen during the build of qtbase (qt5)?

Yes

> Do you have any changes of your own in qtbase.mk? 

I tried with 2 different times: once with an existing package set and once a clean build with just qt5
 
>  I'm just trying to understand why I don't get this same error.

Here are the contents of my settings.mk in case they are can help replicate the behavior
```
JOBS := 6

MXE_TARGETS := i686-w64-mingw32.static

override MXE_PLUGIN_DIRS += plugins/gcc10

LOCAL_PKG_LIST := cc qt5
.DEFAULT_GOAL  := local-pkg-list
local-pkg-list: $(LOCAL_PKG_LIST)
```

I attached my build logs incase they might point out the problem
[qtbase_i686-w64-mingw32.static.txt](https://github.com/mxe/mxe/files/6447494/qtbase_i686-w64-mingw32.static.txt)

Jan200101:
can confirm that with this PR qtbase compiles just fine

mabrand:
@Jan200101 :
> 
> > I'm just trying to understand why I don't get this same error.
> 
> Here are the contents of my settings.mk in case they are can help replicate the behavior
> 
> ```
> JOBS := 6
> 
> MXE_TARGETS := i686-w64-mingw32.static
> 
> override MXE_PLUGIN_DIRS += plugins/gcc10
> 

Maybe GCC 10 (instead of 11) makes the difference, or i686 instead of x64, but the fix for tqoffsetstringarray_p.h is correct anyway, so I included it.

@jonaski :

Thanks for pointing out the subsequent partial revert, which I've now included. I like to use the upstream commits where possible.

Does this look good now?

jonaski:
OK

PR #2648, opened 2021-05-09
boost: GCC 11 build fix

mabrand:
GCC 11 complains about -U__STRICT_ANSI__.

    usr/lib/gcc/x86_64-w64-mingw32.static/11.1.0/include/c++/\
        x86_64-w64-mingw32.static/bits/c++config.h:558:2: \
        error: #warning "__STRICT_ANSI__ seems to have been \
        undefined; this is not supported" [-Werror=cpp]
    558 | #warning "__STRICT_ANSI__ seems to have been undefined; this is not supported"

I don't know if -U__STRICT_ANSI__ is (still) useful here for other compilers.


tonytheodore:
Works on gcc5 and gcc9 - thanks!

PR #2655, opened 2021-05-21
sdl2: Use SDL2's CMake-based build system

mborgerson:
This patch migrates SDL2 from the autotools build system to the CMake build system. This appears to be the preferred route now, as both [libsdl project CI](https://github.com/libsdl-org/SDL/blob/main/.github/workflows/main.yml) and [MSYS2](https://packages.msys2.org/package/mingw-w64-x86_64-SDL2) both use the CMake path. Three patches are added to fixup minor build issues, two of which originate from the MSYS2 project.


lephilousophe:
@tonytheodore @mabrand I don't know what happened, but it seems the commit in this PR got its way in the [build-2022-04-09 tag](https://github.com/mxe/mxe/releases/tag/build-2022-04-09) while this PR is not merged.
I wonder which action was intended, but it's now in a kind of quantum state.

lephilousophe:
In addition, this PR breaks SDL2 build in the tag because the patches are outdated

tonytheodore:
I must have created that tag from a testing branch that included this PR, but the patches fail as you mentioned.

Are there any downsides to deleting the tag?

lephilousophe:
I don't know, maybe some people using this tag will find it vanished?
A unclean way would be to recreate it with the same name but without this commit. Maybe it's the safest?

tonytheodore:
> Are there any downsides to deleting the tag?

https://git-scm.com/docs/git-tag#_on_re_tagging

[I'll leave it as-is](https://lists.nongnu.org/archive/html/mingw-cross-env-list/2022-04/msg00000.html) and go back to annotated tags that have some notes about how complete the build is:
  - https://github.com/mxe/mxe/releases/tag/build-2021-04-22
  - https://github.com/mxe/mxe/releases/tag/build-2021-01-10

There are many broken packages and this tag would have been:
```
echo "MXE_USE_CCACHE := no" >> settings.mk && \
echo "EXCLUDE_PKGS = ocaml%" >> settings.mk && \
[...]
MXE_BUILD_PKG_TARGETS="`echo {i686-w64-mingw32,x86_64-w64-mingw32}.{static,shared}.gcc11`" \
MXE_BUILD_PKG_PKGS='qt5 qt6 gtkmm3 wxwidgets'
```

tonytheodore:
@mborgerson, can you rebase this PR against the updated SDL2 versions from https://github.com/mxe/mxe/pull/2801

mborgerson:
> @mborgerson, can you rebase this PR against the updated SDL2 versions from #2801

A bit surprised to see activity on this now. Thanks for reviewing.

I've pushed an updated patch to resolve the conflicts from #2801 being merged first, and updated the patches to work with this release of SDL.

The test case builds and this works for my application, but I have not done further testing on it.

By the way, 2.0.22 is coming soon and this patch works for it but will need a minor update, so I can PR an update when it finally releases.

mborgerson:
Rebased to 2.26.2, bumping for review.

PR #2658, opened 2021-05-23
update mingw-w64

jeroen:
https://sourceforge.net/p/mingw-w64/mailman/message/37287751/


jonaski:
Shouldn't the default Windows version in mingw-w64 be changed to Windows 7, or the _WIN32_WINNT defines in the respective .mk files bumped?

PR #2654, opened 2021-05-17
Update packages

fiesh:
Fix gcc11 compilation issues and update and de-bloat `OpenSceneGraph` along the way.

PR #2667, opened 2021-05-31
qtdeclarative: fix compilation with gcc 11

mbunkus:
A couple more instances of gcc 11 needing the `<limits>` header included explicitly.


fiesh:
Note that this is also already included in #2654

mbunkus:
Interesting. That PR didn't show up when I was searching for `qtdeclarative`, hence my PR. I'll those mine in favor of yours.

PR #2670, opened 2021-06-13
Fix qtdeclarative for gcc11: missing include.

janvdijk:
Include limits.h for std::numeric_limits. This fixes a compilation error with the gcc11 plugin. The problem has been observed before, and the patch has been taken from: `https://build.opensuse.org/package/view_file/openSUSE:Factory/libqt5-qtdeclarative/0001-Include-limits-to-fix-build-with-GCC-11.patch`


mbunkus:
Just FYI, I offered #2667 for this a couple of weeks ago, too, but was told that an equivalent fix was already part of another PR #2654 that was supposed to be merged soon(ish).

janvdijk:
@mbunkus, thanks for your message and sorry for the dup. I failed to see your (closed) report, was looking for open ones myself. Let's hope this gets fixed one way or the other soon.

tonytheodore:
#2654 is merged now.

PR #2680, opened 2021-07-05
Add fix to prevent Qt crash with SVG input

fiesh:
We add a patch for

https://bugreports.qt.io/browse/QTBUG-94692

that was taken from the there.

### DIFF
--- a/src/qtbase-1-fixes.patch
+++ b/src/qtbase-1-fixes.patch
@@ -221,3 +221,240 @@ index 4dd9e9603b..c94d64d9aa 100644
 -- 
 2.30.2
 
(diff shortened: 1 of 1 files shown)


fiesh:
@mabrand it actually might make more sense to integrate it as a separate patch file.  That was, it's clear that it corresponds to a single diff from Qt.  What do you think?  We have another back ported bug fix that we include by now, so I could open a second PR for that.

mabrand:
My preference is to add it to qtbase-1-fixes.patch. You could add both patches to this file, for one MXE pull request.

fiesh:
@mabrand Closing this in favor of #2751 where both patches were added.

PR #2688, opened 2021-07-31
libjpeg-turbo: Install files in the correct directories

orgads:
Installing in */libjpeg-turbo is very strange. The package cannot be
built without crafted -I and -L, and even pkgconfig is not installed
with all the other packages.

### DIFF
--- a/src/libjpeg-turbo.mk
+++ b/src/libjpeg-turbo.mk
@@ -21,9 +21,7 @@ define $(PKG)_BUILD
     cd '$(BUILD_DIR)' && $(TARGET)-cmake '$(SOURCE_DIR)' \
         -DENABLE_SHARED=$(CMAKE_SHARED_BOOL) \
         -DENABLE_STATIC=$(CMAKE_STATIC_BOOL) \
-        -DCMAKE_INSTALL_BINDIR='$(PREFIX)/$(TARGET)/bin/$(PKG)' \
-        -DCMAKE_INSTALL_INCLUDEDIR='$(PREFIX)/$(TARGET)/include/$(PKG)' \
-        -DCMAKE_INSTALL_LIBDIR='$(PREFIX)/$(TARGET)/lib/$(PKG)' \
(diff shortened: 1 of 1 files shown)


orgads:
ping

orgads:
ping?

mabrand:
We don't want libjpeg-turbo to clobber files installed by regular jpeg.

orgads:
Then maybe at least install the pkgconfig pc file as libjpeg-turbo.pc, but in the right directory?

mabrand:
libjpeg-turbo installs both libjpeg.pc and libturbojpeg.pc. How would you deal with this?

orgads:
I can think of 2 options:
1. Do not allow installing both packages (is there a "conflicting package" feature on mxe?)
2. Install libturbojpeg.pc in the default pkgconfig directory, and optionally install libjpeg-turbo/libjpeg.pc.

mabrand:
As far as I know, there isn't any general "conflicting package" feature in MXE. It would probably get complicated on account of other packages depending on one or the other alternative. The general approach has always been to configure packages in a such a way that all the packages can be built without clobbering each other. Otherwise, we'd be faced with the problem of testing builds for all the possible option combination. This at least provides a known baseline for users who need to modify package configurations for specific needs, such as completely replacing jpeg by libturbojpeg.

Your 2nd option create a default situation where the standard lib directory contains files from both jpeg packages. It wouldn't be immediately obvious  which package provided libjpeg.pc and libjpeg.a.

orgads:
So if you want to support both, then you should *not* deploy `libjpeg.pc` for `libjpeg-turbo`, and leave only `libturbojpeg.pc`.

orgads:
I compared them now, and I see that the libraries are different. `libjpeg` provided by Turbo has an API that is compatible to the official `libjpeg`, while `libturbojpeg` extends it with more functions. From what it looks, it should work though.

mabrand:
Have you considered using a plugin to make libturbojpeg the default jpeg library for your build? https://github.com/mxe/mxe/blob/master/plugins/README.md

PR #2690, opened 2021-08-01
curl: build with schannel support

lotharsm:
Currently, the curl package specifies `--with-winssl` at configure time. However, in current versions, this got renamed to `--with-schannel` [1]

The old `--with-winssl` parameter is no longing working, which unfortunately currently breaks SSL/TLS support entirely.

[1] https://github.com/curl/curl/blob/master/configure.ac#L146


lotharsm:
Is there anything left to do before this can be merged? :-)

PR #2694, opened 2021-08-04
Fix the host GCC to find the driver on MinGW builds

kalibera:
This trivial GCC 10 patch fixes the host toolchain to find the driver (cc1, ...). This is originally from Martin Storsjo, added to GCC 11 (https://gcc.gnu.org/pipermail/gcc-patches/2021-April/567815.html), but not GCC 10.

More details: GCC 10 driver is compiled with g++ without-D__USE_MINGW_ACCESS on MinGW. That option is only added to CFLAGS, but not CXXFLAGS. Consequently, the driver will not find executables such as cc1, because access(,X_OK) will always return an error on Windows, as access() on Windows does not check for executability anymore. With __USE_MINGW_ACCESS, MinGW will use its own implementation of access() which for X_OK behaves the same as R_OK.

Reported to GCC via https://gcc.gnu.org/bugzilla/show_bug.cgi?id=101238 and https://gcc.gnu.org/pipermail/gcc-patches/2021-July/574404.html.

### DIFF
--- a/plugins/gcc10/gcc10-1-driver-access-fix.diff
+++ b/plugins/gcc10/gcc10-1-driver-access-fix.diff
@@ -1,0 +1,13 @@
+diff -Nru gcc-10.2.0-orig/config/mh-mingw gcc-10.2.0-patched/config/mh-mingw
+--- gcc-10.2.0-orig/config/mh-mingw	2020-07-23 02:35:16.916379838 -0400
++++ gcc-10.2.0-patched/config/mh-mingw	2021-07-04 03:36:49.987025066 -0400
+@@ -1,7 +1,9 @@
+ # Add -D__USE_MINGW_ACCESS to enable the built compiler to work on Windows
+ # Vista (see PR33281 for details).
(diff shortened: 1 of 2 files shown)


kalibera:
Thanks. In the meantime this has been ported also to GCC, but it will take some time before it gets released.

mabrand:
Did I understand correctly that gcc 11 doesn't need this patch>

kalibera:
Yes, gcc 11.2.0 already has the patch.
However, gcc 9.4.0 still does not, it should be hence patched by MXE as well.

There are actually more patches that are needed, I just had submitted this one first. I am using MXE to build a cross and native toolchain and libraries for use with R, now using gcc 10.3, see here https://svn.r-project.org/R-dev-web/trunk/WindowsBuilds/winutf8/ucrt3/toolchain_libs/mxe/plugins/gcc10. Particularly gcc10-5-setjmp-longjmp-seh.diff and gcc10-6-reorder-unwind-info.diff are important broadly (fix crashes of compiled code on Windows) with gcc 10.3. They are also already on the gcc 10 branch, but have been added after 10.3.0 has been released.

PR #2699, opened 2021-08-09
sdl2_net: Add required libraries to pkgconfig on static build

orgads:
See https://github.com/libsdl-org/SDL_net/pull/35

### DIFF
--- a/src/sdl2_net-2-pkg-config.patch
+++ b/src/sdl2_net-2-pkg-config.patch
@@ -1,0 +1,80 @@
+This file is part of MXE. See LICENSE.md for licensing information.
+
+From e2782ccaa642202e50aebefe4107e4313f6bbb53 Mon Sep 17 00:00:00 2001
+From: Cameron Cawley <ccawley2011@gmail.com>
+Date: Sat, 17 Jul 2021 15:56:53 +0100
+Subject: [PATCH] Include required libraries in pkg-config output
+
+---
+Taken from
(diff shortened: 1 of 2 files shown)


mabrand:
I think it would be better to do this with a patch instead of a "sed hack".

orgads:
Done. Source: https://github.com/scummvm/dockerized-bb/blob/master/toolchains/mxe/packages/sdl2_net/sdl2_net-1-pkg-config.patch

mabrand:
Sorry about nitpicking, but could you please mention the origin of the patch in the commit message of the patch in the .patch file? Such as "Taken from <URL>".

orgads:
No problem at all. Done.

src/sdl2_net-1-pkg-config.patch
```diff
@@ -0,0 +1,80 @@
+From e2782ccaa642202e50aebefe4107e4313f6bbb53 Mon Sep 17 00:00:00 2001
+From: Cameron Cawley <ccawley2011@gmail.com>
+Date: Sat, 17 Jul 2021 15:56:53 +0100
+Subject: [PATCH] Include required libraries in pkg-config output
+
+---
```
mabrand, line 6: Please include "Taken from.." here.
Also, the stats aren't necessary. The script tools/patch-tool-mxe can be used to "normalize" patch format.

orgads: Done.

PR #2707, opened 2021-09-13
Added db48.mk - Oracle Berkeley DB v4.8.30

SharkyRawr:
Made a copy of db.mk and adjusted the version number. Building works just as good as the original db.mk. :)

BDB 4.8.30 is still being used by older Altcoin wallets and it would be great to see in MXE.

### DIFF
--- a/src/db48.mk
+++ b/src/db48.mk
@@ -1,0 +1,31 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := db48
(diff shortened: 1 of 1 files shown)


0T34:
I recommend merge this PR, this version of BDB is very used to build a lot of cryptocurrency projects (including Bitcoin).

tonytheodore:
Does this install side-by-side with different lib names etc.?

PR #2715, opened 2021-10-01
meson-wrapper: add sys_root and pkg_config_libdir

v1993:
Those are available in newer versions of meson.


v1993:
Note: this is superseded by upcoming GTK update PR, which moves this from plugin into src.

PR #2721, opened 2021-10-12
gdal 2.2.4: Support for openjpeg 2.4

HuidaeCho:
This PR updates `src/gdal-1-fixes.patch` so that it supports openjpeg 2.4. `configure.ac` and `frmts/openjpeg/openjpegdataset.cpp` will be patched.


mabrand:
Should be fixed by eb7df8f39184da38c47c69e9345f7bf091e165c1 and c2cdd3a06bf9b0e596974d70b241c06889ba16f8.

PR #2730, opened 2021-11-20
Bump ncurses to latest master.

toots:
Current `ncurses` version fails to build with recent compilers. This one works!


h0tw1r3:
This is a major upgrade to ncurses. 5.9 to 6.3. Likely breaks dependencies. See #2720 for 5.9 patch.

PR #2751, opened 2022-01-19
Add two qt patches

fiesh:
This supersedes #2680.

Add two bug fix patches for Qt / Qt Declarative.

### DIFF
--- a/src/qtbase-1-fixes.patch
+++ b/src/qtbase-1-fixes.patch
@@ -221,6 +221,7 @@ index 4dd9e9603b..c94d64d9aa 100644
 -- 
 2.30.2
 
+
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Tony Theodore <tonyt@logyst.com>
 Date: Thu, 11 Nov 2021 18:18:25 +1100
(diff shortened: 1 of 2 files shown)


src/qtbase-1-fixes.patch
```diff
@@ -221,6 +221,7 @@ index 4dd9e9603b..c94d64d9aa 100644
 -- 
 2.30.2
 
+
```
mabrand, line 224: stray whitespace

fiesh: No it's not. All patches area separated by double newlines, and this one was missing that, so I added it.  But I can remove that fix if it matters.

mabrand:
I had to revert this because it does not build. The patch does not apply. Please test the build, both static and shared.

fiesh:
> I had to revert this because it does not build. The patch does not apply. Please test the build, both static and shared.

I see, sorry about that.  I had tried it when I prepared the original commit.  I had actually started a new MXE build but that hadn't finished yet.

fiesh:
@mabrand So the patch just seems to have made it upstream in the meantime.  The second one was still meaningful, and reverting the first one is thus the correct solution.  Sorry for the broken pull request!

mabrand:
Can we wait for the 2nd patch to be merged upstream? I wouldn't mind cherry-picking it from another upstream branch, but I'm a bit reluctant to to add it here in MXE before it's been accepted upstream.

fiesh:
> Can we wait for the 2nd patch to be merged upstream? I wouldn't mind cherry-picking it from another upstream branch, but I'm a bit reluctant to to add it here in MXE before it's been accepted upstream.

From what I can tell, there are no plans to merge it into Qt5 any more, which seems entirely political to me as there's really no reason not to.

If you don't want to include it on a _let's stick with stock Qt5 as much as possible_ principle, I can understand that.  From a merely technical point of view, adding it seems preferable to me.

mabrand:
Has it been merged into a Qt 6 branch?

jonaski:
If you are talking about the patch replacing C setjmg/longjmp with C++ exceptions it has not even been submitted to code review because Qt does not use exceptions, so I'm not sure it will be accepted, so it is not in Qt 6 either. In my MXE repos I use the patch for Qt 6 and have been using it for over 1/2 year for Strawberry without any issues. The bug does not only affect SVG, it affects regular painting too, see issues https://bugreports.qt.io/browse/QTBUG-94692 and https://bugreports.qt.io/browse/QTBUG-93476
Qt 5 is end of life as far as open source is concerned so we should upgrade to Qt 6, and Qt 5 should probably be moved to plugins like Qt 4.

fiesh:
> Has it been merged into a Qt 6 branch?

Yes, see the https://bugreports.qt.io/browse/QTBUG-30801

fiesh:
> If you are talking about the patch replacing C setjmg/longjmp...

No we were talking (or *I* was talking ;-)) about the patch that fixes hover states in `qtdeclarative` and never made it to Qt5, only to Qt6.

mabrand:
Thanks for the explanation. I don't think the qtbase patches belong in MXE. In general, I don't think we should get ahead of upstream without a very good reason, such as fixing build problems in MXE, especially if the patch hasn't been reviewed and approved upstream.

mabrand:
We do want to add qt6. The packages will be named so they don't clash with qt5 packages. The obstacle to qt6 in MXE now, as far as I know is the switch to cmake,  headaches with static linking transitive dependencies being one of the issue.

PR #2754, opened 2022-01-21
Update meson

mabrand:


danger89:
Do NOT merge. We fixed Meson builds already here: https://github.com/danger89/mxe (I will create a PR soon)

v1993:
To be exact - meson-wrapper already was part of MXE. We've only moved it to main tree and extended.

Issue #2756, opened 2022-01-26
Upgrade libgphoto2 from 2.5.27 to 2.5.28 


cgilles:
Hi all,

digiKam use now under Windows the libgphoto2 shared library to support Gphoto2 camera devices.

Currently, the libgphoto2 in MXE is the 2.5.27.

Under MacOS, digiKam using 2.5.27 do not list the camera drivers. Just upgrading libgphoto2 to 2.5.28 fix the problem:

digiKam under MacOS with libgphoto2 2.5.27 : https://i.imgur.com/JG5hV7g.png
digiKam under MacOS with libgphoto2 2.5.28 : https://i.imgur.com/elfzIcG.png

Under Windows using libgphoto2 2.5.27, we have exactly the same problem than under MacOS. The list of camera driver is empty:

https://i.imgur.com/vVLZFDf.png

So i suspect a bug fixed in libgphoto2. 

Please, can you upgrade libgphoto2 to last stable 2.5.28:

https://github.com/gphoto/libgphoto2/releases/tag/v2.5.28

Thanks in advance

Gilles Caulier

mabrand:
A tested pull request would be welcome!

cgilles:
Just for info, i also tested the libgphoto2 commit in MXE for the libusb detection. I

https://github.com/mxe/mxe/commit/5e2f623d08b02cc4604a68dfa75df4b0d79b789b

upgraded MXE, recompiled the Windows installer, but the result still the same : list of Gphoto2 camera drivers still empty.

Gilles Caulier

yarda:
I created PR for the rebase  to 2.5.28. I haven't tested whether it resolves the original problem.

PR #2757, opened 2022-01-27
libgphoto2: updated to 2.5.28

yarda:
Also fixed whitespaces in the mk file.

Fixes: #2756

Signed-off-by: Jaroslav Škarvada <jskarvad@redhat.com>

  * install .pc file,

Thanks!


yarda:
I dropped the patches, because:
- patch1 was upstreamed,
- patch2 is not needed, because it's now controlled through the enviromental variables during the build

yarda:
The CI failure doesn't seem to be related to this PR.

PR #2766, opened 2022-02-17
add patch to get ilmbase building under gcc11

tomager:
This fixes the ilmbase package-- recent build with gcc11 made default gives errors such as:

```
ImathVec.h:795:35: error: ISO C++17 does not allow dynamic exception specifications
  795 | Vec3<int>::normalizedExc () const throw (IEX_NAMESPACE::MathExc);
```

This applies patch that removes the throw syntax that is no longer valid past C++17.

### DIFF
--- a/src/ilmbase-1-fixes.patch
+++ b/src/ilmbase-1-fixes.patch
@@ -1,0 +1,439 @@
+This file is part of MXE. See LICENSE.md for licensing information.
+
+diff -ru a/Imath/ImathMatrix.h b/Imath/ImathMatrix.h
+--- a/Imath/ImathMatrix.h	2014-08-10 02:03:40.000000000 +0000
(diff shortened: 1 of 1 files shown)


tomager:
I agree regarding ```noexcept``` being the proper replacement for ```throw``` without type list, however I don't see that there is a replacement in later C++ for ```throw(type-id-list)```, which is what is used here. I think the most correct change is to remove the noexcept/throw specification altogether, since in fact the older code says that is _can_ throw (of the type given in the throw expression).

AFAIK, noexcept/throw maybe impacts downstream ```constexpr``` usage (in much newer code) and whether std::unexpected is called on a throw (i.e., catastrophic program failure). So for older code it is fine to remove them, and for newer code they should not be marked as ```noexcept``` anyway.

Issue #2691, opened 2021-08-01
gdal compile error under ubuntu 20.4


dbyoung720:
Failed to build package gdal for target i686-w64-mingw32.static!
------------------------------------------------------------
 libs="-L/home/dbyoung/Desktop/mxe/usr/i686-w64-mingw32.static/lib -lnetcdf -lmfhdf -ldf -lhdf5_hl -lhdf5 -lpthread -lz -lm -lcurl", includedir="/home/dbyoung/Desktop/mxe/usr/i686-w64-mingw32.static/include"
checking for nc_open in -lnetcdf... no
configure: JasPer (JPEG2000) support disabled.
configure: error: openjpeg.h not found in /home/dbyoung/Desktop/mxe/usr/i686-w64-mingw32.static/include/openjpeg-2.0 or /home/dbyoung/Desktop/mxe/usr/i686-w64-mingw32.static/include/openjpeg-2.1 or /home/dbyoung/Desktop/mxe/usr/i686-w64-mingw32.static/include/openjpeg-2.2 or /home/dbyoung/Desktop/mxe/usr/i686-w64-mingw32.static/include/openjpeg-2.3
make[1]: *** [Makefile:874: build-only-gdal_i686-w64-mingw32.static] Error 1
make[1]: Leaving directory '/home/dbyoung/Desktop/mxe'

mabrand:
Should be fixed by eb7df8f39184da38c47c69e9345f7bf091e165c1 and c2cdd3a06bf9b0e596974d70b241c06889ba16f8.

Issue #2610, opened 2021-02-02
Failed to build package gdal


coonce:
It looks like gdal is looking for openjpeg 2.0, 2.1, 2.2, or 2.3, but mxe is downloading 2.4: 
https://github.com/mxe/mxe/blob/f879e4ea2303e0db5c26ab3a8d3267ed350c4e0f/src/openjpeg.mk#L8
```
Failed to build package gdal for target x86_64-w64-mingw32.static!
------------------------------------------------------------
 libs="-L/home/daniel/dvlp/mxe/usr/x86_64-w64-mingw32.static/lib -lnetcdf -lmfhdf -ldf -lhdf5_hl -lhdf5 -lpthread -lz -lm -lcurl", includedir="/home/daniel/dvlp/mxe/usr/x86_64-w64-mingw32.static/include"
checking for nc_open in -lnetcdf... no
configure: JasPer (JPEG2000) support disabled.
configure: error: openjpeg.h not found in /home/daniel/dvlp/mxe/usr/x86_64-w64-mingw32.static/include/openjpeg-2.0 or /home/daniel/dvlp/mxe/usr/x86_64-w64-mingw32.static/include/openjpeg-2.1 or /home/daniel/dvlp/mxe/usr/x86_64-w64-mingw32.static/include/openjpeg-2.2 or /home/daniel/dvlp/mxe/usr/x86_64-w64-mingw32.static/include/openjpeg-2.3
make[1]: *** [Makefile:870: build-only-gdal_x86_64-w64-mingw32.static] Error 1
make[1]: Leaving directory '/home/daniel/dvlp/mxe'
real    0m13.282s
user    0m11.415s
sys     0m1.520s
------------------------------------------------------------
[log]      /home/daniel/dvlp/mxe/log/gdal_x86_64-w64-mingw32.static
```

3togo:
@coonce, 

How to fix it ?

jonaski:
gdal probably needs to be upgraded, it's at version 2.2.4, and the newest is 3.2.2.
did you try do upgrade it?

tonytheodore:
gdal is hard to update, [workaround from last package build](https://github.com/mxe/mxe/releases/tag/build-2021-01-10) was to disable openjpeg support with:
```
make gdal_CONFIGURE_OPTS=--with-openjpeg=no ...
```

HizzyCO:
Even with the work around mine still fails

`[build]       gdal                   i686-w64-mingw32.static

[...]

PR #2772, opened 2022-02-20
Fix gdal - openjpeg 2.4 incompatibility

brodieG:
I also ran into #2691 and #2610. After applying the patch proposed by @HuidaeCho in #2721 I encountered a build failure caused by (also seen in i686-w64-mingw32):

```
/home/vagrant/mxe/usr/bin/x86_64-w64-mingw32.static-ld: /home/vagrant/mxe/tmp-gdal-x86_64-w64-mingw32.static/gdal-2.2.4/.libs/libgdal.a(openjpegdataset.o):openjpegdataset.cpp:(.text+0x10bc): undefined reference to `__imp_opj_stream_create'
```

The [fix to add tests to the openjpeg build](https://github.com/mxe/mxe/blob/master/src/openjpeg-1-fixes.patch#L19) sets the `-DOPJ_STATIC` flag which causes the library to be build without the [`__declspec(dllimport)`]( https://github.com/uclouvain/openjpeg/blob/883c31dbe09771aab043744ac2b490d7386878e3/src/lib/openjp2/openjpeg.h#L85), and thus symbols without the `__imp_`:

```
nm usr/x86_64-w64-mingw32.static/lib/libopenjp2.a | grep opj_stream | head -n3
0000000000000620 T opj_stream_create
0000000000000720 T opj_stream_default_create
0000000000000060 T opj_stream_default_read
```

`gdal` however includes the openjpeg headers, but does not set the flag, which produces the `__imp_` variety in openjpegdataset.o.

This patch fixes the issue by adding `-DOPJ_STATIC` to the MXE gdal configure to match the `openjpg` library produced by the MXE recipe.  I am a novice at cross compilation (and builds in general), so I don't know if this is appropriate, but the build does succeed with the change.  I tested also the x86_64-w64-mingw32.shared build, and that worked, although it seems weird to set that flag for the shared builds.  It might be better to modify the make files in gdal-2.2.4/frmts/openjpg directly via patch instead of the top level change I made here.

Assuming you agree with this analysis of the problem, ideally you would first accept @HuidaeCho's patch (#2721) and then I can produce a new one with the additional change (or of course you can just make that additional change).

### DIFF
--- a/src/gdal-1-fixes.patch
+++ b/src/gdal-1-fixes.patch
@@ -14,7 +14,7 @@ diff --git a/configure.ac b/configure.ac
 index 1111111..2222222 100644
 --- a/configure.ac
 +++ b/configure.ac
-@@ -1124,7 +1124,7 @@ else
+@@ -1124,7 +1124,7 @@
  
    AC_MSG_RESULT([yes])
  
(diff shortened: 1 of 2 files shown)


brodieG:
Now that I better understand `tools/patch-tool-mxe` I see this patch does not correctly preserve attribution of the changes.  I will provide a new one that does.

mabrand:
Thanks. It's always good to know where the patch comes from, and even better when the patch comes from the upstream project. If this patch is not from upstream, have you proposed it there?

brodieG:
@mabrand thank you for looking at this PR.  I wrongly assumed you would not look at closed ones; I re-opened it as #2774 and failed to reference this closed PR.  Sorry for thus wasting your time.

I have not tried to submit the patch upstream because my ultimate objective here is to update to a new version of GDAL that has completely changed the configure check and does not have the problem we see here.  I wanted to fix this issue with the current version to ensure that my MXE builds are working correctly before I start modifying it.  E.g. right now I'm trying to get postgresql updated to the current version, and before doing so I wanted to confirm that all the postgresql rev-deps continue to work (including GDAL) after my changes.

PR #2774, opened 2022-02-21
Add openjpeg 2.4 capability to gdal

brodieG:
>  Note, this supercedes #2772.  It is not submitted upstream because the current versions of GDAL already fix this issue.  My objective is to eventually update GDAL to the current version, but I'm a long ways from that.

Currently the gdal build is broken as it does not support openjpeg 2.4 and MXE recently updated to that version.  See #2691 and #2610 for reports. @HuidaeCho in #2721 adds support for openjpeg 2.4.  I manually applied the changes in the patch to the sources (so that I can then use `tools/patch-tool/mxe`), but after the version fix I hit:

```
/home/vagrant/mxe/usr/bin/x86_64-w64-mingw32.static-ld: /home/vagrant/mxe/tmp-gdal-x86_64-w64-mingw32.static/gdal-2.2.4/.libs/libgdal.a(openjpegdataset.o):openjpegdataset.cpp:(.text+0x10bc): undefined reference to `__imp_opj_stream_create'
```

This is a mismatch between the MXE openjpeg configuration and MXE gdal likely introduced by the [fix to add tests to the openjpeg build](https://github.com/mxe/mxe/blob/master/src/openjpeg-1-fixes.patch#L19).  That change sets the `-DOPJ_STATIC` flag for MXE openjpeg, so the library builds without [`__declspec(dllimport)`]( https://github.com/uclouvain/openjpeg/blob/883c31dbe09771aab043744ac2b490d7386878e3/src/lib/openjp2/openjpeg.h#L85), and thus symbols without the `__imp_`:

```
nm usr/x86_64-w64-mingw32.static/lib/libopenjp2.a | grep opj_stream | head -n3
0000000000000620 T opj_stream_create
0000000000000720 T opj_stream_default_create
0000000000000060 T opj_stream_default_read
```

This patch adds `-DOPJ_STATIC` to the gdal sub configure file to make it consistent with the openjpeg MXE built library.  I only modified the GNU make file as it seems the  corresponding `nmake` "makefile.vc" is likely irrelavant for MXE builds.  I am a novice at cross compilation (and builds in general), so I don't know if this is appropriate, but the build does succeed with the change including on x86_64-w64-mingw32.shared, although it does seem a bit odd to set that flag for the shared builds.

Most of the credit for this should go to @HuidaeCho and the work he did for #2721.

### DIFF
--- a/src/gdal-1-fixes.patch
+++ b/src/gdal-1-fixes.patch
@@ -5,7 +5,7 @@ Contains ad hoc patches for cross building.
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: Luke Potgieter <fried.roadkill+ght@gmail.com>
 Date: Tue, 29 Aug 2017 16:31:09 +0200
-Subject: [PATCH 1/2] Detect spatialite more effectively than just checking for
+Subject: [PATCH 1/4] Detect spatialite more effectively than just checking for
  headers in the system paths. PostgreSQL detection requires -lpthread when
  building statically.
 
@@ -23,7 +23,7 @@ index 1111111..2222222 100644
  
    if test "${HAVE_PG}" = "yes" ; then
      LIBS=-L`$PG_CONFIG --libdir`" -lpq $LIBS"
-@@ -3557,16 +3557,21 @@ AC_ARG_WITH(spatialite-soname,
+@@ -3598,16 +3598,21 @@ AC_ARG_WITH(spatialite-soname,
  
  HAVE_SPATIALITE=no
  SPATIALITE_AMALGAMATION=no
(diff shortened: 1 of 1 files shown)


HuidaeCho:
@brodieG How is this PR different from my #2721? To my understanding, this PR fixes a new issue introduced by https://github.com/mxe/mxe/blob/master/src/openjpeg-1-fixes.patch#L19 applied after my #2721?

brodieG:
@HuidaeCho that's right, this PR is mostly yours, except that I generated the patch file using `tools/patch-tool-mxe`, with the addition of the -DOPJ_STATIC flag.

mabrand:
Should be fixed by eb7df8f39184da38c47c69e9345f7bf091e165c1 and c2cdd3a06bf9b0e596974d70b241c06889ba16f8.

PR #2778, opened 2022-02-24
fix issue#2775 (documentation)

basiliscos:
pkg m4 definitions are missing, so the issue arises.


danger89:
If `pkgconfig` is missing then you miss a dependency definition in the makefile to **the MXE `pkgconf` package**: https://github.com/mxe/mxe/blob/8d58a506b468f8f4eba4b11af09fa0defb32b1cc/src/pkgconf.mk

tonytheodore:
Thanks

PR #2780, opened 2022-02-25
Remove unnecessary pe-util dependency

danger89:
Remove unnecessary `pe-util` dependency in icu4c package, only causing issues and additional compile time.

Compiled and tested, also works without pe-util dep.


danger89:
@mabrand ping :)

tonytheodore:
I would have expected this to fail on shared builds where `$(TARGET)-peldd` is called to create a dll archive for the test program - it's not critical to the build and can be removed.

cgilles:
This PR will fix this report ?

https://github.com/mxe/mxe/issues/2789

danger89:
> This PR will fix this report ?
> 
> #2789

Maybe. But that was not the intention.

tonytheodore:
@cgilles [your comment](https://github.com/mxe/mxe/issues/2789#issuecomment-1072966309) is the failure I would have expected after this was merged but it's from before.

PR #2781, opened 2022-02-25
Remove unnecessary dependency from harfbuzz

danger89:
Harfbuzz has built-in Unicode-data functions. Also we do not enabled ICU. So remove extra build dependency. 

Do not waste extra build time.

Build still works fine.


danger89:
@mabrand ping :)

danger89:
Thank you for your time and merges.

danger89:
All those changes are also a preparation for the huge gtk3 stack update.

I'm now delivering some changes in advance, to lower the impact later down the line :).

PR #2790, opened 2022-03-09
qt6: Add qtcharts and qtserialport.

nouwaarom:
I used one of the other qt6 modules as template. I tested this and it seems to work.


nouwaarom:
Good question. I have tested both static and shared with cmake. I will test with qmake today

nouwaarom:
The static build with qmake works. I cannot test the shared builds. An executable is generated but I cannot get windeployqt.exe to work so I cannot test that it actually works :(

jonaski:
Use the `tools/copydlldeps.sh`, or use `peldd.exe` on the files manually and copy the dependencies.

PR #2795, opened 2022-03-21
Followed up qt update to 5.15.3

imresteiner:


mabrand:
Thanks. The version numbers are supposed to get updated automatically, but that system is not working at the moment.

PR #2800, opened 2022-03-25
Qt6: properly include conf Makefile

### LINKED ISSUE #2796: qt6-qtshadertools won't be downloaded with make < 4.3, consequently fails to build
Currently debugging why Qt6 cannot be built properly on my Ubuntu 20.04 machine whereas it can be built on my Arch Linux, I stumbled across something funny: on Ubuntu 20.04 `make` simply doesn't want to download the source archive for `qt6-qtshadertools`.

This here is my Ubuntu 20.04 machine:

```
[0 mosu@islena ~/tmp] git clone https://github.com/mxe/mxe
Cloning into 'mxe'...
remote: Enumerating objects: 51598, done.
remote: Counting objects: 100% (1066/1066), done.
remote: Compressing objects: 100% (490/490), done.
remote: Total 51598 (delta 650), reused 882 (delta 549), pack-reused 50532
Receiving objects: 100% (51598/51598), 25.82 MiB | 27.72 MiB/s, done.
Resolving deltas: 100% (37523/37523), done.
[0 mosu@islena ~/tmp] cd mxe/
[0 mosu@islena (master) ~/tmp/mxe] make download-only-qt6-qtshadertools
[create settings.mk]
make: Nothing to be done for 'download-only-qt6-qtshadertools'.
[0 mosu@islena (master) ~/tmp/mxe] ls pkg
ls: cannot access 'pkg': No such file or directory
[2 mosu@islena (master) ~/tmp/mxe]
```

and this is my Arch Linux:

```
[0 mosu@sweet-chili ~/tmp] git clone https://github.com/mxe/mxe
Cloning into 'mxe'...
remote: Enumerating objects: 51598, done.
remote: Counting objects: 100% (1066/1066), done.
remote: Compressing objects: 100% (490/490), done.
remote: Total 51598 (delta 649), reused 886 (delta 549), pack-reused 50532
Receiving objects: 100% (51598/51598), 25.84 MiB | 18.47 MiB/s, done.
Resolving deltas: 100% (37522/37522), done.
[0 mosu@sweet-chili ~/tmp] cd mxe
[0 mosu@sweet-chili (master) ~/tmp/mxe] make download-only-qt6-qtshadertools
[create settings.mk]
[download]    qtshadertools-everywhere-src-6.2.4.tar.xz
[0 mosu@sweet-chili (master) ~/tmp/mxe] ls pkg
qtshadertools-everywhere-src-6.2.4.tar.xz

```

The logical consequence is that `qt6-qtshadertools` cannot be built, causing the rest of my Qt6 requirements to fail.

This is a freshly cloned MXE revision dabfa7bfe6ec47ddcc4dd779eb316d7d090c8112 in both cases as can be seen above. A plain, new checkout without any further modification runs into this issue.

`make` on Ubuntu is 4.2.1 and 4.3 on Arch.

Retrying with `make -d download-only-qt6-qtshadertools` yields [this output](https://gist.github.com/123166644e5a2b6ed488a8697eab68d9). What's interesting to see here starts around line 215787 (only visible in full download, not in the web UI):

```
215785:Considering target file 'download-only-qt6-qtshadertools'.
215786: File 'download-only-qt6-qtshadertools' does not exist.
215787:  Considering target file 'download-only-'.
```

Comparing this with debug output from my Arch where it works it should look like this:

```
215785:Considering target file 'download-only-qt6-qtshadertools'.
215786: File 'download-only-qt6-qtshadertools' does not exist.
215787:  Considering target file 'download-only-qtshadertools-everywhere-src-6.2.4.tar.xz'.
```

stloeffler:
All Qt6 Makefiles that depend on `QT6_METADATA` need to include `qt6-conf.mk` (instead of relying on `make` to automatically parse it first). In turn, `qt6-conf.mk` depends on `qt6-qtbase.mk` for some of its variable assignments.

For make < 4.3, this is essential as those versions don't sort globbing results alphabetically (see #2796 and [make bug 52076](https://savannah.gnu.org/bugs/?52076)). For later versions of make, globbing results are sorted, but still some package names (e.g., `qt6-qt5compat`) would require special treatment as they are sorted before `qt6-qtbase` (which includes some required variables).

Caveats of this patch: it results in `qt6-conf.mk` (and in turn `qt6-qtbase.mk`) being parsed multiple times when running make. However, IMO, this is still the most reliable way (especially in case files are added/removed in the future).

Fixes #2796

### DIFF
--- a/src/qt/qt6/qt6-conf.mk
+++ b/src/qt/qt6/qt6-conf.mk
@@ -1,5 +1,7 @@
 # This file is part of MXE. See LICENSE.md for licensing information.
 
+include src/qt/qt6/qt6-qtbase.mk
+
 PKG            := qt6-conf
 $(PKG)_VERSION  = $(qt6-qtbase_VERSION)
 $(PKG)_DEPS    := qt6-qtbase
--- a/src/qt/qt6/qt6-qt5compat.mk
+++ b/src/qt/qt6/qt6-qt5compat.mk
@@ -1,16 +1,12 @@
 # This file is part of MXE. See LICENSE.md for licensing information.
 
+include src/qt/qt6/qt6-conf.mk
+
 PKG := qt6-qt5compat
 $(eval $(QT6_METADATA))
 
-# The following variables don't seem to be set correctly by the $(eval) above,
-# presumably because this file is included before qtbase (qt5compat < qtbase).
-qt6-qt5compat_SUBDIR    = $(subst qtbase,qt5compat,$(qt6-qtbase_SUBDIR))
-qt6-qt5compat_FILE      = $(subst qtbase,qt5compat,$(qt6-qtbase_FILE))
-qt6-qt5compat_URL       = $(subst qtbase,qt5compat,$(qt6-qtbase_URL))
-
 $(PKG)_CHECKSUM := 5de2b9e25bf7de161fbb88ecdd468ed1788bc899392fc05ed80aa590ebb352fa
-$(PKG)_DEPS     := cc qt6-qtbase
+$(PKG)_DEPS     := cc qt6-conf qt6-qtbase
 
 QT6_PREFIX   = '$(PREFIX)/$(TARGET)/$(MXE_QT6_ID)'
 QT6_QT_CMAKE = '$(QT6_PREFIX)/bin/qt-cmake-private' \
--- a/src/qt/qt6/qt6-qtcharts.mk
+++ b/src/qt/qt6/qt6-qtcharts.mk
@@ -1,10 +1,12 @@
 # This file is part of MXE. See LICENSE.md for licensing information.
 
+include src/qt/qt6/qt6-conf.mk
+
 PKG := qt6-qtcharts
 $(eval $(QT6_METADATA))
 
 $(PKG)_CHECKSUM := 3c2e6267ef0fb5345c7737e3a12e09ce9ec09117792e78af9a8617f828b2249a
-$(PKG)_DEPS     := cc qt6-qtbase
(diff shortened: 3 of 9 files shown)


tonytheodore:
I can see the same issue on Ubuntu with make 4.2.1, but not on Darwin with any version - tested 3.81, 4.0, 4.1, 4.2.1.

I'll revisit this next week, I'm not happy with the `eval` in `src/*.mk` files.

stloeffler:
I basically only followed the pattern that was established in the initial commit of qt6 support (e.g., [src/qt/qt6/qt6-qttools.mk@772b91c](https://github.com/mxe/mxe/blob/772b91c6b6d2f58bab572902d0409bc48c2e6829/src/qt/qt6/qt6-qttools.mk)). That said, the `eval` is by no means essential. One aspect that (probably) *is* important is to have the same version for all (main line) qt6 files, i.e., have `$(PKG)_VERSION = $(qt6-qtbase_VERSION)`. `$(PKG)_WEBSITE` can probably also refer to its qtbase counterpart.
`$(PKG)_DESCR` should probably be set per-package to be more descriptive, and `$(PKG)_SUBDIR`, `$(PKG)_FILE`, `$(PKG)_URL` can easily be set per-package (possibly with a sort of `qt6_BASE_URL` defined in `qt6-conf.mk`).
So really, getting rid of the `eval` is a matter of a bunch of copy-pastes.
Note that in any case each qt6 `.mk` file has to include `qt6-qtbase.mk` (either directly or indirectly via `qt6-conf.mk`) to ensure (at least) `qt6-qtbase_VERSION` is available - unless there is some `make`-magic with assignment operators that makes this happen regardless of include order.

Issue #2799, opened 2022-03-24
Problem cross-compiling Qt6 with CMake


ju3:
Hi,

I'm trying to cross-build (x86_64-w64-mingw32.shared) a Qt6 binary with CMake in a Docker container. But when calling the MXE CMake wrapper script, the configuration fails due to missing Qt6WidgetsTools (```-- Could NOT find Qt6WidgetsTools (missing: Qt6WidgetsTools_DIR```).

I've uploaded Dockerfile, CMakeLists.txt and build outputs here: https://gist.github.com/ju3/6b6212f24718770f3e89e6b385c9ccd5

Doing the same with a Qt5 project (and the corresponding qt5.mk build) works fine.

Thank you for your help.

stloeffler:
One thing I did that might help here is to set `-DQT_HOST_PATH="/opt/mxe/usr/x86_64-pc-linux-gnu/qt6"` when invoking cmake to tell it where to find native qt tools (such as moc, uic, rcc, ...) - see also the discussion at #2768.

ju3:
Thank you very much, it worked!

Adding the file `/opt/mxe/usr/x86_64-w64-mingw32.shared/share/cmake/mxe-conf.d/qt6.cmake` with content `set(QT_HOST_PATH "/opt/mxe/usr/x86_64-pc-linux-gnu/qt6")` makes CMake work out of the box.

ju3:
I've created pull request #2802.

PR #2802, opened 2022-03-28
qt6-qtbase: Add CMake package configuration for Qt6 host path

ju3:
The Qt6 cross-build uses tools from the host build and thus depends on it. CMake configuration of target projects fails if the path to the Qt host build is missing. This pull requests adds a Qt6 package configuration for the CMake target build.

Refers to #2799.


mabrand (approved):
Makes sense to me.

mabrand:
Tony, You might want to take a look at this. I'm not completely confident in my own cmake fu.

tonytheodore:
Only concern would be confusing any qt5 builds, but this was [introduced in qt6](https://doc.qt.io/qt-6/cmake-variable-qt-host-path.html).

PR #2811, opened 2022-04-08
Qt6: add qtmultimedia

mbunkus:
Modeled on qt6-qtimageformats. Patch done via `patch-tool-mxe`. Tested with my own MKVToolNix: successful audio playback on Windows.

The patch is needed as the library & include file names are all over the place:

* libraries are all lowercase in the file system, but the `cmake` snippet tries to find them in mixed case
* several includes that are all lowercase in the file system are spelled in mixed case
* the `QtCore` include sub-directory is misspelled `qtcore`

### DIFF
--- a/src/qt/qt6/qt6-qtmultimedia-1-fixes.patch
+++ b/src/qt/qt6/qt6-qtmultimedia-1-fixes.patch
@@ -1,0 +1,191 @@
+This file is part of MXE. See LICENSE.md for licensing information.
+
+Contains ad hoc patches for cross building.
+
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Moritz Bunkus <mo@bunkus.online>
+Date: Fri, 8 Apr 2022 21:22:34 +0200
+Subject: [PATCH 1/1] fix file name wrt. case
+
+
+diff --git a/cmake/FindWMF.cmake b/cmake/FindWMF.cmake
+index 1111111..2222222 100644
+--- a/cmake/FindWMF.cmake
++++ b/cmake/FindWMF.cmake
+@@ -22,11 +22,11 @@ find_library(WMF_UUID_LIBRARY uuid HINTS ${CMAKE_C_IMPLICIT_LINK_DIRECTORIES})
+ find_library(WMF_MSDMO_LIBRARY msdmo HINTS ${CMAKE_C_IMPLICIT_LINK_DIRECTORIES})
+ find_library(WMF_OLE32_LIBRARY ole32 HINTS ${CMAKE_C_IMPLICIT_LINK_DIRECTORIES})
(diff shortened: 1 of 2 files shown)


mbunkus:
Argh… for some reason it doesn't compile at the moment, even though I was able to compile it successfully earlier. Probably a missing dependency & therefore packages are compiled in the wrong order. The actual error message is:

> Unknown CMake command "qt_internal_add_shaders"

I'll look into it tomorrow. Please don't merge until I've had a chance to look into it & fix the issue.

mbunkus:
It was a missing dependency on `qt6-qtshadertools`, which I've added. Now building `qt6-qtmultimedia` from a clean tree works just fine.

Ready to merge!

tonytheodore:
Thanks!

tonytheodore:
I'm seeing some errors, log https://gist.github.com/tonytheodore/7bdbaf1c1a2cd24b90a46bbaaa1ac830

mbunkus:
I'll investigate.

mbunkus:
A clean checkout of revision 3b5610218c83f382e1e8f76ccd7b38f5f588cdf9 followed by `make MXE_TARGETS=i686-w64-mingw32.static qt6-qtmultimedia` works fine.

My guess is that the issue is due to you having GStreamer installed, which I don't have. I'll look into forcing it off even if it is installed.

mbunkus:
Hmmm nope, simply installing `gstreamer` doesn't actually break the build, probably because `qt6-qtmultimedia` still fails to detect some gstreamer components (e.g. it finds `gstreamer-1.0` and `gstreamer-base-1.0` but not `gstreamer-audio-1.0` and others).

Can you please post which packages you have installed & which packages `pkg-config` lists:

```
ls usr/i686-w64-mingw32.static/installed
i686-w64-mingw32.static-pkg-config --list
```

tonytheodore:
Unfortunately, I just did a `make clean` to try qt6-qtmultimedia on it's own.

Previously, I was building:
```
qt5 qt6 \
gtkmm3 \
gst-libav gst-plugins-bad gst-plugins-base gst-plugins-good gst-plugins-ugly \
sdl2_gfx sdl2_image sdl2_mixer sdl2_net sdl2_ttf \
wxwidgets
```

mbunkus:
Those `gst-` look like likely culprits. I'll start with those. Thanks.

tonytheodore:
> Those gst- look like likely culprits

I can confirm clean build of qt6-qtmultimedia works without those installed.

mbunkus:
And I can confirm that installing `gst-*` prior to compiling `qt6-qtmultimedia` triggers the problem. I'm currently looking into how to disable gstreamer detection unconditionally.

mbunkus:
Fix proposed [here](https://github.com/mxe/mxe/pull/2812).

PR #2812, opened 2022-04-09
Qt6 qtmultimedia: fix compilation if gstreamer is installed

mbunkus:
This forces gstreamer detection off. Otherwise cmake will complain
that EGL isn't found.


tonytheodore:
Thanks!

Newer versions of mesa [have EGL](https://www.phoronix.com/scan.php?page=news_item&px=Microsoft-Mesa-EGL-Windows) so we might be able to re-enable it in the future.

PR #2824, opened 2022-04-26
Improve host build of cmake (fix install locations, default to Unix Makefiles)

kalibera:
This pull request includes small fixes needed to make the host "cmake" work without a wrapper or overrides. It fixes cmake-host to take into account cmake patches and adds two. One patch changes the default generator to "Unix Makefiles", so that it can be used with the host toolchain built by MXE by default. The other patch fixes host cmake installation location so that cmake finds its own files. Without that patch, one gets

```
$ cmake --version
CMake Error: Could not find CMAKE_ROOT !!!
CMake has most likely not been installed correctly.
```
The two patches are based on Msys2 (0003-Fix-install-destinations.patch is re-used, 0005-Default-to-ninja-generator.patch was updated for a different default).

### DIFF
--- a/plugins/examples/host-toolchain/README.md
+++ b/plugins/examples/host-toolchain/README.md
@@ -31,7 +31,7 @@ FC: no differences encountered
 make cmake-host MXE_PLUGIN_DIRS=plugins/examples/host-toolchain/
 ```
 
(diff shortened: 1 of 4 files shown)


tonytheodore:
Nice - thanks!

PR #2827, opened 2022-04-26
libzmq: disable IPC to allow building.

kalibera:
libzmq cannot be built because of errors like this:

```
tmp-libzmq-i686-w64-mingw32.static/zeromq-libzmq-c89390f/src/ipc_address.hpp:40:10: fatal error: sys/socket.h: No such file or directory
   40 | #include <sys/socket.h>
      |          ^~~~~~~~~~~~~~
```
This patch simply disables IPC to make it build, but, there may be a better way (?epoll poller).


tonytheodore:
> but, there may be a better way (?epoll poller)

They have an internal copy of `wepoll`, but someone familiar with it will need to enable it.

Thanks.

PR #2828, opened 2022-04-26
cfitsio: include description when generating .pc file

kalibera:
A description seems to be required when using a newer pkg-config (and does no harm with the one used currently by MXE).

With pkg-config 1.8.0 and without this fix, I get this error during cfitsio build, when it is linking the test:

```
Package cfitsio was not found in the pkg-config search path.
Perhaps you should add the directory containing `cfitsio.pc'
to the PKG_CONFIG_PATH environment variable
```


tonytheodore:
Thanks!

PR #2830, opened 2022-05-01
docs: add openssl and which Fedora packages requirements

stefanha:
The 'openssl' and 'which' required packages may require installation on
Fedora.

Most systems will have these packages installed already, but they are
not installed by default on the
registry.fedoraproject.org/fedora-minimal container image.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -879,6 +879,7 @@ USE_OSGPLUGIN(&lt;plugin2&gt;)
     libtool \
     lzip \
     make \
+    openssl \
     openssl-devel \
     p7zip \
     patch \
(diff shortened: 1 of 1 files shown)


tonytheodore:
Are these in the `epel-release` package (~20 lines above)?

stefanha:
> Are these in the `epel-release` package (~20 lines above)?

`epel-release` is not available for Fedora, only CentOS/RHEL.

I looked at the `epel-release.el9` package anyway to see if it has dependencies that would bring in openssl and which. It only contains yum .repo files that add the EPEL repository needed for some of the packages already listed in the mxe requirements. So I don't believe `epel-release` helps with this issue even on CentOS/RHEL.

tonytheodore:
Thanks

PR #2833, opened 2022-05-04
tidy-html5: update to 5.8.0

kalibera:
Update to the current version.

### DIFF
--- a/src/tidy-html5.mk
+++ b/src/tidy-html5.mk
@@ -4,8 +4,8 @@ PKG             := tidy-html5
 $(PKG)_WEBSITE  := https://www.html-tidy.org/
 $(PKG)_DESCR    := HTML/XML syntax checker and reformatter
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 5.6.0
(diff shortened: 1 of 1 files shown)


src/tidy-html5.mk
```diff
@@ -17,8 +17,8 @@ define $(PKG)_BUILD
     $(MAKE) -C '$(BUILD_DIR)' -j 1 install
 
     $(if $(BUILD_STATIC),
-        cd '$(PREFIX)/$(TARGET)/lib' && mv libtidys.a libtidy.a,
-        rm -f '$(PREFIX)/$(TARGET)/lib/libtidys.a')
+        cd '$(PREFIX)/$(TARGET)/lib' && mv libtidy_static.a libtidy.a,
+        rm -f '$(PREFIX)/$(TARGET)/lib/libtidy_static.a')
```
mabrand, line 21: Does line 21 serve any purpose? Line 20 renames the file that would otherwise be deleted in line 21.

kalibera: It is a conditional - the diff context shown here is too small. Line 20 is executed in static build, line 21 in dynamic build.

PR #2834, opened 2022-05-05
gcc: enable precise timers (clock_gettime) in the C++ library

kalibera:
This makes the C++ library use clock_gettime and hence more precise timers. Without this setting, system_clock from <chrono> only uses gettimeofday. Msys2 enables the precise timers as well in their builds.


tonytheodore:
Thanks

PR #2835, opened 2022-05-05
boost: remove file accidentally included in boost release (with space in its name)

kalibera:
Boost 1.78.0 includes accidentally file "collection_size_type copy.hpp", which is not needed but has a space in its name, which may cause trouble for further processing. This patch deletes it.

https://github.com/boostorg/serialization/commit/07c99b8649d7bb900271ca93ba3eea9dcb478f42#diff-c7259865ae65dd94c715482e2abf4fbb3ee9f77e63eb1b6677eada21bd239b08


tonytheodore:
Thanks

Issue #2862, opened 2022-06-30
"make gd" only builds static libraries


markhamnr:
It seems that `gd.mk` is written to only build a statically linked gd and not a shared library.  I looked around but couldn't find any particular reason for this.  Does anyone more familiar with the codebase know why this is?

It seems like it would be good to either remove this limitation, or at least make clear why it exists.

markhamnr:
Closing since my PR was merged.

PR #2863, opened 2022-06-30
update libgd to 2.2.5 and enable shared libraries for #2862

markhamnr:
Looks like the issue I reported earlier was because libgd 2.1 used an old libtool that didn't handle 64-bit libraries well.  Updating to libgd 2.2 fixed that and is presumable an improvement anyway.


mabrand:
thanks

PR #2873, opened 2022-08-17
upgrade spatialiate to 5.0.1

brodieG:
### Overview

We need to upgrade spatialite to > 5 so that we can upgrade GDAL to 3.5.  This
is a follow up to the [upgrade to postgresql][1] for the same reason.  There will be
further follow-ups culminating eventually in getting GDAL to 3.5.

With this patch applied all combinations of spatialite i686/x86 static/shared
build.

To effect the spatialite upgrade several changes were needed, detailed below.

### Geos Upgrade

We need to upgrade GEOS as spatialite requires >= 3.7.

### Minizip Changes

Spatialite adds a dependency to minizip, however it expects the minizip that
is available from e.g. [debian distribution][2], which is the older one that is
distributed with zlib.  The changes included here are to allow spatialite to
build with the newer fork of minizip that MXE uses.

One small change was required for minizip: declare the openssl dependency.

### Disable Rttopo

Spatialite also adds a dependency to rttopo, but we disable it as it is not
needed for gdal, and it is not currently part of MXE.

### Reverse Dependencies

I checked potentially affected reverse dependencies of spatialite and other
things I changed to make sure they did not become worse after updates.

#### Gdal

Reverse dependency gdal is currently broken due to openjpeg version
check issues such like [we've seen in the past][3].  Since that fix, a new
version of openjpeg is on MXE.  The new version of gdal fixes the check to be
robust to version changes, but to ensure we're not breaking the current version
we modifies the -1-fixes.patch to work with the current version of MXE openjpeg.
With this patch gdal builds on all combinations of spatialite i686/x86
static/shared.

#### Ossim

Geos upgrade breaks [ossim][4], which is then later fixed (possibly addressed
[here][5]), so we upgrade ossim to the latest release.
 
The patches in ossim-1-fixes.patch are already merged upstream (except they use
BUILD_SHARED_LIBS instead of BUILD_SHARED).

In order to get Ossim to build I had to add an explicit dependency to jsoncpp.
While ossim bundles an "amagamated" jsoncpp, the detection scripts for it assume
that there exists a parent folder called that contains both source and build
directories, and that the source dir is called "ossim" whereas e.g. for us the
folder is called "ossimlabs-ossim-1af3774" and I have not figured out a way to
tickle MXE to name that folder "ossim" while also working correctly.

Finally, I had to modify ifdef code for handling use of "sys/ioctl.h".

Note: shared builds fail prior to any changes with errors like:

      /home/vagrant/mxe/usr/x86_64-pc-linux-gnu/bin/i686-w64-mingw32.shared-g++ -O3 -DNDEBUG -shared -o ../bin/libossim.dll -Wl,--out-implib,../lib/libossim.dll.a -Wl,--major-image-version,0,--minor-image-version,0 -Wl,--whole-archive CMakeFiles/ossim.dir/objects.a -Wl,--no-whole-archive @CMakeFiles/ossim.dir/linklibs.rsp
      /home/vagrant/mxe/usr/bin/i686-w64-mingw32.shared-ld: CMakeFiles/ossim.dir/objects.a(ossimImageHandlerFactory.cpp.obj):ossimImageHandlerFactory.cpp:(.text+0x1cbe): undefined reference to `TIFFSetErrorHandler'

My patches do not fix these errors, but since they are pre-existing I do not
address them.

[2]: https://github.com/madler/zlib/tree/master/contrib/minizip
[3]: https://github.com/mxe/mxe/commit/c2cdd3a06bf9b0e596974d70b241c06889ba16f8
[4]: https://github.com/ossimlabs/ossim/issues/79
[5]: https://github.com/ossimlabs/ossim/commit/c456397821cf369af4d9ff1fb60a5dfd1bcefc24


tonytheodore:
> except they use BUILD_SHARED_LIBS instead of BUILD_SHARED

BUILD_SHARED_LIBS is a [standard cmake flag](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) but the mxe toolchain sets both.

> I have not figured out a way to tickle MXE to name that folder "ossim" while also working correctly

I had a quick look and it's too much fiddling around.

> Note: shared builds fail prior to any changes with errors like:

I'm seeing the same.

> culminating eventually in getting GDAL to 3.5

Looking forward to it.

Thanks!

PR #2883, opened 2022-10-01
qt6-qtbase: Disable shared libraries for native build

jonaski:
Fixes the following build error on openSUSE:

```
/usr/lib64/gcc/x86_64-suse-linux/12/../../../../x86_64-suse-linux/bin/ld: /__w/jonaski-mxe/jonaski-mxe/usr/x86_64-pc-linux-gnu/lib/libzstd.a(zstd_common.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
/usr/lib64/gcc/x86_64-suse-linux/12/../../../../x86_64-suse-linux/bin/ld: failed to set dynamic section sizes: bad value
[348/1264] Building C object src/3rdparty/freetype/CMakeFiles/BundledFreetype.dir/src/bdf/bdf.c.o
ninja: build stopped: subcommand failed.
```


jonaski:
This breaks qtdeclarative build. New fix in #2889

PR #2887, opened 2022-10-01
qt6-qtmultimedia: Update patch

jonaski:
Fixes the following build error:
```
/home/jonas/Projects/jonaski-mxe/tmp-qt6-qtmultimedia-x86_64-w64-mingw32.static/qtmultimedia-everywhere-src-6.4.0/src/plugins/multimedia/ffmpeg/qwindowscamera.cpp:14:10: fatal error: Mfreadwrite.h: No such file or directory
   14 | #include <Mfreadwrite.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```


mabrand:
Patch is obviously correct. This problem didn't break my build, presumably because the file wasn't compiled because a feature (ffmpeg?) wasn't detected. Maybe we need to add the dependency to qt6-qtmultimedia for consistent results. What do you think?

PR #2768, opened 2022-02-18
fix(qt6): Disable check for OpenGL

lbguilherme:
Currently, Qt6 fails to build with the following message. As OpenGL is configured as dynamic, this isn't really a problem. This PR ignores the message.

> The OpenGL functionality tests failed!  You might need to modify the include and library search paths by editing QMAKE_INCDIR_OPENGL[_ES2], QMAKE_LIBDIR_OPENGL[_ES2] and QMAKE_LIBS_OPENGL[_ES2] in the mkspec for your platform.

PR #2918, opened 2022-11-15
Update GDAL to 3.6.0

brodieG:
# Overview

This patch brings GDAL up to 3.6.0.  This is a major upgrade with lots of moving parts.  I tried to find things to pare down to produce a series of smaller PRs, but the best I could do was the upgrades to [postgresql][1] and [spatialite][2]  that you have already merged into MXE.

Further upgrades become incompatible with the current version of GDAL, or with current versions of other dependencies.  Thus unfortunately this large pull request to bring everything up to more current versions.

For each package I changed, I checked the direct reverse dependencies (and if I made changes to those, then continued the process to their reverse dependencies).  I did this across the four permutations of x86/i686 static/shared.  Not all revdeps build on all four configurations, but this also true on the current master branch (see below)

Additionally, I tested running the x86 static and shared versions of the GDAL binaries on a windows machine (tested gdalinfo, gdal_countour, and  gdal_transform).

This is a collaborative effort with Tomas Kalibera (@kalibera) from the R project.

# Host System Requirements

I ran into two missing system requirements that I added locally with `apt-get` on my Ubuntu system:

- [sqlite3 binary][3] (not the library) for PROJ.
- [opengl][4] (known MXE issue with e.g. Ubuntu).

Would it make sense to add these to the MXE system requirements?

----

# GDAL Dependencies Requiring Updates / Fixes

## Expat

Comes with pre-built CMake files despite being autotools on MXE.  I delete those files to avoid failures on static builds.

Some Expat reverse dependencies fail to be build, but those also fail on the MXE master branch (see "GDAL Dependency Reverse Dependency Section").

## HDF4

I added pkgconfig files via the mk file to ensure linkage happens in the correct order for netcdf.

## HDF5

## netcdf

## AEC

Needed for SZIP support for HDF5.  It does not support only-static or only-shared via configuration options in cmake so I modified the CMakeLists.txt to allow it.

## kealib

This is a new package.  Strictly speaking this could be submitted as a separate PR, but it is needed for my GDAL application (R Project Toolchain).

Requires similar libname cleanup to netcdf.

## Minizip

In order to build spatialite correctly I had to upgrade up to 3.0.1.  Upgrading further (e.g. 3.0.2) causes the shared library to fail to build properly.  I have not looked into why this is.

## Armadillo

Specify HDF5 include directory directly.

## Postgresql

Required edits to the pkgconfig file for correct detection for GDAL static builds.  The "-lz" flag was not generated (similar to [this issue][11]).

## Blosc

This is a new package.  Strictly speaking this could be submitted as a separate PR, but it is needed for my GDAL application (R Project Toolchain).  I tried briefly to add a cmake EXPORT but ran into problems so just resorted to patching GDAL to use the pkgconfig file (see below).

### PROJ

PROJ 6 or later [is required now][15] (vs v5 on MXE currently) for GDAL, so we update.

Additionally, PROJ has several [optional but recommended requirements][16] I added to the dependency list:

### libgeotiff

Update to 1.5.1 for compatibility with the PROJ update.  Patches no longer needed as either fixed or obviated by changes upstream.

----

# GDAL Dependency Reverse Dependencies

These are reverse dependencies of the dependencies of GDAL.  For example, GDAL depends on HDF5, and so does matio.  Since I modified HDF5, I checked both GDAL and matio (and other revdeps of HDF5).

## matio, revdep of HDF5

Matio has version specific HDF5 interfaces, so I updated it to a recent version with support for HDF5 1.12.

The new version requires a work-around for [cross compilation][8].

## Ossim, revdep of HDF5

This is not needed by GDAL.  I get the following (X == fail), both on my branch and the MXE master branch:

             |      i686       |       x86
             | static | shared | static | shared
    ossim    |   ok   |    X   |   ok   |    X

This does not tie out with what build-matrix.html reports, which has all builds succeeding for these packages.  I'm a little confused by this as @tonytheodore sees the same as me for at least [ossim][13].

## VTK and ITK, revdeps of HDF5 and of Expat

These fail across the board for me, but that is also true when I build them separately from the master branch, so not a change for the worse AFAICT (notwithstanding inconsistency with the build-matrix.html data).

             |      i686       |       x86
             | static | shared | static | shared
    vtk      |    X   |    X   |    X   |    X
    itk      |    X   |    X   |    X   |    X

## netcdf-cxx4, revdep of netcdf

Same problem as netcdf.  I did not spend the time to figure out exactly how to prevent the bad library name from netcdf from spreading; instead I hack linklibs.rsp file directly after configuration.

## cegui, revdep of Expat

Fails because its dependency freeimage fails with a CXX17 error, but this also happens on the MXE master branch (i686-static).  Since it is not a change for the worse (notwithstanding inconsistency with the build-matrix.html data) I have not looked further .

----

# GDAL

I switched the build to CMake as other build platforms become unsupported with 3.6.0.  The CMake build auto-detects available packages and enables features accordingly.

Most of the patches are fixing the package auto-detection which does not work well for non-CMake packages, or for CMake packages that do not correctly EXPORT themselves.  Most of the fixes involve hacking the detection code to use the pkgconfig files and generate the corresponding `target_link_libraries` directives.

I removed a setting of __MSVCRT_VERSION__ no longer needed.

----

# GDAL RevDeps

## openscenegraph

tiff was not linked properly in the shared builds so I hacked in a fix.  This is likely the same type of issue we run into with GDAL where the package is expecting fully CMake compliant dependencies, but the world is a harsh place...

This is also broken in the current master branch so is not a regression introduced by other changes in this patch.

## osgearth

`osgearth` is broken in the master MXE branch because the current version is not compatible with the new [GEOS][14].  I upgraded GEOS for the new [spatialite][2], but because `osgearth.mk` didn't declare a dependency on GEOS I didn't notice this issue at the time.

I started trying to fix it but ran into some issues that I don't quite know how to resolve simply.  I provide a partial patch.

The problem is that the current version uses git submodules for some of the third party libraries, and those are not included as part of the build tarball on github, so the build fails.  I _think_ fixing this will require changes on the MXE build scripts to clone the repository with `--recurse-submodules` instead of using the Github pre-built tarball.

---

Thanks for your work building and keeping MXE up and running.


brodieG:
@tonytheodore any interest in reviewing this?  Or giving some guidance of what I can do to get you interested?  I realize this is a very big PR, but not sure if there is much I can do to make it substantially smaller.

brodieG:
Ok, thanks for the feedback.  There are a few small things that it looks like I can chip off.  I'll work on trying to get as much of that through separately.  But at the end of the day we'll still be left with a big PR because GDAL is so far behind it's not possible to update all its dependencies to what the current GDAL requires without breaking the current MXE GDAL.

brodieG:
Closing this in favor of #2943 and all the above referenced components.  @mabrand let me know if you want me to do anything more here.

PR #2924, opened 2022-12-02
Add Apache Thrift 0.17.0 runtime libraries.

hdclark:

  * install .pc file,

Thanks!

### DIFF
--- a/src/thrift.mk
+++ b/src/thrift.mk
@@ -1,0 +1,34 @@
+PKG             := thrift
+$(PKG)_WEBSITE  := https://thrift.apache.org/
+$(PKG)_DESCR    := Apache Thrift (runtime libraries only)
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 0.17.0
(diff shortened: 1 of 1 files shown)


hdclark:
This commit adds Apache Thrift RPC v0.17.0 runtime libraries (note: not the thrift compiler).

The build stage includes a minor 'dynamic patch' to fix Windows case-insensitivity. An upstream patch has been proposed, see https://github.com/apache/thrift/pull/2649 .

hdclark:
Also note that I'm planning to use this for `DICOMautomaton`, which already uses MXE for Windows builds. (Thanks!!)

Reference issue tracking Thrift additions for `DICOMautomaton`: https://github.com/hdclark/DICOMautomaton/issues/21

hdclark:
@mabrand: is this something you could take a look at?

hdclark:
@mabrand any possibility of this being merged? I'm planning to be away for the next few weeks and want to be around if there are questions or concerns.

hdclark:
Bump. Anyone available to comment on this?

hdclark:
Bump. Updated to version 0.18.1.

PR #2935, opened 2022-12-20
Update PROJ to 8.2.1 + related fixes

brodieG:
This is another small part of the GDAL 3.6 upgrade (https://github.com/mxe/mxe/pull/2918) (@mabrand).

PROJ 6 or later [is required now][15] (vs v5 on MXE currently) for GDAL, so we update.  Additionally, PROJ has several [optional but recommended requirements][16] I added to the dependency list.

Libgeotiff is not compatible with the new version, so I also had to upgrade that (the upgrade obviates previously required patches).

With these changes all the PROJ revdeps pass, except for ossim shared builds, which as noted in the [prior spatialite upgrade were already failing](https://github.com/mxe/mxe/pull/2873#ossim).


brodieG:
@mabrand noticed you merged the other PRs but not this one (thanks!).  I'm going to see if I can do the libgeotiff separately first to split this PR into two pieces.  If I can I'll close this one and re-open it as one for libgeotiff, and once that's done, one for proj.

Notwithstanding I think I'm running out of things I can chip off and most likely HDF5/netcdf/gdal will need to be done together.

brodieG:
Nope, new libgeotiff requires a newer PROJ.  There is probably some way to make it work, but that seems like potentially a lot of effort to reach a transient state that will then converge back to this patch.

```
Failed to build package libgeotiff for target i686-w64-mingw32.static!
checking for PROJ >= 6 library... checking for proj_create_from_wkt in -lproj... no
checking for internal_proj_create_from_wkt in -lproj... no
checking for internal_proj_create_from_wkt in -linternalproj... no
configure: error: PROJ 6 symbols not found
make[1]: *** [Makefile:900: build-only-libgeotiff_i686-w64-mingw32.static] Error 1
make[1]: Leaving directory '/home/vagrant/mxe-master'
real	0m13.188s
user	0m11.549s
sys	0m0.740s
```

brodieG:
Fyi, 2dd7a867 (upgrade libwebp to 1.3) seems to break this.  Likely: https://github.com/libjxl/libjxl/issues/2072.  I'll submit a fix soon.

brodieG:
This patch needs to be updated due to https://github.com/mxe/mxe/commit/d8bfdc137d7c01e0458861303b25ecd1d7cfc5e0.

Keeping requires instead of includes allows relying on pkgconfig, which for the pkgconfig file seems better?  @mabrand I can address if you're still interested in the #2948.

brodieG:
I removed the part of the patch made redundant by https://github.com/mxe/mxe/commit/d8bfdc137d7c01e0458861303b25ecd1d7cfc5e0.

mabrand:
Could you please just squash the last 2 commits? Then I think we can merge this. Sorry about the delay.

brodieG:
Last two commits squashed into one.

PR #2942, opened 2022-12-29
Add -lz to postgresql pkgconfig

brodieG:
Last small piece of #2918.


brodieG:
So from what I can tell this patch is not needed anymore.  I'm not sure what changed.  There have been a couple of Openssl updates which might be the reason.  I tried building both x86 static and shared without the patch and the build succeeds.

PR #2943, opened 2022-12-29
Minimal GDAL 3.6.0 Upgrade

brodieG:
# Overview

This is a minimal version of #2918, having removed all changes that could be made stand-alone, and also removed additional optional packages.  **Prior to merging this one, the following outstanding PRs need to be merged in**:

* #2935
* #2941 
* #2942 

HDF5, Netcdf, Gdal, and a few other small things had to be upgraded together as the old versions don't work with the new and vice-versa.

For each package I changed, I checked the direct reverse dependencies (and if I made changes to those, then continued the process to their reverse dependencies).  I did this across the four permutations of x86/i686 static/shared.  Not all revdeps build on all four configurations, but this also true on the current master branch.  Note I did this with the original #2918, assuming that all the merged and to be merged pieces are equivalent (they should be).  Please see #2918 for details.

This is a collaborative effort with Tomas Kalibera (@kalibera) from the R project.

# Host System Requirements

I ran into two missing system requirements that I added locally with `apt-get` on my Ubuntu system:

- [sqlite3 binary][3] (not the library) for PROJ.
- [opengl][4] (known MXE issue with e.g. Ubuntu).

Would it make sense to add these to the MXE system requirements?

----

# GDAL Dependencies Requiring Updates / Fixes

## Expat

Comes with pre-built CMake files despite being autotools on MXE.  I delete those files to avoid failures on static builds.

Some Expat reverse dependencies fail to be build, but those also fail on the MXE master branch (see "GDAL Dependency Reverse Dependency Section").

## HDF5

## netcdf

## Armadillo

Specify HDF5 include directory directly.

----

# GDAL

I switched the build to CMake as other build platforms become unsupported with 3.6.0.  The CMake build auto-detects available packages and enables features accordingly.

Most of the patches are fixing the package auto-detection which does not work well for non-CMake packages, or for CMake packages that do not correctly EXPORT themselves.  Most of the fixes involve hacking the detection code to use the pkgconfig files and generate the corresponding `target_link_libraries` directives.

I removed a setting of __MSVCRT_VERSION__ no longer needed.

----

# Other

See #2918 for discussion of GDAL revdeps and GDAL dependency revdeps.

---

Thanks for your work building and keeping MXE up and running.


brodieG:
Fyi, GDAL has recalled 3.6.0 due to some [major bugs](https://github.com/OSGeo/gdal/blob/v3.6.1/NEWS.md) that are not properly fixed until 3.6.2.  Unfortunately the upgrade to 3.6.2 is not completely trivial due to some changes in how armadillo is used.  I'm going to close this PR until I have a chance to modify it to target 3.6.2 instead of 3.6.0, and re-open at that point, hopefully in a week or so.

PR #2948, opened 2023-01-22
Upgrade GDAL to 3.6.2

brodieG:
# Background

This replaces #2943 because gdal recalled 3.6.0 due to [major bugs](https://github.com/OSGeo/gdal/blob/v3.6.1/NEWS.md).

At @mabrand's request I split as much of the original PR (#2918) into separate PRs as I could.  The following PRs must be applied before this one:

* #2935 
* #2941 
* ~~#2942~~ (seemingly obviated as of 4/2023) 
* #2945 
* #2946 

The remaining changes in this PR need to be a applied together as all the packages involved are very out of date.

# Testing

I tested building on all permutations of x86/i686 and static/shared.  Additionally I tested some of the x86 binaries (`gdalinfo.exe`, `gdaltranform.exe`) both static and shared on a windows machine (these are very basic tests that the binaries run).

I also tested builds of revdeps of the updated packages (see #2943, #2918 and other linked PRs for commentary).

# Comments on Patches

## GDAL

CMake is required to build >= 3.6.0.

Most of the patches are fixing the package auto-detection which does not work well for non-CMake packages, or for CMake packages that do not correctly EXPORT themselves. Most of the fixes involve hacking the detection code to use the pkgconfig files and generate the corresponding target_link_libraries directives.

I removed a setting of MSVCRT_VERSION no longer needed.

## HDF5

## netcdf

# Other

Thanks for your work building and keeping MXE up and running.  Let me know if there is anything else I can do to make this merge easier.  I realize it's very large and a lot of work for you to review.  Unfortunately, the beastly nature of gdal and the very long time span since it was last updated make it a hard problem.


brodieG:
@mabrand anything else I can do to facilitate the review of this PR?

brodieG:
@mabrand @tonytheodore ping?  I realize this creates no obligation for you whatsoever, but it was quite the task to get from the 5 year old version currently at MXE (2.2.4) to this one, particularly for something as dependency laden as GDAL.  It would really be a shame to have all this go to waste.  I know both of you have shown some interest in this in the past, and I thank you for the review and merging of several of the component PRs you have done already.

kalibera:
I would be grateful if this could be incorporated. We are using the changes in Rtools43 (for the upcoming release of R 4.3) and for us smaller the differences from upstream MXE the better, that makes the maintenance easier. I believe the benefit also works the other way: MXE (the subset used by R) gets additional testing via R packages and the R community. Particularly the geospatial packages are challenging to maintain, as they have frequent releases and frequent changes, they are also the biggest source of differences between Rtools and upstream MXE. Thanks!

mabrand:
Would like to merge this very soon. Just a few prerequisites remaining such as https://github.com/mxe/mxe/pull/2935.

brodieG:
Thanks for looking at this again.  Note #2942 does not appear to be needed anymore.

mabrand:
The other commits are merged, but now this one has a merge conflict. Could you take a look at this?

brodieG:
Ah yes when I tweaked the src/gdal.mk in #2935 I missed that I introduced a merge conflict.  Looks like this is good to go now.  I tested the build on x64 shared/static and it built.

PR #2957, opened 2023-02-16
qt6: add openssl and postgresql

### LINKED ISSUE #2950: please enable openssl for qt6-base
The Makefile for Qt5 base builds Qt wit support for openssl (--openssl-linked) but in Qt6 openssl is disabled.
Can you enable it?

sabelka:
Enable openssl and the PostgreSQL QSql plugin for qt6, like it is also in qt5
fixes #2950


mabrand:
Qt6 supports Window's native schannel. Is there a strong case for building with openssl support too, given the extra dependencies?

sabelka:
Actually, I wasn't aware of Qt's schannel support on Windows. I just noticed, that QNetworkManager was not able do establish https connections, and enabling openssl for Qt fixed my problem. But I quess schannel would be fine, too.
However, it looks like it is not enabled by default (or I did something wrong in my test).
So, I'd be fine with any ssl solution.

mabrand:
I think schannel is enabled. At least it works for me.

If we're going to enable the postgresql driver, that's going to drag in openssl anyway. Hmm.
Maybe you can do some checks with building support for both schannel and openssl and choosing the desired provider at runtime, just to make sure it works.

sabelka:
I'll try and let you know!

stefanha:
I've been using Qt6 QNetworkRequest successfully for HTTPS without explicitly enabling OpenSSL.

sabelka:
Yes SSL works also without OpenSSL! (I had a mistake in my packager script, so the schannel plugin was in the wrong directory). So you are absolutely right, we don't need it on Windows.
So, I've modified this PR to include PostgreSQL only.

sabelka:
> Maybe you can do some checks with building support for both schannel and openssl and choosing the
> desired provider at runtime, just to make sure it works.
It is possible to build both plugins, but I think it is not easily possible to choose the back end at runtime.
However, if you package just one of the two plugins with your application, Qt will use whatever it gets ;-)

fiesh:
This adds a very heavy dependency to Qt permanently... I wouldn't want this for our build images, for example.

ikocs:
Hi, I want to use your makefile to build a driver version. Can you tell me if I need to install additional packages before installing qt via mxe?

sabelka:
@ikocs, the dependency on postgres is included in this pull request, so PostgreSQL and its dependencies will be automatically downloaded and built. No need to install any extra packages for the psql plugin.
PS: I have rebased this pull request to the current master.

ikocs:
Thanks for a job well done!

fiesh:
Gentlemen, this is an anti-feature for a lot of users and therefore at least of debatable value.  Having this as a plugin would be a better alternative since it wouldn't force everybody to build such a heavy optional dependency.

ikocs:
@fiesh I'm new to mxe, and haven't written my own plugins yet (and don't know how to do it right). But it sounds logical. Do I understand correctly that the plugin documentation is [here](https://github.com/mxe/mxe/blob/master/plugins/README.md)? 

But I tried building a docker container from the current PR. It turned out to be only 100mb bigger than my qt container I made before. 

I'm fine with it at the moment, but I'll still try to figure out the plugins.

Issue #2968, opened 2023-04-02
Move poppler-qt5 and poppler-qt6 into their own packages


darealshinji:
By default building poppler will build qt5 and qt6, which is a lot. I think it would be better if poppler-qt5 and poppler-qt6 were separate packages so you don't have to build both together.

mabrand:
moved to https://github.com/mxe/mxe/pull/2973

PR #2973, opened 2023-04-10
poppler: move poppler-qt libraries into their own packages

darealshinji:
Moving the poppler-qt libraries into their own packages so people don't have to build Qt5 and Qt6 if they just want poppler.
See my request https://github.com/mxe/mxe/issues/2968

Builds on all 4 targets for me.

### DIFF
--- a/src/poppler-qt5.mk
+++ b/src/poppler-qt5.mk
@@ -1,0 +1,23 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := poppler-qt5
+$(PKG)_WEBSITE   = $(poppler_WEBSITE)
+$(PKG)_IGNORE    = $(poppler_IGNORE)
+$(PKG)_VERSION   = $(poppler_VERSION)
+$(PKG)_CHECKSUM  = $(poppler_CHECKSUM)
+$(PKG)_SUBDIR    = $(poppler_SUBDIR)
+$(PKG)_FILE      = $(poppler_FILE)
+$(PKG)_URL       = $(poppler_URL)
+$(PKG)_DEPS     := cc poppler qtbase
+
+define $(PKG)_BUILD
+    $(subst @build_with_cpp@,OFF, \
+    $(subst @build_with_glib@,OFF, \
+    $(subst @build_with_qt5@,ON, \
+    $(subst @build_with_qt6@,OFF, \
+    $(poppler_BUILD_COMMON)))))
+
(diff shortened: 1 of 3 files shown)


src/poppler-qt5.mk
```diff
@@ -0,0 +1,23 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := poppler-qt5
+$(PKG)_WEBSITE  := $(poppler_WEBSITE)
+$(PKG)_IGNORE   := $(poppler_IGNORE)
+$(PKG)_VERSION  := $(poppler_VERSION)
+$(PKG)_CHECKSUM := $(poppler_CHECKSUM)
+$(PKG)_SUBDIR   := $(poppler_SUBDIR)
+$(PKG)_FILE     := $(poppler_FILE)
+$(PKG)_URL      := $(poppler_URL)
```
tonytheodore, line 10: I think all these need to be expanded variables i.e `=` instead of `:=`.

darealshinji: Done.

mabrand:
I'm not sure I understand completely, but I wonder if these won't just clobber each other when they are installed. Is the intention that either one or both of poppler-qt5 and poppler-qt6 can be installed?

darealshinji:
It should be fine to install them sizde by side, but I will check again to be sure.

darealshinji:
I just check it. There are no conflicts and no clobbering. I used `ls -lR usr/x86_64-w64-mingw32.static/` to check for new files as well as timestamp changes.
These are the files/directories that will be installed:

poppler:
```
bin/test-poppler.exe
include/poppler/
installed/poppler
lib/libpoppler.a
lib/libpoppler-cpp.a
lib/libpoppler-glib.a
lib/pkgconfig/libpoppler.pc
lib/pkgconfig/libpoppler-cpp.pc
lib/pkgconfig/libpoppler-glib.pc
```

poppler-qt5:
```
include/poppler/qt5/
installed/poppler-qt5
lib/libpoppler-qt5.a
lib/pkgconfig/libpoppler-qt5.pc
```

poppler-qt6:
```
include/poppler/qt6/
installed/poppler-qt6
lib/libpoppler-qt6.a
lib/pkgconfig/libpoppler-qt6.pc
```

`include/poppler/qt{5,6}/` is not installed by the poppler package.

mabrand:
Thanks. 
We still need to follow up on the effect qt6-qtimageformats. poppler-qt6 should list it as a dependency. However, at least for static builds, it breaks the build of poppler-qt6. Are you able to look into this?

PR #2975, opened 2023-04-23
qt6-qtimageformats: fix issues with static targets

darealshinji:
This will fix the builds of static poppler-qt6 mentioned here: https://github.com/mxe/mxe/pull/2973#issuecomment-1517668761


mabrand:
This looks like it should work, but I wonder if you would be willing to dig into the Qt build system to try to fix the underlying problem. Sed hacks can be fragile. Also, I suspect that qtmultimedia is also affected by this problem.

darealshinji:
I've tried to fix the build files with patches but I couldn't figure out which source files. I might try it again though.

darealshinji:
I can manually set the prefix in the cmake files to `Qt6::` with patches but it still chokes on the automatically added library names `Qt::PlatformPluginInternal` and `Qt::PlatformModuleInternal`. There seems to be something missing that automatically aliases the namespace Qt:: to Qt6:: and I don't know enough about cmake to fix this.

darealshinji:
I found a fix that involves patching the poppler sources.

PR #2986, opened 2023-05-23
glib: update 2.70.2 -> 2.76.3

Jan200101:
existing patches were rebased
new patch was added to fix libintl detection

Thus far I have only build gtk3 against it, which needed `-Werror` removed, since it makes use of deprecated functionality.

This would be required for any future efforts of adding GTK4 to MXE.


Jan200101:
upstream PR for the libintl fix https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3448

Talked to the person who initially made the change and it appears that it was left out because it didn't appear used.

Jan200101:
rease because I noticed with a clean recompile that glib replaced PCRE with PCRE2 since 2.73.2 (relevant merge: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2529)

PR #2987, opened 2023-05-25
qt6: Add qtwebsockets Qt module

svenihoney:
Add Qt Websockets module (https://doc.qt.io/qt-6/qtwebsockets-index.html) to mxe.

Used the existing qt6-qtsvg and simply changed PKG and CHECKSUM. Compiles locally.


tonytheodore:
Thanks!

PR #2988, opened 2023-05-27
graphene: add package

Jan200101:
Adds graphene package

compiles against glib 2.70.2 and 2.76.3

Tests were explicitly disabled because they depend on `mutest-1`.

No test source was included with this package because it is quite a complicated and vast project.

This would be required for any future efforts of adding GTK4 to MXE.


mrwormhole:
@mabrand sorry to bother but can this be merged so that gtk4 can also be added?

mabrand:
Merged. It builds. Please try and test it.

PR #2989, opened 2023-05-27
cairo: declare dependency on lzo2

Jan200101:
cairo-script depends on `lzo2` but never defines its dependency in any of its pkg-config files.

Future cairo releases use meson, where this is no longer an issue.

Also adds another patch to remove the gtkdocize detection in the autogen script, which prevents the project file from being remade.


mabrand:
cairo has been updated in the meantime and uses meson. Please reopen if this is still relevant.

Jan200101:
Rechecked if what I initially claimed is correct
this is no longer an issue with meson

PR #2985, opened 2023-05-22
sdl2: update to 2.26.2

Jan200101:
fixes #2983

The patch was rebased against 2.26.2 and, at least as far as I can tell, does what its suppose to.

I've build a few packages that depend on it, but haven't found time to test against any projects.

PR #2990, opened 2023-05-27
gtk4: add package

Jan200101:
depends on #2989 #2986 #2988 #2985 

very much WIP, but is at least capable of compiling

### DIFF
--- a/src/gtk4.mk
+++ b/src/gtk4.mk
@@ -1,0 +1,26 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := gtk4
+$(PKG)_WEBSITE  := https://gtk.org/
+$(PKG)_DESCR    := GTK4
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 4.11.2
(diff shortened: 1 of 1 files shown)


mrwormhole:
hello @Jan200101 sorry to bother, but what is the state of adding gtk4 to this mxe? thanks

mrwormhole:
I have continued this effort on my fork so currently meson has a bug and can not detect vulkan with pkg-config when vulkan-headers are installed

https://github.com/mrwormhole/mxe/pull/1

jd-boyd:
This says it depends on #2989,, and #2989 says that it will be a non-issue when MXE is updated to use a cairo that uses meson, which has already happened.  I think that means this PR no longer depends on #2989.

This also notes that it depends on #2986, and #2986 has a note that it was cherry-picked in https://github.com/mxe/mxe/commit/2774d22f4b317bc92df6db09d9891096fb52d3c3

It looks like #2988, the graphene dependency is still unmerged, so this may still depend on that PR.

The final one listed as this depending on it is #2985, which has been merged.

I guess I will not try actually building the #2985 and this PR.

mabrand:
Merged. It builds. Please try and test it.

PR #3002, opened 2023-07-29
add biosig, sigviewer, stimfit, edfbrowser and prerequisites

schloegl:

  * install .pc file,

Thanks!


mabrand:
Hi,
Are you interested in rebasing this on MXE master? Should we try to get it merged?

src/wxwidgets.mk
```diff
@@ -7,8 +7,10 @@ $(PKG)_IGNORE   :=
 $(PKG)_VERSION  := 3.2.7
 $(PKG)_CHECKSUM := 69a1722f874d91cd1c9e742b72df49e0fab02890782cf794791c3104cee868c6
 $(PKG)_GH_CONF  := wxWidgets/wxWidgets/releases/latest,v,,,,.tar.bz2
+$(PKG)_URL      := https://github.com/wxWidgets/wxWidgets/releases/download/v$($(PKG)_VERSION)/wxWidgets-($(PKG)_VERSION).tar.bz2
```
mabrand, line 10: ?

src/wxwidgets.mk
```diff
@@ -7,8 +7,10 @@ $(PKG)_IGNORE   :=
...
 $(PKG)_DEPS     := cc expat jpeg libiconv libpng sdl tiff zlib
 
+
```
mabrand, line 13: ?

PR #3003, opened 2023-08-05
Add package cyrus-sasl

be1:
Heavily based on Starius's unmaintained commit #2f2ba1b
https://github.com/mxe/mxe/commit/2f2ba1ba42b893d27fbf93a7288e6f983e0423c1

* I understand the given patch does not follow the guidelines, but maybe one can tell me how to generate a good one.
* thanks for reviewing and telling me other things that have to be done

  * install .pc file,

Thanks!


be1:
FYI, this package doesn't build `saslauthd`, but only sasl libraries. This is useful for **client** programs that want to use `cyrus_sasl` to authenticate.

PR #3007, opened 2023-08-23
This fixes #3006

### LINKED ISSUE #3006: zlib maintainers dgaf about non-bleeding-edge versions
As of today, it seems `zlib` is no longer supplying their version `1.2.13` in favor of `1.3`, which was released 4 days ago.  This means building MXE's `zlib` package which is still on `1.2.13` fails as all downloads 404 out.

I'd therefore request setting

```
$(PKG)_URL_2    := https://github.com/madler/zlib/releases/download/v$($(PKG)_VERSION)/$($(PKG)_FILE)
```

in `src/zlib.mk`.  At least this made things work again for us.

Or maybe allow setting `_URL_n` for any value of `n`, so that arbitrarily many fallback servers are supported, and make this `_URL_3`?

fatduckling:
> As of today, it seems zlib is no longer supplying their version 1.2.13 in favor of 1.3, which was released 4 days ago. This means building MXE's zlib package which is still on 1.2.13 fails as all downloads 404 out.


tonytheodore:
See https://github.com/mxe/mxe/commit/31c616929691f78af276ff39cbe787bc8332baef, `GH_CONF` simplifies the handling of downloads, updates, etc.

PR #3010, opened 2023-09-12
qt6: Add qthttpserver Qt module

casept:
Add Qt HTTP server module (https://doc.qt.io/qt-6/qthttpserver-index.html) to mxe.

Used the existing qt6-qtsvg and simply changed PKG and CHECKSUM. Compiles in a clean docker container for `x86_64-w64-mingw32.static`.

Once PR #2987 is merged, this module should also depend on QtWebsockets as some optional features require it.


tonytheodore:
Thanks!

PR #3013, opened 2023-09-17
Update libssh2.mk

jeroen:
Update libssh2 and get rid of the annoying gpg dependency.

### DIFF
--- a/src/libssh2.mk
+++ b/src/libssh2.mk
@@ -3,12 +3,12 @@
 PKG             := libssh2
 $(PKG)_WEBSITE  := https://libssh2.org/
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 1.10.0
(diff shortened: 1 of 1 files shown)


src/libssh2.mk
```diff
@@ -21,7 +21,8 @@ define $(PKG)_BUILD
     cd '$(1)' && ./configure \
         $(MXE_CONFIGURE_OPTS) \
         --disable-examples-build \
-        --with-crypto=libgcrypt \
+        --disable-rpath \
+        --with-crypto=wincng \
         LIBS="`$(PREFIX)/$(TARGET)/bin/libgcrypt-config --libs`" \
```
tonytheodore, line 26: This `libgcrypt-config` line should no longer be required.

jeroen: Thanks, I'll remove it.

tonytheodore:
Thanks!

PR #3014, opened 2023-09-19
libftdi1: update 1.4 -> 1.5

colinbourassa:


colinbourassa:
It seems libftdi is building the static libs when invoked in the shared lib environment. I'll update this PR once I've fixed this.

colinbourassa:
Commit amended -- ready for review.

PR #3015, opened 2023-09-21
Fix static build.

kalibera:
This is needed for a static build, otherwise one runs into linking errors such as "undefined reference to `__imp_pkgconf_strlcpy'."

### DIFF
--- a/plugins/examples/host-toolchain/pkgconf-host.mk
+++ b/plugins/examples/host-toolchain/pkgconf-host.mk
@@ -18,7 +18,7 @@ define $(PKG)_BUILD
     cd '$(SOURCE_DIR)' && ./autogen.sh
     cd '$(BUILD_DIR)' && '$(SOURCE_DIR)/configure' \
         $(MXE_CONFIGURE_OPTS) \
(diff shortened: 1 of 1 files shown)


plugins/examples/host-toolchain/pkgconf-host.mk
```diff
@@ -17,7 +17,8 @@ endef
 define $(PKG)_BUILD
     cd '$(SOURCE_DIR)' && ./autogen.sh
     cd '$(BUILD_DIR)' && '$(SOURCE_DIR)/configure' \
-        $(MXE_CONFIGURE_OPTS)
+        $(MXE_CONFIGURE_OPTS) \
+        CFLAGS="$(CFLAGS) $(if $(BUILD_STATIC),-DPKGCONFIG_IS_STATIC)"
```
tonytheodore, line 21: I think `$(CFLAGS)` is a `make` variable in this context and likely to be empty

kalibera: Thanks, then perhaps it is not needed even in cairo.mk?
https://github.com/mxe/mxe/commit/c950f3635caa9b2dd562286baa51138e6e30dca1

tonytheodore:
Thanks.

PR #3020, opened 2023-09-22
Update packages.

kalibera:
This is a simple update of several packages, where it does not require any further adaptations. Tested by building full MXE for x86_64-w64-mingw32.static before and after the changes: no packages newly failing to build with these updates.

The set is biased to what is used in Rtools - there may be other packages that are easy to update.


mabrand:
Cherry-picked into master as 4cceab40557e2d4c70e4105416efe9a209059c92. Thanks.

Issue #3012, opened 2023-09-16
djvulibre failed to compile under Linux Ubuntu 23.04


cgilles:
Console trace : 

```
[download]    djvulibre-3.5.28.tar.gz
[build]       djvulibre              x86_64-w64-mingw32.shared

Failed to build package djvulibre for target x86_64-w64-mingw32.shared!
------------------------------------------------------------
  631 |   register int z=*data++;
      |                ^
make[2]: Leaving directory '/mnt/data/build.win64/tmp-djvulibre-x86_64-w64-mingw32.shared/djvulibre-3.5.28.build_/libdjvu'
Makefile:884: recipe for target 'build-only-djvulibre_x86_64-w64-mingw32.shared' failed
make[1]: *** [build-only-djvulibre_x86_64-w64-mingw32.shared] Error 2
make[1]: Leaving directory '/mnt/data/build.win64'
real	0m35,599s
user	1m32,299s
sys	0m14,991s
------------------------------------------------------------
[log]      /mnt/data/build.win64/log/djvulibre_x86_64-w64-mingw32.shared

Makefile:884: recipe for target '/mnt/data/build.win64/usr/x86_64-w64-mingw32.shared/installed/djvulibre' failed
make: *** [/mnt/data/build.win64/usr/x86_64-w64-mingw32.shared/installed/djvulibre] Error 1
```

Log trace : 

```
make[1]: Entering directory '/mnt/data/build.win64'
uname -a
Linux gilles-MXE-Build 5.4.0-131-generic #147~18.04.1-Ubuntu SMP Sat Oct 15 13:10:18 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
git log --pretty=tformat:"%H - %s [%ar] [%d]" -1
b3101cc659170b08efef83dcbebfe4806c2feab3 - Update packages.json & build-matrix.html [30 hours ago] [ (HEAD -> master, origin/master, origin/HEAD)]
lsb_release -a 2>/dev/null || sw_vers 2>/dev/null || true
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.6 LTS
Release:	18.04
Codename:	bionic
autoconf --version 2>/dev/null | head -1
autoconf (GNU Autoconf) 2.69
automake --version 2>/dev/null | head -1
automake (GNU automake) 1.15.1
gcc --version
gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
[...]

PR #3028, opened 2023-10-23
djvulibre: fix build failure

darealshinji:
Also...
- remove dependency on "tiff"; it's not used despite the configure script looking for it
- better way to build static library
- don't export all DLL symbols
- enable pthreads

This will resolve https://github.com/mxe/mxe/issues/3012

PS: upstream discussion about MinGW build compatibility -> https://sourceforge.net/p/djvu/bugs/354/


darealshinji:
I've fixed the DLL import/export issues by using `__attribute__((visibility("default")))`

Issue #2617, opened 2021-02-10
Could not find section for RVA happened when trying to find the DLL dependency


wentongwu:
I'm using copydlldeps.py to find the DLL dependency, but got the error 
```
Traceback (most recent call last):
  File "copydlldeps.py", line 155, in <module>
    for dependency_dll in get_imports(pe_filename_full):
  File "copydlldeps.py", line 84, in get_imports
    seek_to_rva(rva)
  File "copydlldeps.py", line 69, in seek_to_rva
    raise ValueError('Could not find section for RVA.')
ValueError: Could not find section for RVA.
```
whether using the exe file built by myself or dll file built by the source.
anyone has the same problem? Appreciate any help, thanks a lot!

sebastienraillet:
Hi @wentongwu,
I'm facing the same issue. Did you finally find a solution ?
Thanks 😃

PR #3030, opened 2023-10-26
Adds PE32+ support to copydlldeps.py

NiklasWitzel:
This PR references #589 and #2617. Both are closed, but were not fixed in copydlldeps.py

The issue is with 64-bit files (PE32+) files, because the field for `NumberOfRvaAndSizes` is in another location compared to PE32 files.

Reference (Magic): https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#optional-header-image-only
Reference (NumberOfRvaAndSizes): https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#optional-header-windows-specific-fields-image-only

### DIFF
--- a/tools/copydlldeps.py
+++ b/tools/copydlldeps.py
@@ -41,10 +41,20 @@ def get_imports(file):
     f.seek(60)
     pe_header_offset = struct.unpack('<L', f.read(4))[0]
 
+    # Is it PE32 or PE32+ ?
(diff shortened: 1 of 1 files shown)


tz-reinberger:
Bumping this because I stumbled upon the issue today.
Tried the script from the pull request and it works flawlessly.
Can we please get this merged?

PR #2964, opened 2023-03-23
Qt6: add QtSCXML

Petzep:
Added the QtSCXML module which adds:
- statemachine and qtscxml

PR #3036, opened 2023-11-28
Add qt6-qtscxml

fiesh:
This supersedes #2964, comes in a single commit, and works with the current Qt6 version of MXE (6.6.0).


fiesh:
So again the checksum needs to be updated, but I won't go ahead any more until there's actually a chance this gets merged...

martinwag:
+1

PR #3063, opened 2024-04-25
Fix build failure for pakage `gtk2`

jakint0sh:
Currently, package `gtk2` target `i686-w64-mingw32.static` fails to build.  This is due to linker errors while trying to run the test compilation in `src/gtk2.mk`.  For some reason or other, the linker is not informed that it needs to link against `libstdc++`, which causes a lot of `undefined reference` errors when trying to link `usr/i686-w64-mingw32.static/lib/libcairo.a`.  Simply adding `-lstdc++` fixes this problem and allows GTK2 to compile sucessfully.

### DIFF
--- a/src/gtk2.mk
+++ b/src/gtk2.mk
@@ -39,7 +39,7 @@ define $(PKG)_BUILD
     '$(TARGET)-gcc' \
         -W -Wall -Werror -Wno-error=deprecated-declarations -ansi \
         '$(TEST_FILE)' -o '$(PREFIX)/$(TARGET)/bin/test-gtk2.exe' \
-        `'$(TARGET)-pkg-config' gtk+-2.0 gmodule-2.0 --cflags --libs`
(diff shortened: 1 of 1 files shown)


mabrand:
Looks like the same issue fixed by ef557517277867b1d66c5efca3e424e89667a2b8. libcairo.a introduces C++ dependencies. There I fixed it by using g++ instead of gcc to compile and link the test program. Not sure which way is better. What do you think?

jakint0sh:
Yep, it would seem that this is the same failure as in that.  Just for consistency's sake, I think that switching to g++ would be best, but I'll defer to your judgement on that, as you've been working on this project a lot longer than me.  I'm just some random who noticed how many packages fail to build and got annoyed, and decided to spend an unreasonable amount of time trying to fix it.

I can set up the switch to g++ in this pull request if you'd like (assuming you'd like to go that route).

mabrand:
Could you change this one to g++ then, just for the sake of consistency?

jakint0sh:
All done, and tested to make sure it works.

Issue #3104, opened 2024-09-26
add missing -fpermissive


e2002e:
agg needs a compiler flag.

```
define $(PKG)_BUILD
    $(foreach f,authors news readme, mv '$(1)/$f' '$(1)/$f_';mv '$(1)/$f_' '$(1)/$(call uc,$f)';)
    cd '$(1)' && autoreconf -fi -I $(PREFIX)/$(TARGET)/share/aclocal
    cd '$(1)' && CFLAGS="-fpermissive" CXXFLAGS="-fpermissive" ./configure \
        $(MXE_CONFIGURE_OPTS) \
        --without-x
    $(MAKE) -C '$(1)' -j '$(JOBS)' install bin_PROGRAMS= sbin_PROGRAMS= noinst_PROGRAMS=
endef

```

I'm leaving this opened so that you can see it, there are a couple other solved issues that I closed.

mabrand:
Would you like to create a pull request for the patch?

PR #3120, opened 2024-11-03
fix compile errors

e2002e:
This should do for issue #3104 and the other closed one, but I couldn't test building if the opened python error is still present.


fiesh:
`-fpermissive` is typically not a flag one wants... not sure if this is a good addition.

Issue #3110, opened 2024-10-10
SDL_mixer 1.2.12 fails to build (patch included)


darealshinji:
```
libtool: compile:  x86_64-w64-mingw32.static-gcc -g -O2 -D_GNU_SOURCE=1 -I/home/djcj/opt/mxe/usr/x86_64-w64-mingw32.static/include/SDL -D_GNU_SOURCE=1 -Dmain=SDL_main -fvisibility=hidden -DWAV_MUSIC -DMODPLUG_MUSIC -I/home/djcj/opt/mxe/usr/x86_64-w64-mingw32.static/include/libmodplug -DMID_MUSIC -DUSE_TIMIDITY_MIDI -I/home/djcj/opt/mxe/tmp-sdl_mixer-x86_64-w64-mingw32.static/SDL_mixer-1.2.12/timidity -DUSE_NATIVE_MIDI -I/home/djcj/opt/mxe/tmp-sdl_mixer-x86_64-w64-mingw32.static/SDL_mixer-1.2.12/native_midi -DOGG_MUSIC -DHAVE_SIGNAL_H -DHAVE_SETBUF -DMP3_MUSIC -I/home/djcj/opt/mxe/usr/x86_64-w64-mingw32.static/include -I/home/djcj/opt/mxe/usr/x86_64-w64-mingw32.static/include/smpeg -I/home/djcj/opt/mxe/usr/x86_64-w64-mingw32.static/include/SDL -D_GNU_SOURCE=1 -Dmain=SDL_main -MMD -MT build/dynamic_ogg.lo -c /home/djcj/opt/mxe/tmp-sdl_mixer-x86_64-w64-mingw32.static/SDL_mixer-1.2.12/dynamic_ogg.c -o build/dynamic_ogg.o
../SDL_mixer-1.2.12/dynamic_ogg.c: In function 'Mix_InitOgg':
../SDL_mixer-1.2.12/dynamic_ogg.c:111:42: error: assignment to 'int (*)(void *, OggVorbis_File *, char *, long int,  ov_callbacks)' from incompatible pointer type 'int (*)(void *, OggVorbis_File *, const char *, long int,  ov_callbacks)' [-Wincompatible-pointer-types]
  111 |                 vorbis.ov_open_callbacks = ov_open_callbacks;
      |                                          ^
```

[full log](https://gist.github.com/darealshinji/7d58ebe75089f4ba9758b2e11587d305)

[...]

PR #3170, opened 2024-12-08
SDL_mixer: fix package fails to build

darealshinji:
argument 3 of ov_open_callbacks in dynamic_ogg.[c|h] must be const char* otherwise the build fails

Tested on x86_64-w64-mingw32.static and i686-w64-mingw32.shared.

Closes https://github.com/mxe/mxe/issues/3110


mabrand:
Does 752c328e91f7d17060c5787619940a7d9ff673f2 fix the problem for you?

darealshinji:
> Does [752c328](https://github.com/mxe/mxe/commit/752c328e91f7d17060c5787619940a7d9ff673f2) fix the problem for you?

yes

Issue #3111, opened 2024-10-10
SDL_image 1.2.12 fails to build (patch included)


darealshinji:
```
/bin/bash ./libtool --tag=CC   --mode=compile x86_64-w64-mingw32.static-gcc -DPACKAGE_NAME=\"\" -DPACKAGE_TARNAME=\"\" -DPACKAGE_VERSION=\"\" -DPACKAGE_STRING=\"\" -DPACKAGE_BUGREPORT=\"\" -DSTDC_HEADERS=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_MEMORY_H=1 -DHAVE_STRINGS_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_UNISTD_H=1 -DLT_OBJDIR=\".libs/\" -DPACKAGE=\"SDL_image\" -DVERSION=\"1.2.12\" -DLOAD_JPG=1 -DLOAD_PNG=1 -DLOAD_WEBP=1 -DLOAD_BMP=1 -DLOAD_GIF=1 -DLOAD_LBM=1 -DLOAD_PCX=1 -DLOAD_PNM=1 -DLOAD_TGA=1 -DLOAD_XCF=1 -DLOAD_XPM=1 -DLOAD_XV=1 -I. -I/home/djcj/opt/mxe/tmp-sdl_image-x86_64-w64-mingw32.static/SDL_image-1.2.12     -I/home/djcj/opt/mxe/usr/x86_64-w64-mingw32.static/include/webp   -I/home/djcj/opt/mxe/usr/x86_64-w64-mingw32.static/include/libpng16   -g -O2 -I/home/djcj/opt/mxe/usr/x86_64-w64-mingw32.static/include/SDL -D_GNU_SOURCE=1 -Dmain=SDL_main -MT IMG_pnm.lo -MD -MP -MF .deps/IMG_pnm.Tpo -c -o IMG_pnm.lo /home/djcj/opt/mxe/tmp-sdl_image-x86_64-w64-mingw32.static/SDL_image-1.2.12/IMG_pnm.c
../SDL_image-1.2.12/IMG_png.c: In function 'IMG_InitPNG':
[...]

PR #3171, opened 2024-12-08
SDL_image: fix package fail to build

darealshinji:
Function arguments had different types. const keywords were either missing or incorrectly placed, causing the build to fail. webp_get_features_internal must return VP8StatusCode.

Tested on x86_64-w64-mingw32.static and i686-w64-mingw32.shared.

Closes https://github.com/mxe/mxe/issues/3111

### DIFF
--- a/src/sdl_image-1-fixes.patch
+++ b/src/sdl_image-1-fixes.patch
@@ -1,0 +1,288 @@
+This file is part of MXE. See LICENSE.md for licensing information.
+
+Contains ad hoc patches for cross building.
+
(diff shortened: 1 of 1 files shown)


mabrand:
Does 752c328e91f7d17060c5787619940a7d9ff673f2 fix the problem for you?

darealshinji:
> Does [752c328](https://github.com/mxe/mxe/commit/752c328e91f7d17060c5787619940a7d9ff673f2) fix the problem for you?

yes

PR #3179, opened 2024-12-31
hyperscan: fix build on arm64 host

starius:
Fix error: "error: narrowing conversion of '161' from 'int' to 'char' [-Wnarrowing]" in generated files src/parser/control_verbs.cpp and util/ExpressionParser.cpp.

hyperscan: update from 5.4.0 to 5.4.2.

hyperscan: update website


tonytheodore:
Thanks!

PR #3211, opened 2025-03-25
Add QuantLib.

kalibera:
Add package QuantLib.

### DIFF
--- a/src/quantlib.mk
+++ b/src/quantlib.mk
@@ -1,0 +1,32 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := quantlib
+$(PKG)_WEBSITE  := https://www.quantlib.org/
+$(PKG)_DESCR    := QuantLib
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 1.37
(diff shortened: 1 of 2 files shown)


src/quantlib.mk
```diff
@@ -0,0 +1,38 @@
...
+
+    # install
+    $(MAKE) -C '$(BUILD_DIR)' -j '$(JOBS)' VERBOSE=1
+    $(MAKE) -C '$(BUILD_DIR)' -j 1 install VERBOSE=1
+
+    # fix pkg-config file
+    $(SED) -i 's/^\(Libs:.*\)/\1 -fopenmp/g' \
+      '$(PREFIX)/$(TARGET)/lib/pkgconfig/$(PKG).pc'
+    $(SED) -i 's/^\(Cflags:.*\)/\1 -fopenmp/g' \
+      '$(PREFIX)/$(TARGET)/lib/pkgconfig/$(PKG).pc'
```
mabrand, line 31: Why not do this with a patch instead of sed hack?

kalibera: I've turned these into a patch and I've also offered the patch upstream. It seems the error was in the cmake file generating the pkg-config file

PR #3213, opened 2025-03-25
Update armadillo. Avoid absolute paths in cmake files.

kalibera:
Update armadillo. Avoid absolute paths in cmake files to allow relocation (so that e.g. the cmake files can be used natively on Windows). Avoid unnecessary '-lgomp' in the test linking (there is already -fopenmp, which is more portable, it works also with LLVM).


src/armadillo.mk
```diff
@@ -36,9 +36,21 @@ define $(PKG)_BUILD
...
+                 '$(PREFIX)/$(TARGET)/share/Armadillo/CMake/ArmadilloLibraryDepends.cmake'
+
+    # fix cmake file, avoid unnecessary absolute path
+    $(SED) -i -e 's!\(ARMADILLO_LIBRARY_DIRS[ \t]\+\)[^ \t()]\+\(.*\)!\1""\2!g' \
+                 '$(PREFIX)/$(TARGET)/share/Armadillo/CMake/ArmadilloConfig.cmake'
+
+    # fix cmake file, avoid unnecessary absolute path
+    $(SED) -i -e 's!\(ARMADILLO_INCLUDE_DIRS[ \t]\+\)[^ \t()]\+\(.*\)!\1""\2!g' \
+                 '$(PREFIX)/$(TARGET)/share/Armadillo/CMake/ArmadilloConfig.cmake'
+
```
mabrand, line 50: Wouldn't it be better to do this with a patch than this sed hack?

kalibera: I've removed this part from the patch. I agree it would be nice to fix that properly upstream, so that the cmake files are generated correctly - I've opened an issue upstream. I don't know enough cmake (nor Armadillo) to be comfortable writing a patch for that.

src/armadillo.mk
```diff
@@ -40,5 +40,5 @@ define $(PKG)_BUILD
     '$(TARGET)-g++' \
         -W -Wall -Werror -fopenmp \
         '$(TEST_FILE)' -o '$(PREFIX)/$(TARGET)/bin/test-$(PKG).exe' \
-        `'$(TARGET)-pkg-config' $(PKG) --cflags --libs` -lgomp
+        `'$(TARGET)-pkg-config' $(PKG) --cflags --libs`
```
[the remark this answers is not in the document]
kalibera: -fopenmp implies also linking of the OpenMP runtime (https://gcc.gnu.org/onlinedocs/libgomp/Enabling-OpenMP.html#Enabling-OpenMP). This is my experience also with LLVM. I want to remove -lgomp so that the test also works with LLVM.

PR #3216, opened 2025-04-14
adding new libbotan2 package

hax0rbana-adam:
This is my first time making a package for MXE, so please let me know if I am not following best practices. I have read http://mxe.cc/#creating-packages and borrowed heavily from existing packages.

I verified the .a and .pc files are installed, and was able to create a $(PKG)_UPDATE section.

Once this package is acceptable to merge in, I would like to make a .deb package for it so I can host it on my apt repo. Are there instructions on how to do so?

Here's how I tested my .mk file (on a fresh installation of Debian 12):

```sh
git clone https://github.com/hax0rbana-adam/mxe
cd mxe
git checkout libbotan2
sudo apt-get install -y autoconf automake autopoint bash bison bzip2 flex g++ g++-multilib gettext git gperf intltool libc6-dev-i386 libgdk-pixbuf2.0-dev libltdl-dev libgl-dev libpcre3-dev libssl-dev libtool-bin libxml-parser-perl lzip make openssl p7zip-full patch perl python3 python3-distutils python3-mako python3-packaging python3-pkg-resources python3-setuptools python-is-python3 ruby sed sqlite3 unzip wget xz-utils
time make MXE_TARGETS='x86_64-w64-mingw32.static i686-w64-mingw32.static' libbotan2
find . -name '*botan*a'
find . -name '*botan*.pc'
```

### DIFF
--- a/src/botan2.mk
+++ b/src/botan2.mk
@@ -1,0 +1,43 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := botan2
+$(PKG)_WEBSITE  := https://botan.randombit.net/
+$(PKG)_DESCR    := botan2
+$(PKG)_IGNORE   :=
+$(PKG)_VERSION  := 2.19.5
+$(PKG)_CHECKSUM := dfeea0e0a6f26d6724c4af01da9a7b88487adb2d81ba7c72fcaf52db522c9ad4
+$(PKG)_SUBDIR   := Botan-$($(PKG)_VERSION)
+$(PKG)_FILE     := Botan-$($(PKG)_VERSION).tar.xz
+$(PKG)_URL      := https://botan.randombit.net/releases/$($(PKG)_FILE)
+$(PKG)_DEPS     := cc
+
+define $(PKG)_UPDATE
+     $(WGET) -q -O- 'https://botan.randombit.net/releases/' | \
+     grep 'a href="Botan-2' | \
+     $(SED) 's/.*Botan-\(2.[0-9.]*\)\..*/\1/' | \
+     $(SORT) -Vr | \
+     head -1
+endef
+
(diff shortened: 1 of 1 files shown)


mabrand:
Thanks. Just a few comments and questions:
1. This should probably be called "botan.mk" instead of "libbotan.mk". We try to name the packages according to the project name.
2. Since the code is hosted on github, have you considered using $(PKG)_GH_CONF to base this on the github repo? This would take care of a lot of details automatically, including the update macro. You could look at the other .mk files for examples.
3. You are proposing version 2.19.5, but Boton 2 is no longer maintained since the beginning of 2025. . Is there a reason not to use Boton 3?

hax0rbana-adam:
1. I was matching the package name in Debian, but I can change it to be botan2.mk
2. I went with the official website instead of GitHub because that's what the project links to on their main page in the Downloads section. I expect changing it to GitHub will likely cause problems with the point below because it'll only want to build the latest, which is the 3.x series.
3. Yes, it's because keepassxc still uses Botan2 for the moment (and I'm trying to switch from using keepassx to keepassxc to re-use their code to import Keepass 4.x databases). I plan on porting keepassxc to use Botan3, creating an MXE package called botan3.mk. The problem with that is that there is no botan3 library in the latest stable version of Debian, so switching to Botan3 would mean Debian users wouldn't be able to compile projects that have moved to botan3. The next release (Debian 13; codename trixie) will have both botan2 and botan3, which should make for a smoother time to transition. In the meantime, botan2 it is.

If there is a security vulnerability in botan2 before Debian 13 is released, I expect that the Botan developers would backport the fix to the 2.x line. If they don't, the Debian security team will. And I'll be very interested in getting that patch from Debian over to this MXE package.

PR #3228, opened 2025-05-10
Fix issue with static build of libepoxy

gballabio:
Backport pull request https://github.com/anholt/libepoxy/pull/265 See also https://github.com/anholt/libepoxy/issues/200 for more information

This patch *replaces* the existing src/libepoxy-1-fixes.patch because it removes the code where it applied

Add test program (to gtkmm3) to demonstrate issue: without the fix it crashes on startup

  * install .pc file,

Thanks!

### DIFF
--- a/src/libepoxy-1-fixes.patch
+++ b/src/libepoxy-1-fixes.patch
@@ -2,29 +2,174 @@ This file is part of MXE. See LICENSE.md for licensing information.
 
 Contains ad hoc patches for cross building.
 
-From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
-From: v1993 <v19930312@gmail.com>
-Date: Mon, 6 Dec 2021 22:38:14 +0300
-Subject: [PATCH 1/1] Fix static builds on Windows
+Backport pull request https://github.com/anholt/libepoxy/pull/265
+See also https://github.com/anholt/libepoxy/issues/200 for more information
 
-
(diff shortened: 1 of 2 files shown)


src/libepoxy-1-fixes.patch
```diff
@@ -2,29 +2,174 @@ This file is part of MXE. See LICENSE.md for licensing information.
...
+-        wgl_init_dispatch_table();
+-    }
+-}
+-
+-/**
+- * This global symbol is apparently looked up by Windows when loading
+- * a DLL, but it doesn't declare the prototype.
+- */
+-BOOL WINAPI
+-DllMain(HINSTANCE dll, DWORD reason, LPVOID reserved);
```
melroy89, line 59: Dll Main is needed when creating a static binary. Why did you remove this?

gballabio:
Honestly, I don't know.
This patch isn't mine. As I wrote, I copied it from here:
https://github.com/anholt/libepoxy/pull/265
It was accepted upstream, so I assume it does make sense.
I tried it and it works for me: without it, the test program crashes. With
it, it runs.

Gerardo


Il giorno dom 11 mag 2025 alle ore 00:52 Melroy van den Berg <
***@***.***> ha scritto:

melroy89:
Did you test a static build?

gballabio:
Yes


Il giorno dom 11 mag 2025 alle ore 22:58 Melroy van den Berg <
***@***.***> ha scritto:

PR #3235, opened 2025-06-13
Do not force cmake to Release type on successive configs.

oe1rsa:
When configuration is done in multiple runs, such as when an initial run is done to populate the options and then succesive runs to apply the options, do not force Release type on second run.

qtcreator does exactly that: specify build type only on first run and then relies on cached variable.

Sorry, forgot to mention: My change is about the cmake wrapper and config scripts.


oe1rsa:
I am using it for several months now. I think it is ready for merge.

PR #3243, opened 2025-08-27
Fix secondary gcc download url, again

sabelka:
Sorry, my previous pull request contained an mistake, it was missing the gcc- prefix from the directory.
Unfortunately, I noted it just now during testing because download from the primary url failed spontaneously.


TWpgo:
We need this fix 👍

PR #3257, opened 2025-10-02
libmpeg2: fix for win64

mgerhardy:
https://github.com/scummvm/dockerized-bb/blob/master/toolchains/mxe/packages/libmpeg2/libmpeg2-9999-fix-win64.patch

https://forums.scummvm.org/viewtopic.php?f=2&t=16293

```
Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ff66119b240 in mpeg2_idct_copy_sse2 ()
(gdb) backtrace
#0  0x00007ff66119b240 in mpeg2_idct_copy_sse2 ()
#1  0x00007ff661195952 in mpeg2_slice ()
#2  0x00007ff66118aaa4 in mpeg2_parse ()
```

### DIFF
--- a/src/libmpeg2-9999-fix-win64.patch
+++ b/src/libmpeg2-9999-fix-win64.patch
@@ -1,0 +1,51 @@
+Index: libmpeg2-0.5.1/configure
+===================================================================
+--- libmpeg2-0.5.1.orig/configure
++++ libmpeg2-0.5.1/configure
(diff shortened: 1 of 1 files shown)


mabrand:
Does your build fail without this patch? libmpeg2 builds for me for targets x86_64-w64-mingw32.static and x86_64-w64-mingw32.shared as-is using gcc15.

PR #3297, opened 2026-03-04
opensubdiv: add opensubdiv.mk package definition

hkunz:
- Introduce a new MXE package for OpenSubdiv (libosdCPU), required as a dependency for Blender.
- Mark OpenSubdiv as depending on crew.mk for proper build ordering.
- Enables building and linking against OpenSubdiv in cross-compiled environments.
- Includes a minimal test for libosdCPU to verify successful compilation and linking.

### DIFF
--- a/src/clew.mk
+++ b/src/clew.mk
@@ -1,0 +1,24 @@
+# This file is part of MXE. See LICENSE.md for licensing information.
+
+PKG             := clew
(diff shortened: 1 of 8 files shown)


mgerhardy (commented):
nice job - from my side it could get merged (except the one note regarding the patch)

but also let others have a look at this, as this is touching code in common/ and not just an isolated mk file.

PR #3299, opened 2026-03-09
Add target libxml++5

gballabio:
File src/libxml++5.mk adds support for libxml++ version 5.0.
This does *not* replace the existing libxml++.mk (version 2.6): they are meant to coexist in parallel, like gtk2 and gtk3.


mabrand:
superseded by https://github.com/mxe/mxe/pull/3335

PR #3304, opened 2026-03-31
Openmpt

pepijn-devries:

  * install .pc file,

Thanks!


src/libopenmpt.mk
```diff
@@ -4,12 +4,12 @@ PKG             := libopenmpt
...
 $(PKG)_SUBDIR   := libopenmpt-$($(PKG)_VERSION)+release.autotools
 $(PKG)_FILE     := libopenmpt-$($(PKG)_VERSION)+release.autotools.tar.gz
 $(PKG)_URL      := https://lib.openmpt.org/files/libopenmpt/src/$($(PKG)_FILE)
-$(PKG)_CHECKSUM := a3e85e49c71633513369a473489895c104279b9087c5545598f45a0d33e9b110
-$(PKG)_DEPS     := cc zlib mpg123 libogg libvorbis flac libsndfile portaudio
+$(PKG)_CHECKSUM := 3c619bb36b3d9cd10b5531ebb83cac7fe38386d0614f9bb1a9a758010bc162d1
+$(PKG)_DEPS     := cc zlib mpg123 ogg vorbis flac libsndfile portaudio
 
 define $(PKG)_BUILD
     cd '$(1)' && ./configure \
```
hkunz:
```suggestion
    cd "$(BUILD_DIR)" && "$(SOURCE_DIR)/configure" \
```

src/libopenmpt.mk
```diff
@@ -4,12 +4,12 @@ PKG             := libopenmpt
 $(PKG)_WEBSITE  := https://lib.openmpt.org
 $(PKG)_DESCR    := libopenmpt
 $(PKG)_IGNORE   :=
-$(PKG)_VERSION  := 0.7.11
+$(PKG)_VERSION  := 0.7.19
```
hkunz, line 7: I'm curious why aren't we using 0.8.2?

hkunz (commented):
Looks good, just a few minor comments. I tested it and it works correctly, including with my test file, which you could also use for a more extensive decoding test: https://pastebin.com/wUXZLSRg

src/libopenmpt.mk
pepijn-devries: There is no specific reason. This was the version that I was using in my project. I'll have a look at whether the makefile can easily be updated to a newer version.

PR #3311, opened 2026-04-20
LLVM: update llvm package since it was very old

hkunz:
Update LLVM package which was very old. No more need patches.


mabrand:
Hi, Is this ready to merge, or does it need a new update?

PR #3325, opened 2026-08-14
KF6 Introduction: Second step - Qt6 D-Bus enablement & i686 build fixes

borrinimarco74:
This is the second step in introducing KDE Framework 6 into MXE. 
It focuses on enabling D-Bus in Qt6 and resolving several build failures specifically affecting the 32-bit (i686) targets.

This PR has been tested locally against all 4 targets with GCC 11 and GCC 16.

Next step: adding new dependencies and updating current packages to support the KDE Frameworks 6 build.

==> 5c9bfd88 qt6-qtbase: enable dbus <==
* Feature Enablement: Explicitly enabled D-Bus support in the qt6-qtbase build configuration, which is a crucial requirement for many KF6 components.

==> cabfaec1 dbus: patch to fix qt6-qtbase i686 static build <==
* Build Fix: Added a patch to resolve static linkage issues encountered when building qt6-qtbase against D-Bus on the i686 architecture.

==> 510151cc qt6-qtmultimedia: patch to fix i686 static build <==
* Build Fix: Fixed a compilation error in qwindows_propertystore.cpp on 32-bit targets. The patch adds the missing __stdcall template specialization for first_arg_is_ptr, which caused the build to fail on i686-w64-mingw32.

==> e566b631 qt6-qtwebview: patch to fix i686 static build <==
* CMake Fix: Resolved a fatal configuration error where FindWebView2.cmake failed to recognize the i686 architecture. The patch maps i686/i386/x86 to the correct x86 SDK directory, allowing the build to succeed.

==> 7bc4a8f1 mariadb-connector-c: fix i686 shared build <==
* Build Fix: Applied a minor correction to ensure mariadb-connector-c compiles successfully as a shared library on the i686 architecture.

### DIFF
--- a/src/dbus-2-fix-cmake-definitions.patch
+++ b/src/dbus-2-fix-cmake-definitions.patch
@@ -1,0 +1,33 @@
+This file is part of MXE. See LICENSE.md for licensing information.
+
+Contains ad hoc patches for cross building.
+
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Marco Borrini <borrinimarco74@gmail.com>
+Date: Tue, 4 Aug 2026 08:22:00 +0200
+Subject: [PATCH 2/2] Fix CMake definitions and missing transitive dependencies
+
+This patch fixes two bugs in DBus1Config.pkgconfig.in that break static builds:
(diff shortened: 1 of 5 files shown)


mabrand:
Thanks. Rebased and pushed as 81ac034695207f597deec8947725c9bfc1ab4001.

